Issue #1, opened 2026-04-30
[SECURITY]  Authorisation bypass in POST /chat/create permits cross-project chat injection


ryanmcdonough:
**Summary**
POST /chat/create accepts a caller-supplied project_id and creates a chat without verifying whether the authenticated user has access to that project.

Because backend routes use createServerSupabase() with SUPABASE_SECRET_KEY (service role), this path is not protected by RLS. With only a foreign key on public.chats.project_id, any authenticated user can attach new chats to any existing project ID.

**Impact**
Data integrity risk: users can write chats into projects they should not be able to modify.
Cross-tenant contamination risk: project-scoped chat views may surface unauthorised content.
Auditability risk: project histories can be polluted by unauthorised actors.

**Severity: High**

**Technical details**
Endpoint: POST /chat/create
Vulnerable behaviour: inserts into chats with project_id from request body and no project membership/ownership check.
Current DB constraints:
public.chats.project_id has FK to public.projects.id (existence only).
No chats RLS/policy/trigger in repo enforcing membership on insert.
Service-role Supabase client bypasses RLS by design.

**Reproduction steps**
Authenticate as User A (who does not own and is not shared on Project P).
Send POST /chat/create with body including project_id = <Project P id>.
Observe successful chat creation with that project_id.
Expected: request is rejected (404/403) unless caller has access to the target project.
Actual: request succeeds if project exists.

**Proposed fix**
Add explicit app-layer authorisation in POST /chat/create:

If project_id is provided, call shared access helper (checkProjectAccess(projectId, userId, userEmail, db)).
Reject when access is not granted.
Consider DB-side guardrails as well:
Add trigger/function on public.chats to validate project membership/ownership for inserts/updates.
Or avoid service-role for user-scoped writes where feasible.

willchen96:
Thanks for flagging, Ryan. This has now been fixed in recent security fixes.

PR #2, opened 2026-04-30
Enhance chat creation endpoint to check project access using user email

ryanmcdonough:
Fix Issue #1 

Added an explicit authorisation check to POST /chat/create so chats can only be created against projects the caller can access.

In backend/src/routes/chat.ts, the route now reads userEmail from auth context and, when project_id is provided, calls checkProjectAccess(projectId, userId, userEmail, db) before insert.
If access is denied, the endpoint now returns 404 Project not found and does not write a chat row.
If access is allowed (owner or shared member), behaviour is unchanged and chat creation proceeds.
This closes the app-layer project spoofing vector where authenticated users could previously create chats under arbitrary existing project_id values.


willchen96:
Thank you Ryan spotting this issue. This is addressed in my PR: https://github.com/willchen96/mike/pull/46

PR #5, opened 2026-05-01
fix: JSONB array queries fail with "invalid input syntax for type json"

sid-swirl:
`shared_with` is a jsonb column, but the Supabase JS .contains() helper serialises JavaScript arrays using Postgres array syntax ({value}) instead of JSON syntax (["value"]), causing every call to GET /projects to return a 500 for any user with a non-empty email address. 

Fixed by replacing .contains() with .filter('shared_with', 'cs', JSON.stringify([email])), which produces valid JSON that PostgREST can pass through to the @> operator correctly. 

Also adds forcePathStyle: true to the S3 client, which is required when running against a local MinIO instance (or any S3-compatible endpoint that doesn't support virtual-hosted-style bucket URLs).

### DIFF
--- a/backend/src/lib/access.ts
+++ b/backend/src/lib/access.ts
@@ -135,7 +135,7 @@ export async function listAccessibleProjectIds(
             ? db
                   .from("projects")
                   .select("id")
-                  .contains("shared_with", [userEmail])
+                  .filter("shared_with", "cs", JSON.stringify([userEmail]))
                   .neq("user_id", userId)
             : Promise.resolve({ data: [] as { id: string }[] }),
     ]);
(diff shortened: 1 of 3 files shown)


osama-ata (approved):
endorsing these modifications

willchen96:
Thanks for this. I manually ported the two intended fixes:

- replaced `shared_with` `.contains(...)` calls with `.filter("shared_with", "cs", JSON.stringify(...))` for JSONB containment
- added `forcePathStyle: true` to the S3 client for MinIO/S3-compatible local storage

I did not merge the PR directly because the branch also includes unrelated changes from an older codebase state, including changes to project folder handling and schema/files that would conflict with newer work. Closing this PR after porting the relevant fixes.

Issue #9, opened 2026-05-01
[Security] All LLM stream events, including document content, are logged to disk in plaintext


robinjfisher:
## Summary
         
In backend/src/lib/llm/claude.ts, every event emitted by the Anthropic streaming API is serialised and appended to a file named claude-raw-stream.log in the server's working directory:                                                                                                                                        
  
```
stream.on("streamEvent", (event) => {                                                                                                                              
    const line = JSON.stringify(event);                                                                                                                            
    console.log("[claude raw stream]", line);
    fs.appendFile(RAW_STREAM_LOG_PATH, line + "\n", () => {});                                                                                                     
});                                                                                                                                                                
```
                                                                                                                                                                     
This log captures the full content of every LLM interaction: user messages, document text passed as context, and assistant responses. The file grows without bound (no rotation, no size limit, no TTL), is written unconditionally in all environments, and is readable by any process on the server.
                                                                                                                                                                     
## Impact          

For a legal document platform, this is a significant confidentiality risk:                                                                                         
  
- Privileged content on disk. Legal documents processed through the assistant — contracts, briefs, discovery material — are serialised in full to a plaintext log file outside any access control.
- Unbounded growth. The file is never truncated or rotated. On a busy instance it will grow to consume available disk space, causing a secondary availability issue.                                                                                                                                                             
- No environment gate. The logging runs in production as well as development; there is no NODE_ENV check or feature flag.
- Regulatory exposure. Firms using this tool may be subject to GDPR, SRA, or equivalent obligations around client data. Accumulating uncontrolled copies of matter content in a log file likely violates those obligations.                                                                                                           
- Accidental commit risk. The log is written to process.cwd(), which is typically the repository root. If a developer runs the server locally, the file may be created alongside the source and accidentally committed or included in a build artefact.                                                                           
                  
##  Steps to reproduce                                                                                                                                                 
                  
1. Start the backend with a valid ANTHROPIC_API_KEY.                                                                                                               
2. Send any chat message that causes the Claude provider to be used.
3. Observe claude-raw-stream.log created in the backend working directory, containing the full streamed response including any document content passed as context. 
                                                                                                                                                                     
##  Proposed fix                                                                                                                                                       
                                                                                                                                                                     
1. Remove the fs.appendFile call entirely. The console.log on the preceding line already provides sufficient debug output when needed.                             
  
```
stream.on("streamEvent", (event) => {                                                                                                                              
    if (process.env.NODE_ENV !== "production") {
[...]

Issue #7, opened 2026-05-01
[Security] Download token HMAC falls back to hardcoded "dev-secret" when env vars are absent


robinjfisher:
## Summary
                                                                                                                                                                     
Download tokens are signed with an HMAC key resolved in backend/src/lib/downloadTokens.ts. If neither DOWNLOAD_SIGNING_SECRET nor SUPABASE_SECRET_KEY is set, the signing key falls back to the literal string "dev-secret":                                                                                                         
                  
```
function getSecret(): string {                                                                                                                                     
    return (    
        process.env.DOWNLOAD_SIGNING_SECRET ??
        process.env.SUPABASE_SECRET_KEY ??                                                                                                                         
        "dev-secret"
    );                                                                                                                                                             
}               
```

Because this codebase is public, the fallback value is known to anyone. An attacker who determines that a deployment is running without these env vars can compute valid HMAC signatures for any storage path they can guess or enumerate, producing forged download tokens that the /download/:token route will accept.
                                                                                                                                                                     
## Impact          

- Any authenticated user (or unauthenticated user if the route were ever relaxed) can forge a download token for an arbitrary R2 storage path without possessing the real secret.
- The /download/:token route still requires a valid Supabase session, so this does not enable fully unauthenticated access — but it does break the token's purpose of scoping downloads to authorised files.                                                                                                                          
- The companion risk: DOWNLOAD_SIGNING_SECRET is not listed in .env.example, so a deployer following the README will not know to set it, and the fallback to SUPABASE_SECRET_KEY (if set) silently doubles that key's role without documentation.                                                                               

## Steps to reproduce                                                                                                                                                 
                  
1. Deploy the backend without setting DOWNLOAD_SIGNING_SECRET or SUPABASE_SECRET_KEY.                                                                              
2. Compute HMAC-SHA256("dev-secret", <base64url-encoded payload>) for a payload {"p":"<known storage key>","f":"<filename>"}.
3. Issue a GET /download/<forged-token> with a valid Supabase session.                                                                                             
4. Observe the file is served without the token having been issued by the server.                                                                                  
                                                                                                                                                                     
## Proposed fix                                                                                                                                                       
                                                                                                                                                                     
1. Remove the "dev-secret" fallback entirely. If neither env var is set, fail fast at startup rather than silently degrading:                                      
  
```
function getSecret(): string {                                                                                                                                     
    const secret =
        process.env.DOWNLOAD_SIGNING_SECRET ??
[...]

PR #17, opened 2026-05-02
Fix three security issues: chat auth bypass, dev-secret fallback, plaintext stream log

DanialBeg:
## Summary

Bundles fixes for three small-surface-area security issues, each with its own commit so they can be reviewed (or split out) individually:

- **`#1`** — `POST /chat/create` now calls `checkProjectAccess` before inserting, closing the cross-project chat injection. The helper was already imported in `routes/chat.ts`.
- **`#7`** — `getSecret()` in `downloadTokens.ts` no longer falls back to the literal `"dev-secret"`; it throws when neither env var is set. `DOWNLOAD_SIGNING_SECRET` is now documented in `backend/.env.example`.
- **`#9`** — Raw Claude stream logging is now gated behind `DEBUG_LLM_STREAM=true`, so privileged matter content is no longer persisted to `claude-raw-stream.log` by default. (`*.log` is already gitignored, so the accidental-commit sub-point in the issue is already covered.)

Each commit references the issue it closes. `npm run build` passes in `backend/`.

## Test plan

- [x] `npm run build` in `backend/` (tsc clean)
- [ ] Manual: `POST /chat/create` with a `project_id` the caller does not own returns 404
- [ ] Manual: starting the backend with neither `DOWNLOAD_SIGNING_SECRET` nor `SUPABASE_SECRET_KEY` set, then hitting a route that signs a download URL, throws instead of producing a forgeable token
- [ ] Manual: with `DEBUG_LLM_STREAM` unset, no `claude-raw-stream.log` file is written during a chat session; setting it to `true` restores the previous behaviour

Happy to split into three PRs if you'd prefer that — let me know.

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -3,6 +3,12 @@ FRONTEND_URL=http://localhost:3000
 SUPABASE_URL=https://your-project.supabase.co
 SUPABASE_SECRET_KEY=your-supabase-service-role-key
 
+# HMAC key for signing /download/:token URLs. Use a randomly generated
+# value distinct from SUPABASE_SECRET_KEY (e.g. `openssl rand -hex 32`).
+# If unset, the backend falls back to SUPABASE_SECRET_KEY; if neither
+# is set, download signing throws at runtime rather than use a default.
+DOWNLOAD_SIGNING_SECRET=your-random-32-byte-hex-secret
+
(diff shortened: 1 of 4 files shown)


DanialBeg:
Closing at requester's request — no longer pursuing this change.

PR #20, opened 2026-05-02
Add LocalLLM/vLLM provider support

jpbreda:
### Overview
Adds support for OpenAI-compatible local LLM endpoints (vLLM) as a third provider alongside Claude and Gemini. This enables users to connect Mike to self-hosted LLM instances while maintaining compatibility with commercial APIs.

### Key Changes

**Backend:**
- New `openai.ts` provider adapter for OpenAI-compatible endpoints
- Configurable model names via environment variables (`VLLM_MAIN_MODEL`, `VLLM_LIGHT_MODEL`)
- Server-side configuration (no user API key management required)
- Added fallback for missing `title` parameter in `generate_docx` tool

**Frontend:**
- Added "LocalLLM" model group to model selector
- Default models: `LocalLLM Main` and `LocalLLM Lite`
- Updated model availability checks to treat LocalLLM as always available
- Updated account settings page to reflect LocalLLM configuration

**Configuration:**
- New environment variables:
  - `VLLM_BASE_URL` - Endpoint URL for OpenAI-compatible API
  - `VLLM_API_KEY` - Optional API key for authentication
  - `VLLM_MAIN_MODEL` - Model name for main chat tasks
  - `VLLM_LIGHT_MODEL` - Model name for lightweight tasks (title generation)

### Testing
- Tested with vLLM endpoint at `https://bredaai.com/api/v1`
- Verified document generation workflow
- Confirmed model selection and routing works correctly

### Notes
- LibreOffice is required for document conversion (DOCX ↔ PDF)
- LocalLLM models are server-configured and available to all users
- Maintains backward compatibility with existing Claude/Gemini providers

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,23 @@ Open `http://localhost:3000`.
 - At least one supported model provider key, depending on which models you enable
 - LibreOffice for DOC/DOCX to PDF conversion
 
+## LLM Configuration Options
+
+The backend supports multiple LLM providers via environment variables in `backend/.env`:
+
+### Provider Keys
+- `GEMINI_API_KEY` - Google Gemini models
+- `ANTHROPIC_API_KEY` - Anthropic Claude models
+- `OPENROUTER_API_KEY` - OpenRouter (aggregates multiple providers)
+- `RESEND_API_KEY` - Resend (for email functionality)
+
+### Local LLM (vLLM) Configuration
+For self-hosted vLLM endpoints:
+- `VLLM_BASE_URL` - Base URL for your vLLM server (e.g., `https://your-vllm-endpoint.com/v1`)
+- `VLLM_API_KEY` - API key for vLLM authentication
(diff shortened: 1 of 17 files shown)


willchen96:
Thanks for this, Joseph. I agree that supporting self-hosted inference is important, and I also agree that OpenAI-compatible endpoints as implemented in #20 are the right approach here.

However, the OpenAI chat endpoint support has now been implemented ( , so I’m going to close this PR rather than merge it as-is.

Will need to follow up with a new PR that builds on the current provider layer and implements the proposed configurable OpenAI-compatible endpoint changes properly, so it can support self-hosted engines like Ollama, vLLM, LM Studio, LocalAI, etc. without separate one-off integrations. 

I’ll probably pick this up later in the week in a new PR. In the meantime, if you’re able to rebase your changes on the current main branch, that would be very helpful. Thank you!

punyaslokdutta:
https://github.com/willchen96/mike/pull/122 (Ollama + qwen)

PR #28, opened 2026-05-04
fix(projects): validate folder ownership before folder mutations

fayerman-source:
## Summary
- validate parent folders against the current project before moving folders
- validate target folders against the current project before moving documents
- verify folder ownership before deletion and scope document cleanup to the current project

## Why
Project folders are project-scoped, but a few mutation paths accepted folder IDs without first proving they belonged to the same project. In a legal document workspace, avoiding cross-project references is part of the basic trust boundary.

## Test


fayerman-source:
Small legal-workflow note: this is intentionally scoped to project/folder boundary integrity rather than a larger permissions refactor. In a document workspace used for confidential material, preventing cross-project folder references is one of those quiet data-boundary checks that helps keep later sharing/collaboration features from drifting into confusing states.

willchen96 (approved):
Thank you.

PR #31, opened 2026-05-05
docs: add safe local testing guide

fayerman-source:
## Summary
- add a safe local testing guide for lawyers and firms evaluating Mike
- link the guide from the README setup flow
- remove the Supabase service-role key from the frontend env example

## Why
Mike is likely to attract legal users who want to test it with sensitive workflows. The first-run docs should steer people toward disposable infrastructure, synthetic documents, capped provider keys, and keeping service-role/model-provider secrets server-side.

This also removes a footgun from the frontend example env file: the browser-facing app should not ask new self-hosters to place a Supabase service-role key in `frontend/.env.local`.

## Test
- Documentation-only change


willchen96 (approved):
Thank you.

PR #32, opened 2026-05-05
feat(mcp): add Connectors — URL+headers and OAuth 2.1

ZachLaik:
## Summary

Adds **Connectors**: users can plug their own [Model Context Protocol](https://modelcontextprotocol.io) servers into Mike via Settings → Connectors, with no code change. Tools discovered from each enabled connector are merged into the chat assistant's per-request tool list and dispatched back to the right server at call time.

Two auth modes are supported:

- **API key / headers** — paste an MCP URL + optional custom headers (e.g. `Authorization: Bearer ...`). Works for self-hosted servers and any token-issuing service.
- **OAuth 2.1 (auto-discover)** — for spec-conformant servers (RFC 9728 + RFC 7591 dynamic client registration + PKCE). One-click sign-in via popup, no token to paste, auto-refresh on expiry. Verified end-to-end against `https://legaldatahunter.com/mcp`.

## What's in

**Backend** (Express/TS)

- `user_mcp_servers` table (RLS owner-only) with two migrations + the same DDL inlined into `000_one_shot_schema.sql`.
- `lib/mcp/{client,servers,types,oauth}.ts` — Streamable-HTTP client wrapper around the official `@modelcontextprotocol/sdk`, per-request loader, MCP `inputSchema` → Mike's `OpenAIToolSchema` converter (with 64-char tool-name guard), `OAuthClientProvider` impl backed by the row's oauth_* columns.
- `routes/mcpServers.ts` (`/user/mcp-servers`) — CRUD + `/test` (connect + list_tools probe) + `/oauth/start` (returns authorize URL for popup).
- `routes/mcpOauth.ts` (`/mcp/oauth/callback`) — public callback that verifies an HMAC-signed state token, finishes the SDK `auth()` flow, and `postMessage`s the opener.
- `lib/chatTools.ts` — extends `runLLMStream` and `runToolCalls` with an `mcp__` dispatch branch; emits `mcp_tool_result` SSE events with capped args/output for in-chat observability.
- `routes/{chat,projectChat}.ts` — load enabled connectors at request start, close clients in `finally`.

**Frontend** (Next.js)

- New Settings tab `/account/mcp`: add / edit / delete / enable-disable connectors. Auth-mode radio (headers vs OAuth). For OAuth: Save & sign in opens a popup, the page polls until tokens land, then auto-runs tool discovery. For headers: tools are auto-discovered immediately on save.
- **Connectors** button in the chat input next to Documents / Workflows — popover with per-server toggle switches (hides itself when no connectors).
- `mcp_tool_result` block in assistant messages: `Called <Server> · <tool>` with a one-line preview and an expand for full pretty-printed JSON args + raw output.
- Trust warning at the top of the page; secret-leak guard that redacts the Name field if it looks like a Bearer token was pasted into it.
- `mikeApi.ts` typed client wrappers.

**Env**

- New optional `BACKEND_PUBLIC_URL` (defaults to `http://localhost:${PORT}`) used to build the OAuth callback URL. Added to `.env.example`.

## What's not in (deliberate, follow-up PRs)

- **Curated marketplace** of trusted connectors with one-click install — straightforward layer on top once this lands.
- **Per-row encryption** of header values + `oauth_tokens`. Currently they're stored at-rest in jsonb (RLS owner-only), which matches the existing precedent set by `user_profiles.{claude,gemini}_api_key`. Worth a dedicated hardening PR.

## Security notes for reviewers

- Backend uses the Supabase service-role key, so all `/user/mcp-servers/*` handlers explicitly filter by `user_id = res.locals.userId`. RLS on the table is belt-and-suspenders for any direct client access.
- URL validation rejects non-HTTPS URLs except for `localhost`/`127.0.0.1`.
- Headers capped at 20 entries × 4 KB.
- The `GET /user/mcp-servers` response returns header **keys** only and a boolean `oauth_authorized` — header values and access tokens never round-trip to the browser, even to the row's owner (defense in depth — RLS would allow it).
- OAuth state token: HMAC-signed with `DOWNLOAD_SIGNING_SECRET`, 5 min TTL, carries `user_id` + `server_id` only. CSRF-safe across the popup hop with no server-side session needed.
- Mike registers as a **public** OAuth client (`token_endpoint_auth_method=none`, PKCE-protected) — no client secret to store confidentially.
- `mcp_tool_result` SSE events truncate args + output to 4 KB before persistence to keep `chat_messages.content` from bloating; the model still sees the full untruncated tool output.

## Test plan

- [x] `npm run build --prefix backend` clean
- [x] `npx eslint` clean on changed frontend files
- [x] Manual: add a public no-auth header-mode connector (e.g. `https://mcp.deepwiki.com/mcp`) → tools auto-discovered → chat invokes a tool → `mcp_tool_result` block shows args/output → reload chat, persisted blocks render the same.
- [x] Manual: bad URL → graceful failure, `last_error` populated, chat still works.
- [x] Manual: disabled connector → tools not exposed.
- [x] Manual: connector with `Authorization` header against a real authed MCP → end-to-end tool call succeeds.
- [x] Manual: OAuth-mode connector against `https://legaldatahunter.com/mcp` → popup opens at LDH → sign in → popup closes → row flips to `OAuth · signed in` → tools auto-discover → chat invokes an `mcp__legal-data-hunter__search` tool successfully.

## Note (not introduced by this PR)

`npm run build --prefix frontend` fails at prerender on `/account/*` pages because `frontend/src/lib/supabase.ts` calls `createClient("", "")` at module load when env vars are absent. This affects `main` identically; dev mode is unaffected. Happy to fix in a separate PR if helpful.

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -1,5 +1,8 @@
 PORT=3001
 FRONTEND_URL=http://localhost:3000
+# Externally-reachable backend URL. Used to build the OAuth callback URL for
+# MCP connectors. Defaults to http://localhost:${PORT} when unset.
+BACKEND_PUBLIC_URL=http://localhost:3001
 SUPABASE_URL=https://your-project.supabase.co
 SUPABASE_SECRET_KEY=your-supabase-service-role-key
 
--- a/backend/migrations/000_one_shot_schema.sql
+++ b/backend/migrations/000_one_shot_schema.sql
@@ -273,6 +273,51 @@ end;
 $$;
 
 -- ---------------------------------------------------------------------------
+-- User MCP servers
+-- ---------------------------------------------------------------------------
+
+create table if not exists public.user_mcp_servers (
+  id uuid primary key default gen_random_uuid(),
+  user_id uuid not null references auth.users(id) on delete cascade,
+  slug text not null,
+  name text not null,
+  url text not null,
+  headers jsonb not null default '{}'::jsonb,
+  enabled boolean not null default true,
+  last_error text,
+  created_at timestamptz not null default now(),
+  updated_at timestamptz not null default now(),
+  constraint user_mcp_servers_slug_format
+    check (slug ~ '^[a-z0-9_-]{1,24}$'),
+  unique (user_id, slug)
+);
+
+create index if not exists idx_user_mcp_servers_user
+  on public.user_mcp_servers(user_id, enabled);
+
+alter table public.user_mcp_servers enable row level security;
+
+drop policy if exists "Users can view their own MCP servers" on public.user_mcp_servers;
+create policy "Users can view their own MCP servers"
+  on public.user_mcp_servers for select
+  using (auth.uid() = user_id);
+
+drop policy if exists "Users can insert their own MCP servers" on public.user_mcp_servers;
+create policy "Users can insert their own MCP servers"
+  on public.user_mcp_servers for insert
+  with check (auth.uid() = user_id);
+
+drop policy if exists "Users can update their own MCP servers" on public.user_mcp_servers;
+create policy "Users can update their own MCP servers"
+  on public.user_mcp_servers for update
+  using (auth.uid() = user_id);
+
+drop policy if exists "Users can delete their own MCP servers" on public.user_mcp_servers;
+create policy "Users can delete their own MCP servers"
+  on public.user_mcp_servers for delete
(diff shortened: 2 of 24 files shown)


willchen96:
Thanks Zach. Just added MCP connectors in the latest update. Did my own implementation as the code has moved along significantly since this. Will likely offer integration with your project in the future.

PR #43, opened 2026-05-08
Update README.md with a few gotchas

dave-parachute:
Hey @willchen96 really enjoyed playing with Mike OSS.
                                                                                                                        
I tried installing it as if I were a lawyer rather than an engineer and ran into a handful of papercuts that I think are easy fixes:                                                                                                 
                                                                                                                        
  ### README changes                                                                                                    
  - Add a `git clone` step (the existing setup jumps straight to `npm install`, which assumes the reader has already cloned)                                                          
  - Promote prerequisites (Node, git, Supabase, R2, model provider) into their own section with links so lawyers know what tools they're signing up for                                            
  - Walk through the steps in order: clone → install → env files → run schema → start backend → start frontend → sign up → add per-user API key                                                     
  - Document the per-user API key gotcha — `ANTHROPIC_API_KEY` in `.env` is a  server-side fallback only; the model picker is gated on a per-user key (or at least seemed to be)                                                
  - Note that Supabase moved its API page recently — Project URL is now under Settings → Data API, and the env vars expect the "Legacy" `anon` /  `service_role` keys (not the new `sb_publishable_*` / `sb_secret_*` format)  - possibly want to update this to new format
  - Add an env-var reference table with where to get each value                                                         
  - Mark LibreOffice as optional (only needed for `.doc` / `.docx` upload) - figure people could still play around with out it                                          
  - Add a Troubleshooting section covering the three issues I hit as a first-time user: `npm install` peer-dep warnings vs hard `ERESOLVE`  failures, Supabase confirmation emails (rate-limited / team-restricted by default on newer Supabase projects) and the per-user model API key gating

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,31 +8,58 @@ Open-source release containing the Mike frontend and backend.
 - `backend/` - Express API, Supabase access, document processing, and migrations
 - `backend/migrations/000_one_shot_schema.sql` - one-shot Supabase schema for fresh databases
 
+## Prerequisites
+
+You'll need accounts and/or tools for the following before you start:
+
+- **[Node.js](https://nodejs.org/)** 20+ - verify with `node -v`
+- **[git](https://git-scm.com/downloads)** - verify with `git --version`
+- A **[Supabase](https://supabase.com/)** project - free tier is fine
+- A **[Cloudflare R2](https://developers.cloudflare.com/r2/)** bucket - or any S3-compatible object store
+- At least one model provider API key:
+  - **[Anthropic](https://console.anthropic.com/settings/keys)** (Claude)
+  - **[Google Gemini](https://aistudio.google.com/apikey)**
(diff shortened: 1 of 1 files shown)


willchen96 (commented):
Hi Dave, thanks for the PR.

Could you please make the following updates?

- Keep the note about email confirmation, but word it as:
  “Currently, signup does not require email confirmation. For local development, disable email confirmation in Supabase → Authentication → Providers → Email. For production, configure custom SMTP in Supabase. The built-in mailer is heavily rate-limited and, on newer projects, may be restricted to team-member email addresses only.”

- Please remove the model picker comment as well, since that will be resolved in a separate PR.

Thanks again.

willchen96:
Thanks, Readme updated in latest commit. Did not directly merge due to changes which rendered parts of the PR out of date.

PR #44, opened 2026-05-08
Add Docker Compose dev stack and local Supabase setup

0xNadr:
## Summary

- `docker compose up` now brings up frontend (`:4000`) and backend (`:4001`) with bind-mounted source for hot reload.
- `supabase init` scaffolding plus the existing one-shot schema copied into `supabase/migrations/`, so `supabase start` brings up local Postgres + Auth + Storage + Realtime and auto-applies the schema.
- README gains a "Run locally with Docker" section above the existing manual setup, including ports, env wiring (`host.docker.internal` for server-side, `localhost` for browser), and the one-time storage bucket bootstrap.

The container `command` runs `npm install` on start instead of relying on the image's baked `node_modules`. This works around a Docker Desktop quirk where a named volume mounted on top of a bind-mounted `/app` gets a partial copy of `node_modules` from the image (e.g. Next was missing `format-cli-help-output`).

Default ports moved to `4000`/`4001` to avoid colliding with anything else commonly running on `3000`/`3001`. The legacy `npm run dev` workflow still uses `3000`/`3001` and is unaffected.

## Test plan

- [x] `supabase start` brings up the Supabase stack and applies the migration (15 tables in `public`)
- [x] `docker compose up -d` starts both Mike services
- [x] `curl http://localhost:4001/health` returns `{"ok":true}`
- [x] `curl http://localhost:4000/` redirects to `/assistant` (HTTP 200, ~30KB)
- [x] Backend container reaches Supabase Auth via `host.docker.internal:54321`
- [x] Source edits hot-reload inside both containers

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,76 @@ Open-source release containing the Mike frontend and backend.
 - `backend/` - Express API, Supabase access, document processing, and migrations
 - `backend/migrations/000_one_shot_schema.sql` - one-shot Supabase schema for fresh databases
 
-## Setup
+## Run locally with Docker + local Supabase
+
+Spins up the frontend, backend, and a full local Supabase stack (Postgres, Auth, Storage, Realtime) — no cloud account required.
+
+Prerequisites: Docker Desktop, [Supabase CLI](https://supabase.com/docs/guides/cli) (`brew install supabase/tap/supabase`).
+
+```bash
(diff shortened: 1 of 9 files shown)


willchen96:
Thanks for this. Sorry I wasn't able to get to merging this. Closing as main has moved and there are newer implementations.

PR #47, opened 2026-05-08
R2 -> Supabase Object storage

punyaslokdutta:
R2 requires a separate Cloudflare account, API tokens, and bucket configuration on top of the existing Supabase
  setup. Since Supabase Storage covers the same use cases (upload, download, delete, signed URLs) and the project
  already depends on Supabase for auth and Postgres, using it for storage eliminates an entire external service from
  the local setup and production stack.


willchen96:
Thanks for the PR. I agree with the motivation: using Supabase Storage would make the OSS/local setup simpler because the project already depends on Supabase for auth and Postgres.

I’m going to hold off on merging this as a replacement for R2, mainly because of the cost tradeoff. Document-heavy usage can involve a lot of storage reads: previews, downloads, generated versions, PDF renditions, and tabular-review workflows. Supabase Storage can incur bandwidth/egress costs as that usage grows, whereas R2/S3-compatible storage may be materially cheaper for high-volume storage and egress-heavy workloads.

I don’t want to remove the lower-cost R2/S3 path as the default production option.

A future version of this would be very welcome if it makes storage pluggable rather than replacing R2 outright, for example:

```env
STORAGE_PROVIDER=s3
# or
STORAGE_PROVIDER=supabase

punyaslokdutta:
Makes sense, can have STORAGE_PROVIDER=s3
# or
STORAGE_PROVIDER=supabase

PR #52, opened 2026-05-10
fix: handle JSONB shared_with filters and path-style S3

willchen96:
## Summary

Fixes project sharing queries against the `shared_with` JSONB column and improves compatibility with local/S3-compatible storage providers.

## Changes

- Replaced Supabase `.contains("shared_with", [email])` calls with `.filter("shared_with", "cs", JSON.stringify([email]))`.
- Applied the JSONB containment fix to:
  - `GET /projects`
  - `listAccessibleProjectIds`
- Added `forcePathStyle: true` to the S3 client configuration.

## Why

`shared_with` is a JSONB column. Passing a JavaScript array to `.contains()` can be serialized in a way that PostgREST treats as Postgres array syntax instead of JSON, causing `GET /projects` to fail for users with shared projects.

`forcePathStyle: true` allows the S3 client to work with local MinIO and other S3-compatible endpoints that do not support virtual-hosted-style bucket URLs.

## Testing

- Synced from private repo to OSS export.

### DIFF
--- a/backend/src/lib/access.ts
+++ b/backend/src/lib/access.ts
@@ -135,7 +135,7 @@ export async function listAccessibleProjectIds(
             ? db
                   .from("projects")
                   .select("id")
-                  .contains("shared_with", [userEmail])
+                  .filter("shared_with", "cs", JSON.stringify([userEmail]))
                   .neq("user_id", userId)
             : Promise.resolve({ data: [] as { id: string }[] }),
     ]);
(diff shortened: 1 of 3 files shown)


willchen96:
Implements PR fix: JSONB array queries fail with "invalid input syntax for type json" #5 by sid-swirl without conflicting project validation changes

PR #58, opened 2026-05-11
Replace custom DOCX tracked changes with docx-track-changes library

becker-charles:
## Changes

- Replace approx. 1179 line custom docxTrackedChanges.ts with approx. 370 line wrapper around docx-track-changes library
- Switch from fuzzy text matching (find + context_before/after) to paragraph ID based editing (paraId)
- Mint and persist stable paragraph IDs on first document read
- Fix stale preview after accept/reject by bumping refetchKey
- Remove unused fast-diff and fast-xml-parser dependencies

## Benefits

- More reliable edits: Paragraph IDs are unambiguous vs fuzzy text matching that could match wrong location
- Simpler maintenance: Leverage maintained library instead of custom XML manipulation
- Better error messages: Clear feedback when paraId does not exist
- Smaller codebase: Approx. 800 fewer lines to maintain


becker-charles:
**Additional fixes for tracked changes UI:**

- Noticed two separate chat routes (`/assistant/chat` and `/projects/.../chat`) with duplicate but inconsistent `handleEditResolved` implementations (should be consolidated in future refactor)
- Fixed bulk "Accept all" leaving markup visible in standalone chat view by adding `refetchKey` to ChatView/AssistantSidePanel/DocPanel flow
- Fixed project view EditCards not updating to show accepted/rejected status by adding `resolvedEditStatuses` state and prop
- Added missing `onEditResolveStart` and `isEditReloading` to project view for progress tracking during bulk operations
- Both routes now handle edit resolution consistently with loading states, progress indicators, and UI updates

becker-charles:
- **Bulk accept/reject optimization**: Added `isBulk` flag and `onBulkComplete` callback so bulk operations trust optimistic updates during processing and only trigger one refetch at the end (instead of per-edit)

- **Fixed optimistic update not hiding deleted text**: Backend was returning tracked change IDs without text, and frontend was matching by index order (which didn't match DOM order). Now backend returns text content with each ID, and frontend matches elements by text content to ensure correct tagging.

becker-charles:
@willchen96: The scope of this PR expanded slightly. While implementing the docx-track-changes library, I noticed some bugs/inefficiencies in how Mike handles displaying tracked changes optimistically (and ultimately syncing with the actual document state), which I've fixed as part of the work. If you would prefer separate PRs, let me know.

Otherwise, it should be good to go after review.

willchen96:
Hi @becker-charles thanks for this very helpful PR. Unfortunately I was unable to get to this at the time when you made the PR as I was doing a large refactor of the codebase in parallel. Never the less the issues highlighted here are very helpful. I am slightly concerned about replacing custom code with a package as the size reduction is not that substantial and we might lose the ability of customisation. But the package is very clean and has great ideas. I will test it out at some point. Created an issue linking to this #264

willchen96:
Created an issue linking to this #264

PR #61, opened 2026-05-11
fix(security): patch Next.js and React

nikohann:
Bump Next.js and React to latest patch versions for security fixes.


willchen96:
Thank you Niko for flagging this. Fixed in latest PR merge.

PR #73, opened 2026-05-13
fix: remove auth bypass dev fallback in getUserIdFromRequest

### LINKED ISSUE #65: [Security][Critical] Auth bypass dev fallback — unauthenticated account takeover
## Severity: Critical
**File:** `frontend/src/lib/supabase-server.ts:27-29`
**CWE:** CWE-287 — Improper Authentication
**OWASP:** A07:2021 — Identification and Authentication Failures

## Description
`getUserIdFromRequest()` contains a branch that, when `NEXT_PUBLIC_SUPABASE_URL` or `SUPABASE_SECRET_KEY` is absent, skips all token verification and returns the raw Bearer string verbatim as the authenticated user ID:

```typescript
if (!supabaseUrl || !serviceKey) {
    // Dev fallback — accept raw token as user ID
    return token;
}
```

This affects any environment where these variables are unset — CI pipelines, staging, or self-hosted deployments with incomplete configuration.

## Impact
An attacker who knows any valid user UUID passes it as a Bearer token and receives full authenticated access to that user's documents, projects, and profile data. In a legal document assistant this constitutes a material privacy breach reportable under GDPR/state privacy laws.

## Proof of Concept
```bash
TARGET_USER_UUID="aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"
curl -s https://app.example.com/api/documents \
  -H "Authorization: Bearer ${TARGET_USER_UUID}"
# Returns target user's document list when env vars are absent
```

## Fix
Remove the fallback. Fail hard when env vars are missing:

```typescript
if (!supabaseUrl || !serviceKey) {
    throw new Response("Server auth is not configured", { status: 500 });
}
```

**Remediation tier:** Immediate — block all production deploys until resolved.

### LINKED ISSUE #87: Nit PR #73: createServerSupabase() silently creates client with empty credentials when env vars are missing
## Found in PR #73 (fix/65-auth-bypass-fallback)

**Severity:** Minor

`createServerSupabase()` uses `|| ""` for both env vars:

```typescript
export function createServerSupabase() {
    const url = process.env.NEXT_PUBLIC_SUPABASE_URL || "";
    const key = process.env.SUPABASE_SECRET_KEY || "";
    return createClient(url, key, { auth: { persistSession: false } });
}
```

If either env var is missing, this creates a Supabase client with empty strings. The client will be constructed without error, but every subsequent DB call will fail with an obscure Supabase error rather than a clear config error at startup. 

`getUserIdFromRequest` (also in this file) correctly throws a 500 when env vars are missing. `createServerSupabase` should be consistent.

**Fix:** Throw early if either var is absent, matching the pattern in `getUserIdFromRequest`.

### LINKED ISSUE #89: Nit PR #73: getUserIdFromRequest still uses || "" fallback for env vars (inconsistent with createServerSupabase)
## Found in PR #73 (fix/65-auth-bypass-fallback)

**Severity:** Nit

`createServerSupabase()` was updated to read env vars directly without a fallback:
```typescript
const url = process.env.NEXT_PUBLIC_SUPABASE_URL;
const key = process.env.SUPABASE_SECRET_KEY;
if (!url || !key) { throw ... }
```

But `getUserIdFromRequest` in the same file still reads with `|| ""` fallback:
```typescript
// frontend/src/lib/supabase-server.ts lines 27-28
const supabaseUrl = process.env.NEXT_PUBLIC_SUPABASE_URL || "";
const serviceKey = process.env.SUPABASE_SECRET_KEY || "";

if (!supabaseUrl || !serviceKey) {
    throw new Response("Server auth is not configured", { status: 500 });
}
```

The logic is correct (empty string is falsy so the guard still triggers), but the pattern is inconsistent within the same file. The `|| ""` is dead code.

**Fix:** Remove the `|| ""` fallbacks — read the env vars directly as in `createServerSupabase()`.

### LINKED ISSUE #90: Nit PR #73: No test for createServerSupabase() throwing when env vars are missing
## Found in PR #73 (fix/65-auth-bypass-fallback)

**Severity:** Nit

PR #73 added a guard to `createServerSupabase()` that throws when `NEXT_PUBLIC_SUPABASE_URL` or `SUPABASE_SECRET_KEY` are missing:

```typescript
export function createServerSupabase() {
    const url = process.env.NEXT_PUBLIC_SUPABASE_URL;
    const key = process.env.SUPABASE_SECRET_KEY;
    if (!url || !key) {
        throw new Error("NEXT_PUBLIC_SUPABASE_URL and SUPABASE_SECRET_KEY must be set");
    }
    return createClient(url, key, { auth: { persistSession: false } });
}
```

The test file `frontend/src/lib/__tests__/supabase-server.test.ts` covers `getUserIdFromRequest` thoroughly but has zero tests for `createServerSupabase()`. The new guard that throws on missing env vars is completely untested.

**Fix:** Add tests to the existing test file:
```typescript
describe("createServerSupabase", () => {
    it("throws when NEXT_PUBLIC_SUPABASE_URL is missing", async () => {
        process.env.SUPABASE_SECRET_KEY = "some-key";
        const { createServerSupabase } = await import("../supabase-server");
        expect(() => createServerSupabase()).toThrow();
    });
    it("throws when SUPABASE_SECRET_KEY is missing", async () => {
        process.env.NEXT_PUBLIC_SUPABASE_URL = "https://project.supabase.co";
        const { createServerSupabase } = await import("../supabase-server");
        expect(() => createServerSupabase()).toThrow();
    });
});
```

bmersereau:
## Summary
- Removes the dev fallback in `getUserIdFromRequest()` that accepted any Bearer token verbatim as a user ID when env vars were absent
- Missing `NEXT_PUBLIC_SUPABASE_URL` or `SUPABASE_SECRET_KEY` now throws a 500 response instead of silently bypassing authentication
- `createServerSupabase()` also now throws when env vars are missing instead of silently creating a client with empty credentials
- Removes redundant `|| ""` fallbacks from `getUserIdFromRequest` for consistency
- Adds vitest and unit tests covering missing-env-var cases for both functions

Closes #65
Closes #87
Closes #89
Closes #90

## Changes
- `frontend/src/lib/supabase-server.ts` — `createServerSupabase` throws on missing env vars; `getUserIdFromRequest` removes `|| ""` fallbacks; bypass fallback replaced with `throw new Response(..., { status: 500 })`
- `frontend/src/lib/__tests__/supabase-server.test.ts` — 10 unit tests covering both `createServerSupabase` and `getUserIdFromRequest`
- `frontend/vitest.config.ts` — minimal vitest config

## Test plan
- [x] Unit tests added and passing (10/10)
- [x] TypeScript clean
- [ ] Build requires real Supabase env vars (pre-existing prerender failure without credentials)

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -7,6 +7,7 @@
         "": {
             "name": "mike",
             "version": "0.1.0",
+            "license": "AGPL-3.0-only",
             "dependencies": {
                 "@aws-sdk/client-s3": "^3.1025.0",
                 "@aws-sdk/s3-request-presigner": "^3.1025.0",
@@ -62,6 +63,7 @@
                 "tailwindcss": "^4",
                 "tw-animate-css": "^1.4.0",
                 "typescript": "^5",
+                "vitest": "^4.1.6",
                 "wrangler": "^4.90.0"
             }
         },
@@ -1866,9 +1868,9 @@
             }
         },
         "node_modules/@emnapi/core": {
-            "version": "1.9.2",
-            "resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.9.2.tgz",
-            "integrity": "sha512-UC+ZhH3XtczQYfOlu3lNEkdW/p4dsJ1r/bP7H8+rhao3TTTMO1ATq/4DdIi23XuGoFY+Cz0JmCbdVl0hz9jZcA==",
+            "version": "1.10.0",
+            "resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.10.0.tgz",
+            "integrity": "sha512-yq6OkJ4p82CAfPl0u9mQebQHKPJkY7WrIuk205cTYnYe+k2Z8YBh11FrbRG/H6ihirqcacOgl2BIO8oyMQLeXw==",
             "dev": true,
             "license": "MIT",
             "optional": true,
@@ -1878,9 +1880,9 @@
             }
         },
         "node_modules/@emnapi/runtime": {
-            "version": "1.9.2",
-            "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.9.2.tgz",
-            "integrity": "sha512-3U4+MIWHImeyu1wnmVygh5WlgfYDtyf0k8AbLhMFxOipihf6nrWC4syIm/SwEeec0mNSafiiNnMJwbza/Is6Lw==",
+            "version": "1.10.0",
+            "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.10.0.tgz",
+            "integrity": "sha512-ewvYlk86xUoGI0zQRNq/mC+16R1QeDlKQy21Ki3oSYXNgLb45GV1P6A0M+/s6nyCuNDqe5VpaY84BzXGwVbwFA==",
             "license": "MIT",
             "optional": true,
             "dependencies": {
@@ -3896,6 +3898,16 @@
                 "zod": "^3.25.0 || ^4.0.0"
             }
         },
+        "node_modules/@oxc-project/types": {
+            "version": "0.129.0",
+            "resolved": "https://registry.npmjs.org/@oxc-project/types/-/types-0.129.0.tgz",
+            "integrity": "sha512-3oz8m3FGdr2nDXVqmFUw7jolKliC4MoyXYIG2c7gpjBnzUWQpUGIYcXYKxTdTi+N2jusvt610ckTMkxdwHkYEg==",
+            "dev": true,
+            "license": "MIT",
+            "funding": {
+                "url": "https://github.com/sponsors/Boshen"
+            }
+        },
         "node_modules/@poppinss/colors": {
             "version": "4.1.6",
             "resolved": "https://registry.npmjs.org/@poppinss/colors/-/colors-4.1.6.tgz",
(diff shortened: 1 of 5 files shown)


bmersereau:
## PR Review: fix: remove auth bypass dev fallback from getUserIdFromRequest

### Summary
This PR closes the dev-mode bypass that allowed any raw Bearer token to be treated as a valid user ID when `NEXT_PUBLIC_SUPABASE_URL` or `SUPABASE_SECRET_KEY` were absent. The fix is correct and the new error is the right behavior. Test coverage is present but has gaps described below.

### Risk Assessment
- **Risk Level**: Low
- **Blast Radius**: Any environment that relied on the dev fallback (missing env vars) will now get a 500 instead of silently accepting the raw token — correct and intentional
- **Rollback Plan**: `git revert` the commit; fallback behavior is restored immediately

---

### Review by Category

#### Security
- **[severity:praise]** Removing the fallback is the right call. The old code at lines 27–29 was a direct authentication bypass — any string passed as a Bearer token became the user ID. The new `throw new Response("Server auth is not configured", { status: 500 })` is safe and fails loud.

#### Correctness
- **[severity:minor]** `createServerSupabase()` still constructs a client with empty strings when env vars are absent:
  ```typescript
  const url = process.env.NEXT_PUBLIC_SUPABASE_URL || "";
  const key = process.env.SUPABASE_SECRET_KEY || "";
  return createClient(url, key, { auth: { persistSession: false } });
  ```
  This is now inconsistent with the new stricter behavior in `getUserIdFromRequest`. A misconfigured `createServerSupabase()` call will silently return a client that fails at query time rather than at construction. Not a blocker, but worth aligning in a follow-up.

#### Test Coverage
- **[severity:minor]** Missing happy-path test: a well-configured server with a valid token should return the user ID. Without this, a regression that breaks the success path won't be caught.

- **[severity:minor]** Missing invalid-token test: `getUser` returning `{ data: { user: null } }` should yield a 401. This is the most security-relevant branch and has no test.

- **[severity:nit]** Test 1 ("throws 500 when NEXT_PUBLIC_SUPABASE_URL is missing") calls `getUserIdFromRequest` twice — once to assert the rejection, once to extract the status. The second call re-runs the async logic redundantly:
  ```typescript
  await expect(getUserIdFromRequest(req)).rejects.toBeInstanceOf(Response);
  const err = await getUserIdFromRequest(req).catch((r: Response) => r); // second call
  ```
  Use a single call with `rejects`:
  ```typescript
  const err = await getUserIdFromRequest(req).catch((r) => r);
  expect(err).toBeInstanceOf(Response);
  expect((err as Response).status).toBe(500);
  ```

---

### Test Coverage Assessment
- Lines added (non-test): 1 (the throw replacing 2 lines net)
- Test lines added: ~57
- Missing test cases: happy path (valid token → returns user ID), invalid token (getUser returns null user → 401)

---

### Verdict
- [ ] **Approve** - Ship it
- [x] **Approve with nits** — Add the happy-path and invalid-token tests before or shortly after merge; fix the double-call in test 1
- [ ] **Request changes** - Blocking issues must be addressed
- [ ] **Needs discussion** - Architecture/approach concerns

### What I Verified
- [x] Read the full diff
- [x] Ran tests locally (4/4 pass)
- [x] Checked `createServerSupabase()` for consistency
- [x] Reviewed linked issue #65

bmersereau:
## PR Review (follow-up): fix: remove auth bypass dev fallback from getUserIdFromRequest

All issues from the previous review have been addressed. This is a clean pass.

### What changed since last review
- `vi.hoisted()` used for a stable `mockGetUser` reference that survives `vi.resetModules()` — correct vitest idiom
- Double-call in test 1 fixed: single `catch` + two assertions
- Happy-path test added: valid token → returns `"real-user-id"`
- Invalid-token 401 test added: `mockGetUser` returns `null` user → 401 response
- `beforeEach` resets `mockGetUser` to the default so tests don't bleed into each other

### Remaining (non-blocking)
- **[severity:nit]** In `beforeEach`, placing `mockGetUser.mockResolvedValue(...)` before `vi.resetModules()` would read more naturally for someone unfamiliar with `vi.hoisted` — reset mock state first, then clear module registry. Behaviour is identical either way.

### Test Coverage
All branches of `getUserIdFromRequest` are now covered: missing URL (500), missing key (500), both missing (500), no auth header (401), valid token → user ID, invalid/expired token (401). Nothing missing.

### Verdict
- [x] **Approve** — Ship it

### What I Verified
- [x] Read the full diff
- [x] Ran tests locally — 6/6 pass
- [x] Confirmed `vi.hoisted()` mock survives `vi.resetModules()` correctly

LaurieScheepers:
These security enhancements should be merged.

willchen96:
Thanks Beau. This has been fixed as [frontend/src/lib/supabase-server.ts](https://github.com/willchen96/mike/pull/73/changes#diff-229027564d4f00c4dd5e4f9add749646ef41b261d103d2c26b380ead23f632ed) has now been deleted as supabase connection is moved to backend except for auth.

PR #76, opened 2026-05-13
fix: replace SHA-256 with HKDF + per-row salt for user API key encryption

### LINKED ISSUE #67: [Security][Critical] Weak AES key derivation for user API key encryption (SHA-256, no KDF)
## Severity: Critical
**File:** `backend/src/lib/userApiKeys.ts:39-48`
**CWE:** CWE-916 — Use of Password Hash With Insufficient Computational Effort
**OWASP:** A02:2021 — Cryptographic Failures

## Description
The AES-256-GCM encryption key for stored user API keys is derived with a raw SHA-256 hash — no KDF, no salt, no iteration count:

```typescript
function encryptionKey(): Buffer {
    const secret = process.env.USER_API_KEYS_ENCRYPTION_SECRET || ...
    return crypto.createHash("sha256").update(secret).digest();
}
```

## Impact
Anyone who obtains the secret computes the identical AES key in a single SHA-256 operation and immediately decrypts every row in `user_api_keys`. There is no per-entry salt, so all rows are decryptable in one pass — no per-entry brute force required.

Users store Anthropic, OpenAI, and Gemini API keys here. These keys have direct monetary value and can be used to run inference workloads billed to the victim.

## Fix
Replace SHA-256 with HKDF (RFC 5869) and store a per-row random salt:

```typescript
import { hkdfSync, randomBytes } from "crypto";

function deriveKey(secret: string, salt: Buffer): Buffer {
    return Buffer.from(hkdfSync("sha256", secret, salt, "mike-api-keys-v1", 32));
}

function encrypt(value: string): Omit<EncryptedKeyRow, "provider"> {
    const salt = randomBytes(16);
    const key = deriveKey(getEncryptionSecret(), salt);
    const iv = randomBytes(12);
    const cipher = createCipheriv("aes-256-gcm", key, iv);
    const encrypted = Buffer.concat([cipher.update(value, "utf8"), cipher.final()]);
    return {
        encrypted_key: encrypted.toString("base64"),
        iv: iv.toString("base64"),
        auth_tag: cipher.getAuthTag().toString("base64"),
        salt: salt.toString("base64"),
    };
}
```

A `salt` column must be added to the `user_api_keys` table. Existing rows will need re-encryption during migration.

**Remediation tier:** Immediate — block all production deploys until resolved.

bmersereau:
## Summary
- Replaces the single-SHA-256 `encryptionKey()` with HKDF (RFC 5869) and a per-row random 16-byte salt, eliminating the one-hash-breaks-all vulnerability
- Backward-compatible: rows with `salt IS NULL` still decrypt via the legacy SHA-256 path; new and updated rows automatically use HKDF
- Adds a `salt text` column to the `user_api_keys` table (nullable for compatibility with existing rows)

Closes #67

## Changes
- `backend/src/lib/userApiKeys.ts` — `encryptKey`/`decryptKey` exported helpers; `deriveKey` uses `hkdfSync` with per-row salt; `legacyKey` covers null-salt rows
- `backend/schema.sql` — `salt text` column added to `user_api_keys`; run `ALTER TABLE public.user_api_keys ADD COLUMN IF NOT EXISTS salt text;` on existing databases
- `backend/vitest.config.ts` + `backend/package.json` — vitest test runner added
- `backend/src/lib/__tests__/userApiKeys.test.ts` — 5 unit tests covering HKDF round-trip, unique salts/IVs, tamper detection, and legacy SHA-256 compat

## Test plan
- [x] Unit tests: encrypt→decrypt round-trip, unique salt per call, tamper detection, legacy null-salt backward compat

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@ next-env.d.ts
 .DS_Store
 .vercel
 coverage
+.current-issue.md
+.implementation-plan.md
+.implement-config
+CLAUDE.md
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -7,6 +7,7 @@
     "": {
       "name": "mike-backend",
       "version": "1.0.0",
+      "license": "AGPL-3.0-only",
       "dependencies": {
         "@anthropic-ai/sdk": "^0.90.0",
         "@aws-sdk/client-s3": "^3.787.0",
@@ -35,7 +36,8 @@
         "@types/node": "^22.14.1",
         "prettier": "^3.8.1",
         "tsx": "^4.19.3",
-        "typescript": "^5.8.3"
+        "typescript": "^5.8.3",
+        "vitest": "^4.1.6"
       }
     },
     "node_modules/@anthropic-ai/sdk": {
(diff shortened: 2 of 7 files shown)


willchen96:
Thanks for this. Main already uses crypto.scryptSync(), a proper password-based KDF, and issue #67  was closed as completed for that reason.

PR #77, opened 2026-05-13
fix: add expiration to download tokens (30-day TTL)

### LINKED ISSUE #68: [Security][High] Non-expiring download tokens allow persistent file access after access revocation
## Severity: High
**File:** `backend/src/lib/downloadTokens.ts:44-52`
**CWE:** CWE-613 — Insufficient Session Expiration
**OWASP:** A07:2021 — Identification and Authentication Failures

## Description
HMAC-signed download tokens carry no expiration field. Once issued and embedded in chat history, they are permanently valid:

```typescript
export function signDownload(path: string, filename: string): string {
    const payload = JSON.stringify({ p: path, f: filename });
    // No `exp` field — token never expires
```

## Impact
Users who later lose access to a project or document — through removal from `shared_with`, document deletion, or account termination — retain the ability to download files via any download URL saved from their chat history. The download route re-checks ownership at request time (providing partial mitigation), but if the signing secret is ever compromised (see #66), all tokens ever issued become immediately forgeable with no expiry to limit the window.

## Fix
Add an `exp` field to the token payload and validate it at verify time:

```typescript
export function signDownload(path: string, filename: string, ttlSeconds = 900): string {
    const exp = Math.floor(Date.now() / 1000) + ttlSeconds;
    const payload = JSON.stringify({ p: path, f: filename, exp });
    const enc = b64urlEncode(Buffer.from(payload, "utf8"));
    const sig = crypto.createHmac("sha256", getSecret()).update(enc).digest();
    return `${enc}.${b64urlEncode(sig)}`;
}

export function verifyDownload(token: string): { path: string; filename: string } | null {
    // ... existing HMAC check ...
    const parsed = JSON.parse(b64urlDecode(enc).toString("utf8"));
    if (!parsed?.p || !parsed?.f) return null;
    if (parsed.exp && Math.floor(Date.now() / 1000) > parsed.exp) return null;
    return { path: parsed.p, filename: parsed.f };
}
```

Consider whether download links in chat history should be refreshed on load (client fetches a new signed URL when rendering old messages).

**Remediation tier:** High — fix before next release.

### LINKED ISSUE #91: Nit PR #77: Missing test for backward-compat with pre-expiry tokens (no exp field)
## Found in PR #77 (fix/68-expiring-download-tokens)

**Severity:** Nit

PR #77 added token expiration and correctly handles old tokens (without `exp`) via:
```typescript
if (parsed.exp !== undefined && Math.floor(Date.now() / 1000) > parsed.exp) return null;
```

But the test suite has no test case for the backward-compat path. Production chat histories contain tokens signed before PR #77 was deployed — these tokens have no `exp` field and should still verify successfully.

**Fix:** Add a backward-compat test to `backend/src/lib/__tests__/downloadTokens.test.ts`:
```typescript
it("verifyDownload accepts a legacy token without exp field", async () => {
    const { verifyDownload } = await import("../downloadTokens.js");
    // Manually build a legacy token (no exp)
    const payload = Buffer.from(JSON.stringify({ p: "path/file.pdf", f: "file.pdf" }));
    const enc = payload.toString("base64").replace(/\+/g, "-").replace(/\//g, "_").replace(/=+$/, "");
    const sig = require("crypto")
        .createHmac("sha256", process.env.DOWNLOAD_SIGNING_SECRET!)
        .update(enc).digest("base64")
        .replace(/\+/g, "-").replace(/\//g, "_").replace(/=+$/, "");
    expect(verifyDownload(`${enc}.${sig}`)).toEqual({ path: "path/file.pdf", filename: "file.pdf" });
});
```

bmersereau:
## Summary
- Adds an `exp` (Unix timestamp) field to every HMAC-signed download token
- `verifyDownload` rejects tokens whose `exp` is in the past
- Default TTL is 30 days — generous enough for links stored in chat history, bounded enough to limit exposure if the signing secret is ever rotated
- Old tokens without an `exp` field are still accepted (backward compat)

Closes #68

## Changes
- `backend/src/lib/downloadTokens.ts` — `signDownload` gains optional `ttlSeconds` param (default 30 days); `verifyDownload` checks `exp` when present
- `backend/src/lib/__tests__/downloadTokens.test.ts` — 5 unit tests covering round-trip, tampering, exp field presence, rejection of expired tokens, and acceptance of valid tokens
- `backend/vitest.config.ts` + `backend/package.json` — vitest added

## Test plan
- [x] Unit tests: round-trip, tamper, exp field present, expired → null, valid → ok
Closes #91

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@ next-env.d.ts
 .DS_Store
 .vercel
 coverage
+.current-issue.md
+.implementation-plan.md
+.implement-config
+CLAUDE.md
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -7,6 +7,7 @@
     "": {
       "name": "mike-backend",
       "version": "1.0.0",
+      "license": "AGPL-3.0-only",
       "dependencies": {
         "@anthropic-ai/sdk": "^0.90.0",
         "@aws-sdk/client-s3": "^3.787.0",
@@ -35,7 +36,8 @@
         "@types/node": "^22.14.1",
         "prettier": "^3.8.1",
         "tsx": "^4.19.3",
-        "typescript": "^5.8.3"
+        "typescript": "^5.8.3",
+        "vitest": "^4.1.6"
       }
     },
     "node_modules/@anthropic-ai/sdk": {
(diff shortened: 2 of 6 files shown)


willchen96:
Thanks for this. Closing this as it is now superseded by the newer, focused [PR #250](https://github.com/Open-Legal-Products/mike/pull/250), which implements download-token expiry with more comprehensive tests and explicit transitional legacy-token handling.

PR #81, opened 2026-05-13
fix: remove early length exit from timingSafeEqStr

### LINKED ISSUE #72: [Security][Low] Timing oracle in download token verification — early length exit before constant-time compare
## Severity: Low
**File:** `backend/src/lib/downloadTokens.ts:39-41`
**CWE:** CWE-208 — Observable Timing Discrepancy
**OWASP:** A02:2021 — Cryptographic Failures

## Description
The HMAC comparison function exits early if the string lengths differ, before reaching the constant-time comparison:

```typescript
function timingSafeEqStr(a: string, b: string): boolean {
    if (a.length !== b.length) return false;  // early exit — timing leak
    return crypto.timingSafeEqual(Buffer.from(a), Buffer.from(b));
}
```

## Impact
A network-positioned attacker can distinguish "wrong length" responses from "correct length but wrong value" responses by measuring timing. In practice, SHA-256 always produces 32 bytes → 43-char base64url, so this leaks no useful information for standard token forgery attempts. The risk is low and theoretical, but the pattern violates the contract of a constant-time comparison and should be corrected.

## Fix
Remove the early length check. Perform the comparison unconditionally, padding to equal length:

```typescript
function timingSafeEqStr(a: string, b: string): boolean {
    const aBuf = Buffer.from(a);
    const bBuf = Buffer.alloc(aBuf.length);
    Buffer.from(b).copy(bBuf);
    // Compare in constant time; also verify lengths match
    return crypto.timingSafeEqual(aBuf, bBuf) && a.length === b.length;
}
```

**Remediation tier:** Backlog — low priority, no immediate exploit path.

bmersereau:
## Summary
- `timingSafeEqStr` returned `false` immediately when string lengths differed, before calling `crypto.timingSafeEqual` — violating the constant-time contract and creating a theoretical timing oracle
- Fixed by always running the constant-time comparison first (on zero-padded buffers), then checking length equality afterward
- In practice, all SHA-256 base64url signatures are 43 characters so this gap has no real exploit path for standard tokens; this is a defense-in-depth fix for the pattern

Closes #72

## Changes
- `backend/src/lib/downloadTokens.ts` — early length check removed; `timingSafeEqStr` now pads the shorter buffer and runs `timingSafeEqual` unconditionally before length verification; exported for testability
- `backend/src/lib/__tests__/timingSafeEq.test.ts` — 5 unit tests

## Test plan
- [x] Unit tests: equal strings, same-length mismatch, different-length mismatch, empty strings

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@ next-env.d.ts
 .DS_Store
 .vercel
 coverage
+.current-issue.md
+.implementation-plan.md
+.implement-config
+CLAUDE.md
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -7,6 +7,7 @@
     "": {
       "name": "mike-backend",
       "version": "1.0.0",
+      "license": "AGPL-3.0-only",
       "dependencies": {
         "@anthropic-ai/sdk": "^0.90.0",
         "@aws-sdk/client-s3": "^3.787.0",
@@ -35,7 +36,8 @@
         "@types/node": "^22.14.1",
         "prettier": "^3.8.1",
         "tsx": "^4.19.3",
-        "typescript": "^5.8.3"
+        "typescript": "^5.8.3",
+        "vitest": "^4.1.6"
       }
     },
     "node_modules/@anthropic-ai/sdk": {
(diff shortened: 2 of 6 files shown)


willchen96:
Closing as superseded by #250, which contains the current download-token implementation with timing-safe signature comparison and expiry coverage. Issue #72 remains open until the replacement lands.

PR #106, opened 2026-05-13
fix: throw on missing env vars in backend createServerSupabase()

### LINKED ISSUE #97: [Security][High] Backend createServerSupabase() silently creates broken client when env vars are missing
## Problem

`backend/src/lib/supabase.ts:7–11` uses `|| ""` fallbacks for both env vars:

```ts
export function createServerSupabase() {
  const url = process.env.SUPABASE_URL || "";
  const key = process.env.SUPABASE_SECRET_KEY || "";
  return createClient(url, key, { auth: { persistSession: false } });
}
```

If `SUPABASE_URL` or `SUPABASE_SECRET_KEY` is unset, this returns a client configured with empty strings. Every subsequent DB call silently fails or returns unexpected data. There is no startup guard, no thrown error, and no log entry.

## Impact

Any route that calls `createServerSupabase()` will silently malfunction rather than failing fast. This makes misconfiguration very hard to diagnose.

## Parity gap

The frontend equivalent (`frontend/src/lib/supabase-server.ts`) was fixed in PR #73 to throw:

```ts
if (!url || !key) {
    throw new Error("NEXT_PUBLIC_SUPABASE_URL and SUPABASE_SECRET_KEY must be set");
}
```

The backend needs the same treatment.

## Fix

Throw `new Error("SUPABASE_URL and SUPABASE_SECRET_KEY must be set")` when either env var is absent.

### LINKED ISSUE #114: fix: add 'test' script to backend/package.json (missing in PRs #106, #107, #110, #111, #112, #113)
## Problem

## Fix required

```json
"test": "vitest run"
```

## Affected PRs
- #106 (fix/97)
- #107 (fix/98)
- #110 (fix/105)
- #111 (fix/99)
- #112 (fix/100)
- #113 (fix/104)

### LINKED ISSUE #117: vitest config: missing include filter causes dist/ compiled test artifacts to run (PRs #106, #107, #112)
## Problem

Three PRs ship a `vitest.config.ts` without an explicit `include` filter pointing only at `src/`:

| PR | vitest.config.ts include | tsconfig excludes __tests__ |
|----|--------------------------|----------------------------|
| #106 (fix/97) | none (uses vitest default) | No |
| #107 (fix/98) | none (no vitest.config.ts at all) | No |
| #112 (fix/100) | none (uses vitest default) | Yes (added) |

Because `tsconfig.json` on PRs #106 and #107 does **not** exclude `src/**/__tests__`, running `npm run build` compiles the test files into `dist/`. Vitest's default include glob (`**/*.{test,spec}.{js,ts,...}`) then picks up both the source `*.test.ts` files **and** the compiled `dist/**/*.test.js` files.

The compiled CJS output fails with:

```
Error: Vitest cannot be imported in a CommonJS module using require().
```

This causes `npm run build && npm test` to report "1 file failed" even though all actual test assertions pass.

## Impact

- Does **not** affect a clean CI run (no pre-existing `dist/`)
- Does affect any developer who runs `npm run build` before `npm test`
- The PR description claims "Unit tests added and passing" which is incorrect in a post-build environment

## Fix

In each affected PR's `vitest.config.ts`, add:

```ts
test: {
  include: ["src/**/__tests__/**/*.test.ts"],
  exclude: ["dist/**", "node_modules/**"],
}
```

And in `tsconfig.json`, add `"src/**/__tests__/**"` to `exclude`.

## Affected PRs
- PR #106 (fix/97)
- PR #107 (fix/98)  
- PR #112 (fix/100)

### LINKED ISSUE #124: PR #106: getUserIdFromRequest in supabase.ts still uses || "" pattern and creates a new client per call — dead code, should be removed
## Problem

During review of PR #106, the batch reviewer identified that `getUserIdFromRequest` in `backend/src/lib/supabase.ts` was not updated as part of the env-guard fix. The function:

- Still uses the old `|| ""` pattern for env vars
- Creates a fresh `createClient` on every call (no singleton)
- Has **zero callers** in the codebase — it is dead code

Since it is exported, leaving it in place risks it being wired to a route in the future with broken behavior.

## Fix

Remove `getUserIdFromRequest` from `supabase.ts` entirely. Auth is handled by `requireAuth` middleware (`auth.ts`) which already uses the singleton pattern.

## Affected

- Branch: `fix/97-backend-supabase-env-guard`
- PR: #106

bmersereau:
## Summary
- `backend/src/lib/supabase.ts` now throws `Error("SUPABASE_URL and SUPABASE_SECRET_KEY must be set")` when either env var is absent, matching the frontend fix from PR #73
- Removes dead `getUserIdFromRequest` export (no callers) which still used the old `|| ""` pattern

Closes #97
Closes #114
Closes #117
Closes #124

## Changes
- `backend/src/lib/supabase.ts` — replace `|| ""` fallbacks with explicit guard; remove dead `getUserIdFromRequest` export
- `backend/src/lib/__tests__/supabaseServer.test.ts` — new test: throws on missing URL, missing key, both missing; succeeds when both set
- `backend/vitest.config.ts` — vitest configuration with `include` filter and `isolate: true`; `"test"` script added to `package.json`

## Test plan
- [x] Unit tests added and passing (4/4)
- [x] Backend build passes

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -7,6 +7,7 @@
     "": {
       "name": "mike-backend",
       "version": "1.0.0",
+      "license": "AGPL-3.0-only",
       "dependencies": {
         "@anthropic-ai/sdk": "^0.90.0",
         "@aws-sdk/client-s3": "^3.787.0",
@@ -35,7 +36,8 @@
         "@types/node": "^22.14.1",
         "prettier": "^3.8.1",
         "tsx": "^4.19.3",
-        "typescript": "^5.8.3"
+        "typescript": "^5.8.3",
+        "vitest": "^4.1.6"
       }
     },
     "node_modules/@anthropic-ai/sdk": {
@@ -972,6 +974,40 @@
         "node": ">=6.9.0"
       }
     },
+    "node_modules/@emnapi/core": {
+      "version": "1.10.0",
+      "resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.10.0.tgz",
+      "integrity": "sha512-yq6OkJ4p82CAfPl0u9mQebQHKPJkY7WrIuk205cTYnYe+k2Z8YBh11FrbRG/H6ihirqcacOgl2BIO8oyMQLeXw==",
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "dependencies": {
+        "@emnapi/wasi-threads": "1.2.1",
+        "tslib": "^2.4.0"
+      }
+    },
+    "node_modules/@emnapi/runtime": {
+      "version": "1.10.0",
+      "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.10.0.tgz",
+      "integrity": "sha512-ewvYlk86xUoGI0zQRNq/mC+16R1QeDlKQy21Ki3oSYXNgLb45GV1P6A0M+/s6nyCuNDqe5VpaY84BzXGwVbwFA==",
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "dependencies": {
+        "tslib": "^2.4.0"
+      }
+    },
+    "node_modules/@emnapi/wasi-threads": {
+      "version": "1.2.1",
+      "resolved": "https://registry.npmjs.org/@emnapi/wasi-threads/-/wasi-threads-1.2.1.tgz",
+      "integrity": "sha512-uTII7OYF+/Mes/MrcIOYp5yOtSMLBWSIoLPpcgwipoiKbli6k322tcoFsxoIIxPDqW01SQGAgko4EzZi2BNv2w==",
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "dependencies": {
+        "tslib": "^2.4.0"
+      }
+    },
     "node_modules/@esbuild/aix-ppc64": {
       "version": "0.27.7",
       "resolved": "https://registry.npmjs.org/@esbuild/aix-ppc64/-/aix-ppc64-0.27.7.tgz",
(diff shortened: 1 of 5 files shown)


bmersereau (commented):
## PR Review

### Summary
Fixes a silent-fail bug in `createServerSupabase()` where missing env vars produced a broken client with empty-string credentials instead of throwing. The fix is correct, minimal, and well-tested. One pre-existing function in the same file (`getUserIdFromRequest`) still uses the old `|| ""` pattern, but that function is currently dead code (no callers) and is out of scope for this PR.

### Findings
- [severity:low] `getUserIdFromRequest` in `supabase.ts` (lines 29-30) still uses the old `|| ""` fallback pattern and creates a new `createClient` on every invocation — both issues this PR fixed for `createServerSupabase`. The function has no callers today, but should be cleaned up before it's wired to a route.
- [severity:nit] Tests use dynamic `import()` inside each `it` block with `isolate: true` in vitest config. This works, but `vi.resetModules()` in a `beforeEach` would make the isolation intent explicit and guard against accidental module-cache sharing if `isolate` is ever removed.

### Verdict
Approve with nits

### What I verified
- Tests: pass (4/4)
- vitest.config.ts has `include` filter: pass (`"src/**/__tests__/**/*.test.ts"`)
- package.json has `"test": "vitest run"`: pass
- `createServerSupabase` throws on missing URL, missing key, both missing, and returns client when both set: pass

bmersereau (commented):
## PR Review

### Summary
Replaces the silent `|| ""` fallbacks in `createServerSupabase()` with an explicit guard that throws when either env var is absent. Four unit tests cover all combinations of missing/present env vars. Simple, correct, well-tested change.

### Findings
- [severity:praise] Error message `"SUPABASE_URL and SUPABASE_SECRET_KEY must be set"` is clear and actionable
- [severity:praise] `persistSession: false` already present — service-role client correctly stateless
- [severity:nit] Tests import via dynamic `await import("../supabase.js")` without module reset between tests — env var deletions in `beforeEach` may not fully isolate if the module is cached. Tests still pass because vitest's `isolate: true` in vitest.config.ts handles this

### Specific checks
- [x] Tests pass: 4/4 ✓
- [x] documents.ts unchanged from origin/main ✓

### Verdict
**Approve** — clean.

bmersereau (commented):
## Follow-up Review (pass 2)

### Change
Removed dead `getUserIdFromRequest` export — function had no callers and still used the old `|| ""` pattern that this PR was eliminating.

### Findings
- [severity:praise] File is now clean — only `createServerSupabase` remains, with the correct guard
- No regressions. Tests still pass 4/4.

### Verdict
**Approve** — all pass-1 findings resolved.

willchen96:
Thanks for the contribution. Already implemented. Current backend/src/lib/supabase.ts throws when SUPABASE_URL or SUPABASE_SECRET_KEY is missing, matching the PR’s fix.

PR #74, opened 2026-05-13
fix: require dedicated secrets for download signing and API key encryption

bmersereau:
## Summary
- Remove `SUPABASE_SECRET_KEY` fallback from `downloadTokens.getSecret()` — `DOWNLOAD_SIGNING_SECRET` is now required
- Remove three-way fallback from `userApiKeys.encryptionKey()` — `USER_API_KEYS_ENCRYPTION_SECRET` is now required
- Add `assertSecretIsolation()` called at server startup to enforce presence, uniqueness, and cross-secret isolation of each secret
- Updates JSDoc comment in `downloadTokens.ts` to reflect the TTL-based expiration added in PR #77

Closes #66
Closes #82
Closes #83
Closes #88

## Changes
- `backend/src/lib/downloadTokens.ts` — `getSecret()` reads only `DOWNLOAD_SIGNING_SECRET`; JSDoc updated to reference configurable TTL
- `backend/src/lib/userApiKeys.ts` — `encryptionKey()` reads only `USER_API_KEYS_ENCRYPTION_SECRET`
- `backend/src/lib/startup.ts` — new `assertSecretIsolation()` validates at boot: missing vars throw, secrets matching `SUPABASE_SECRET_KEY` throw, and `DOWNLOAD_SIGNING_SECRET === USER_API_KEYS_ENCRYPTION_SECRET` throws
- `backend/src/index.ts` — call `assertSecretIsolation()` before the server starts
- `backend/vitest.config.ts` — vitest config excluding `dist/`
- `backend/src/lib/__tests__/secretIsolation.test.ts` — 11 vitest unit tests covering all paths

## Test plan
- [x] `npm test` passes (11/11) in `backend/`
- [x] `npm run build` passes with no TypeScript errors
- [x] `.env` must be updated before restarting backend (see migration note below)

## ⚠️ Migration note

Before restarting the backend after this change, ensure your environment has:

```
DOWNLOAD_SIGNING_SECRET=<dedicated random value>
USER_API_KEYS_ENCRYPTION_SECRET=<dedicated random value>
```

Generate fresh values with `openssl rand -hex 32`.

**If you previously had `API_KEYS_ENCRYPTION_SECRET` set** (an undocumented legacy fallback that was removed in this PR): set `USER_API_KEYS_ENCRYPTION_SECRET` to that **same value**. Do not generate a new value — doing so will change the AES key and make all stored user API keys in the database unreadable.

Issue #66, opened 2026-05-13
[Security][Critical] SUPABASE_SECRET_KEY triple-reuse — single key leak = total system compromise


bmersereau:
## Severity: Critical
**Files:** `backend/src/lib/downloadTokens.ts:14-15`, `backend/src/lib/userApiKeys.ts:40-43`
**CWE:** CWE-798 — Use of Hard-coded Credentials
**OWASP:** A02:2021 — Cryptographic Failures

## Description
`SUPABASE_SECRET_KEY` serves as fallback for three independent secrets simultaneously:

1. Supabase admin credential (full DB bypass, no RLS)
2. HMAC signing secret for download tokens (`downloadTokens.ts:14-15`)
3. AES-256-GCM key derivation secret for stored user API keys (`userApiKeys.ts:40-43`)

```typescript
// downloadTokens.ts
const secret = process.env.DOWNLOAD_SIGNING_SECRET ?? process.env.SUPABASE_SECRET_KEY;

// userApiKeys.ts
const secret = process.env.USER_API_KEYS_ENCRYPTION_SECRET
    || process.env.API_KEYS_ENCRYPTION_SECRET
    || process.env.SUPABASE_SECRET_KEY;
```

## Impact
A single service-role key leak (via `.env` commit, CI log, container image layer, or Supabase dashboard breach) simultaneously grants:
- Ability to **forge download tokens** for any R2 storage path
- Ability to **decrypt every stored Anthropic/OpenAI/Gemini API key** in the database
- **Full Supabase admin DB access** bypassing all RLS policies

## Fix

**Step 1** — Remove the fallback. Require distinct secrets:
```typescript
// downloadTokens.ts
function getSecret(): string {
    const secret = process.env.DOWNLOAD_SIGNING_SECRET;
    if (!secret) throw new Error("DOWNLOAD_SIGNING_SECRET must be set");
    return secret;
}

// userApiKeys.ts
function encryptionKey(): Buffer {
    const secret = process.env.USER_API_KEYS_ENCRYPTION_SECRET;
    if (!secret) throw new Error("USER_API_KEYS_ENCRYPTION_SECRET must be set");
    return crypto.createHash("sha256").update(secret).digest();
}
```

**Step 2** — Add startup assertions in `backend/src/index.ts` (fail fast on bad config):
```typescript
function assertRequiredEnv(): void {
    const required = [
        "SUPABASE_URL", "SUPABASE_SECRET_KEY",
[...]

PR #107, opened 2026-05-13
fix: remove SUPABASE_SECRET_KEY fallback from encryptionKey()

### LINKED ISSUE #98: [Security][High] encryptionKey() retains SUPABASE_SECRET_KEY fallback — secret isolation not enforced at the source
## Problem

`backend/src/lib/userApiKeys.ts:40–47`:

```ts
function encryptionKey(): Buffer {
  const secret =
    process.env.USER_API_KEYS_ENCRYPTION_SECRET ||
    process.env.API_KEYS_ENCRYPTION_SECRET ||
    process.env.SUPABASE_SECRET_KEY;  // ← last-resort fallback
  ...
}
```

`assertSecretIsolation()` (PR #74) prevents the server from starting when secrets overlap — but it only fires at startup. The `encryptionKey()` function itself still accepts the Supabase service key as a fallback. If the isolation check is ever not wired (e.g., a different entrypoint, a test environment), user API keys would be encrypted under the same secret used for JWT verification.

## Impact

A single `SUPABASE_SECRET_KEY` leak would expose all stored user LLM API keys. PR #74 added a startup check but did not remove the root cause.

## Fix

Remove the `SUPABASE_SECRET_KEY` fallback from `encryptionKey()`. Require `USER_API_KEYS_ENCRYPTION_SECRET` unconditionally — throw `new Error("USER_API_KEYS_ENCRYPTION_SECRET must be set")` if absent.

### LINKED ISSUE #114: fix: add 'test' script to backend/package.json (missing in PRs #106, #107, #110, #111, #112, #113)

### LINKED ISSUE #117: vitest config: missing include filter causes dist/ compiled test artifacts to run (PRs #106, #107, #112)

### LINKED ISSUE #118: encryptionKey.test.ts: vi.mock calls inside it() will break in a future vitest version
## Problem

In `backend/src/lib/__tests__/encryptionKey.test.ts` (PR #107 / fix/98), `vi.mock()` calls appear inside individual `it()` callbacks rather than at the top level of the module:

```typescript
it("throws when no encryption secret is configured", async () => {
    vi.mock("../supabase.js", () => ({ ... }));  // ← nested inside it()
    ...
});
```

Vitest currently hoists all `vi.mock()` calls to before the test suite runs regardless of where they appear (this is the Babel/vite-plugin-vitest transform behaviour). The existing vitest run emits a warning:

```
Warning: A vi.mock("../supabase.js") call in "...encryptionKey.test.ts" is not at the top level of the module.
Although it appears nested, it will be hoisted and executed before any tests run.
Move it to the top level to reflect its actual execution order.
This will become an error in a future version.
```

Because all three `vi.mock` calls mock the same module identity with the same factory, the tests currently pass. But the warning explicitly states this **will become an error** in a future vitest release, and the nesting gives a false impression that the mock is scoped to the individual test.

## Fix

Move the `vi.mock("../supabase.js", ...))` call to the top level of the file (after imports, before the `describe` block), and use `vi.mocked(...).mockImplementation()` or `mockReturnValue()` inside each `it()` if different behaviour per test is needed.

bmersereau:
## Summary
- Removes `process.env.SUPABASE_SECRET_KEY` from the fallback chain in `encryptionKey()` in `backend/src/lib/userApiKeys.ts`
- `USER_API_KEYS_ENCRYPTION_SECRET` (or its alias `API_KEYS_ENCRYPTION_SECRET`) is now required; the function throws if neither is set
- Previously, a missing dedicated encryption secret silently fell back to the JWT verification secret, violating secret isolation (PR #74 / issue #66)

Closes #98
Closes #114
Closes #117
Closes #118

## Changes
- `backend/src/lib/userApiKeys.ts` — remove `SUPABASE_SECRET_KEY` from encryptionKey() fallback chain
- `backend/src/lib/__tests__/encryptionKey.test.ts` — tests that SUPABASE_SECRET_KEY fallback is gone and dedicated secret is required; `vi.mock` hoisted to module top-level
- `backend/package.json` / `backend/package-lock.json` — add `vitest` dev dependency and `"test": "vitest run"` script
- `backend/vitest.config.ts` — include filter scoping tests to `src/` to exclude compiled `dist/` artifacts

## Test plan
- [x] Unit tests added and passing (3/3)
- [x] Backend build passes

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -7,6 +7,7 @@
     "": {
       "name": "mike-backend",
       "version": "1.0.0",
+      "license": "AGPL-3.0-only",
       "dependencies": {
         "@anthropic-ai/sdk": "^0.90.0",
         "@aws-sdk/client-s3": "^3.787.0",
@@ -35,7 +36,8 @@
         "@types/node": "^22.14.1",
         "prettier": "^3.8.1",
         "tsx": "^4.19.3",
-        "typescript": "^5.8.3"
+        "typescript": "^5.8.3",
+        "vitest": "^4.1.6"
       }
     },
     "node_modules/@anthropic-ai/sdk": {
@@ -972,6 +974,40 @@
         "node": ">=6.9.0"
       }
     },
+    "node_modules/@emnapi/core": {
+      "version": "1.10.0",
+      "resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.10.0.tgz",
+      "integrity": "sha512-yq6OkJ4p82CAfPl0u9mQebQHKPJkY7WrIuk205cTYnYe+k2Z8YBh11FrbRG/H6ihirqcacOgl2BIO8oyMQLeXw==",
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "dependencies": {
+        "@emnapi/wasi-threads": "1.2.1",
+        "tslib": "^2.4.0"
+      }
+    },
+    "node_modules/@emnapi/runtime": {
+      "version": "1.10.0",
+      "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.10.0.tgz",
+      "integrity": "sha512-ewvYlk86xUoGI0zQRNq/mC+16R1QeDlKQy21Ki3oSYXNgLb45GV1P6A0M+/s6nyCuNDqe5VpaY84BzXGwVbwFA==",
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "dependencies": {
+        "tslib": "^2.4.0"
+      }
+    },
+    "node_modules/@emnapi/wasi-threads": {
+      "version": "1.2.1",
+      "resolved": "https://registry.npmjs.org/@emnapi/wasi-threads/-/wasi-threads-1.2.1.tgz",
+      "integrity": "sha512-uTII7OYF+/Mes/MrcIOYp5yOtSMLBWSIoLPpcgwipoiKbli6k322tcoFsxoIIxPDqW01SQGAgko4EzZi2BNv2w==",
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "dependencies": {
+        "tslib": "^2.4.0"
+      }
+    },
     "node_modules/@esbuild/aix-ppc64": {
       "version": "0.27.7",
       "resolved": "https://registry.npmjs.org/@esbuild/aix-ppc64/-/aix-ppc64-0.27.7.tgz",
@@ -1437,6 +1473,13 @@
         }
       }
     },
+    "node_modules/@jridgewell/sourcemap-codec": {
+      "version": "1.5.5",
+      "resolved": "https://registry.npmjs.org/@jridgewell/sourcemap-codec/-/sourcemap-codec-1.5.5.tgz",
+      "integrity": "sha512-cYQ9310grqxueWbl+WuIUIaiUaDcj7WOq5fVhEljNVgRfOUhY9fy2zTvfoqWsnebh8Sl70VScFbICvJnLKB0Og==",
+      "dev": true,
+      "license": "MIT"
+    },
     "node_modules/@napi-rs/canvas": {
       "version": "0.1.97",
       "resolved": "https://registry.npmjs.org/@napi-rs/canvas/-/canvas-0.1.97.tgz",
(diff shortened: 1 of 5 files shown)


bmersereau (commented):
## PR Review

### Summary
Removes the `SUPABASE_SECRET_KEY` fallback from `encryptionKey()` in `userApiKeys.ts`, closing a secret-isolation vulnerability where the JWT verification key could silently double as the AES-256-GCM encryption key when `USER_API_KEYS_ENCRYPTION_SECRET` was not set. The fix is correct and the test suite explicitly verifies the removal. `vi.mock` is correctly hoisted at module top-level (not inside `describe`/`it`).

### Findings
- [severity:low] `vi.resetModules()` in `beforeEach` re-instantiates the module under test on each dynamic `import()`, which is correct. However, `vi.resetModules()` combined with a top-level `vi.mock` depends on Vitest's mock-registry persistence across module resets — this currently works as expected (tests pass), but it's a subtle interaction worth noting in a comment so future maintainers don't remove the top-level `vi.mock`.
- [severity:nit] `beforeEach` cleans `USER_API_KEYS_ENCRYPTION_SECRET`, `API_KEYS_ENCRYPTION_SECRET`, and `SUPABASE_SECRET_KEY` from the environment. There is no corresponding `afterEach` to restore them. If env vars were set in CI before this test runs, the `beforeEach` delete may interfere with subsequent test files. Low risk since `isolate: true` is set, but worth adding `afterEach` cleanup for hygiene.
- [severity:nit] The `API_KEYS_ENCRYPTION_SECRET` fallback is retained (only `SUPABASE_SECRET_KEY` is removed). If `API_KEYS_ENCRYPTION_SECRET` is a deprecated alias for `USER_API_KEYS_ENCRYPTION_SECRET`, a comment or deprecation notice would help operators know which to use.

### Verdict
Approve with nits

### What I verified
- Tests: pass (3/3)
- vitest.config.ts has `include` filter: pass (`"src/**/__tests__/**/*.test.ts"`)
- package.json has `"test": "vitest run"`: pass
- `vi.mock` is at module top-level (not inside `describe`/`it`): pass
- `SUPABASE_SECRET_KEY` removed from fallback chain in `userApiKeys.ts`: pass
- Test explicitly covers SUPABASE_SECRET_KEY-only case now throwing: pass

bmersereau (commented):
## PR Review

### Summary
Removes `SUPABASE_SECRET_KEY` from the `encryptionKey()` fallback chain in `userApiKeys.ts`. Three tests verify the SUPABASE_SECRET_KEY fallback is gone and that a dedicated secret is required. `vi.mock` is correctly hoisted to module top-level (issue #118 resolved).

### Findings
- [severity:praise] `vi.mock` at module top-level, before `describe` — hoisting semantics correct ✓
- [severity:praise] `API_KEYS_ENCRYPTION_SECRET` alias retained for backward-compat with existing deployments that set the old name
- [severity:nit] Test name `"throws when only SUPABASE_SECRET_KEY is set (fallback removed)"` is slightly verbose but clear

### Specific checks
- [x] `vi.mock` is top-level ✓ (issue #118)
- [x] Tests pass: 3/3 ✓

### Verdict
**Approve** — clean.

willchen96:
Thanks Beau, this has been fixed in a more targeted PR #137 without the test script

PR #110, opened 2026-05-13
fix: paginate GET /chat (limit + before cursor)

### LINKED ISSUE #105: [Low] GET /chat returns all chats without pagination
## Problem

`backend/src/routes/chat.ts:141–166` — the `GET /chat` endpoint fetches every chat the user owns plus every chat in every project they own, with no limit:

```ts
const { data, error } = await db
  .from("chats")
  .select("*")
  .or(filter)
  .order("created_at", { ascending: false });
// no .limit()
```

For an active user with hundreds of chats (or an owner of a busy shared project), this is a large unbounded response on every page load, blocking the sidebar render.

## Fix

Add `limit` / cursor-based pagination. A default of 50 chats per page is reasonable. The sidebar already renders a list — it can lazy-load older chats on scroll.

```ts
const limit = Math.min(Number(req.query.limit) || 50, 200);
const before = req.query.before as string | undefined;  // ISO timestamp cursor

// ...
let q = db.from("chats").select("*").or(filter).order("created_at", { ascending: false }).limit(limit);
if (before) q = q.lt("created_at", before);
```

### LINKED ISSUE #114: fix: add 'test' script to backend/package.json (missing in PRs #106, #107, #110, #111, #112, #113)

bmersereau:
## Summary
- Adds `limit` (default 50, max 200) and `before` (ISO timestamp cursor) query parameters to `GET /chat`
- Prevents unbounded chat list queries for active users with large chat histories
- Backward-compatible: existing callers with no query params get the first 50 chats (most recent)

Closes #105
Closes #114

## Changes
- `backend/src/routes/chat.ts` — limit + before-cursor pagination on the chat list query
- `backend/src/lib/__tests__/chatPagination.test.ts` — static analysis tests verifying pagination is present
- `backend/tsconfig.json` — exclude `__tests__` dirs from the main tsc build (tests use vitest directly)
- `backend/package.json` — `"test": "vitest run"` script added

## Test plan
- [x] Unit tests added and passing (`npx vitest run`)
- [x] Backend type check passes (`npx tsc --noEmit`)

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -5,7 +5,8 @@
   "scripts": {
     "dev": "tsx watch src/index.ts",
     "build": "tsc",
-    "start": "node dist/index.js"
+    "start": "node dist/index.js",
+    "test": "vitest run"
   },
   "dependencies": {
     "@anthropic-ai/sdk": "^0.90.0",
--- a/backend/src/lib/__tests__/chatPagination.test.ts
+++ b/backend/src/lib/__tests__/chatPagination.test.ts
@@ -1,0 +1,18 @@
+import { readFileSync } from "fs";
+import { join } from "path";
+import { describe, it, expect } from "vitest";
+
+describe("GET /chat pagination", () => {
+    const src = readFileSync(
+        join(__dirname, "../../routes/chat.ts"),
+        "utf8",
+    );
+
+    it("applies a limit to the chat list query", () => {
+        expect(src).toMatch(/\.limit\(/);
+    });
+
+    it("supports a before-cursor for pagination", () => {
+        expect(src).toMatch(/before|lt\("created_at"/);
+    });
+});
--- a/backend/src/routes/chat.ts
+++ b/backend/src/routes/chat.ts
@@ -142,6 +142,15 @@ chatRouter.get("/", requireAuth, async (req, res) => {
     const userId = res.locals.userId as string;
     const db = createServerSupabase();
 
+    // Parse pagination params
+    const rawLimit = Number(req.query.limit);
+    const limit = Number.isFinite(rawLimit) && rawLimit > 0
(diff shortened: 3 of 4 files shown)


bmersereau (commented):
## PR Review

### Summary
Adds cursor-based pagination (`limit` + `before` ISO timestamp) to `GET /chat`. Implementation is correct and backward-compatible — callers with no params get 50 chats. The `limit` cap at 200 is reasonable. Two static-analysis tests verify the limit and cursor are present.

### Findings
- [severity:praise] `before` cursor validated with `Number.isNaN(new Date(...).getTime())` — correctly rejects non-ISO strings while accepting valid timestamps
- [severity:nit] The `before` cursor is applied after the `.limit()` call, meaning Supabase will apply `lt("created_at", before)` as a WHERE clause and `limit` as LIMIT — correct behavior, but a comment clarifying cursor-vs-offset semantics would help future readers
- [severity:minor] Static tests use `readFileSync` on the source file; they verify the pattern is present but don't test the actual endpoint behavior (e.g., correct rows returned). Acceptable for now, but a true integration test would give higher confidence

### Specific checks
- [x] Tests pass: 2/2 ✓
- [x] `tsconfig.json` excludes `__tests__` from tsc build ✓

### Verdict
**Approve with nits** — ship it.

willchen96:
Thanks Beau. Main rewrote chat listing as an RPC (get_chats_overview) with a limit param. Once again closing this. At the time of these PRs I was doing a large scale rewrite of the codebase. Apologies.

PR #133, opened 2026-05-14
docs: clarify safe local env setup

b1rdmania:
## Summary

- surface the existing safe local testing guidance earlier in the README
- recommend disposable Supabase/R2/model-provider resources and synthetic documents for first runs
- split frontend env examples into browser-safe public vars and server-side Next runtime vars
- clarify that SUPABASE_SECRET_KEY must never be exposed as a NEXT_PUBLIC_* value

## Validation

- git diff --check

Documentation-only change.

### DIFF
--- a/frontend/.env.local.example
+++ b/frontend/.env.local.example
@@ -1,4 +1,6 @@
 NEXT_PUBLIC_SUPABASE_URL=https://your-project.supabase.co
 NEXT_PUBLIC_SUPABASE_PUBLISHABLE_DEFAULT_KEY=your-supabase-anon-key
-SUPABASE_SECRET_KEY=your-supabase-service-role-key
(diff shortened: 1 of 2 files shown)


frontend/.env.local.example
```diff
@@ -1,4 +1,6 @@
 NEXT_PUBLIC_SUPABASE_URL=https://your-project.supabase.co
 NEXT_PUBLIC_SUPABASE_PUBLISHABLE_DEFAULT_KEY=your-supabase-anon-key
-SUPABASE_SECRET_KEY=your-supabase-service-role-key
```
willchen96, line 3: No longer in use and to be removed

willchen96:
Thanks for the contribution. I will add a note on safe local testing in a separate PR. However the supabase secret key is no longer in use in frontend and will be removed from the the example file and readme in a new PR

PR #143, opened 2026-05-16
fix: cache s3 client construction

a1exxd0:
# Summary

Caching the S3 client in `backend/src/lib/storage.ts`.

# What

Replaced `getClient()`'s `new S3Client(...)` call with module-level cached client. 

# Why

Previously, every `uploadFile`, `downloadFile`, `deleteFile` and `getSignedUrl` call constructed a new client. This is wasteful as config is static.

# Testing

Local runs only. Not sure if this repo is missing tests at the time of writing, happy to be corrected. Also ran a Claude Code `/security-review`, no high-confidence issues.


willchen96:
Thanks Alex for spotting this. I ported this change over manually and added an error where the s3 env variables are not set.

PR #149, opened 2026-05-18
feat: add Docker and Docker Compose support for self-hosting

echomohit:
## Summary

- Adds `backend/Dockerfile` and `frontend/Dockerfile` for both services
- Adds `docker-compose.yml` at the repo root to wire them together
- Adds `.env.example` documenting the three build-time vars needed by Compose
- Sets `output: "standalone"` in `next.config.ts` for a smaller frontend runtime image

## Why

Makes self-hosting straightforward — the full stack can be built and started with a single command:

```bash
docker compose up --build
```

No platform-specific tooling required. The deployment still uses the existing Supabase and Cloudflare R2 infrastructure — this is purely about packaging and running the app, not replacing any services.

## Changes

- **`backend/Dockerfile`**: multi-stage build; installs LibreOffice in the runner stage (required by `libreoffice-convert`), compiles TypeScript and runs `node dist/index.js`
- **`frontend/Dockerfile`**: multi-stage build; `NEXT_PUBLIC_*` vars passed as build args (baked in at build time by Next.js), copies only the standalone output for a minimal image
- **`docker-compose.yml`**: exposes backend on 3001, frontend on 3000; frontend depends on backend; both load env from their respective `.env` files
- **`.env.example`**: documents the root-level build args for Compose
- **`frontend/next.config.ts`**: adds `output: "standalone"` — required for the optimized Docker image, no effect on the existing Cloudflare deployment path

## Testing

Built and ran locally with both containers healthy:
- Backend: `http://localhost:3001/health` returns `{"ok":true}`
- Frontend: `http://localhost:3000` serves the app

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,5 @@
+# Build-time args for the frontend image (baked into the Next.js bundle)
+NEXT_PUBLIC_SUPABASE_URL=https://your-project.supabase.co
+NEXT_PUBLIC_SUPABASE_PUBLISHABLE_DEFAULT_KEY=your-supabase-anon-key
+# URL the browser uses to reach the backend — must be publicly reachable
+NEXT_PUBLIC_API_BASE_URL=http://localhost:3001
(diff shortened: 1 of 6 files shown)


willchen96:
Thanks for this, apologies for not getting to this. Someone else has implemented a new version.

PR #155, opened 2026-05-19
fix: validate request body in POST /projects/:projectId/chat

Shawnaldinho:
## Summary
[`projectChat.ts:32-39`](https://github.com/willchen96/mike/blob/main/backend/src/routes/projectChat.ts#L32-L39) reads \`req.body\` via a bare \`as { ... }\` cast, while sibling routes (\`chat.ts\`, \`user.ts\`) use hand-written parsers. A misshapen body — wrong types for \`messages\`, \`displayed_doc\`, \`attached_documents\` — therefore reaches the DB writes and the LLM call.

## Changes
- Add per-field parsers in \`backend/src/routes/projectChat.ts\` matching the \`parseFoo(value: unknown): { ok: true; ... } | { ok: false; detail }\` pattern already used by \`routes/chat.ts\`: \`parseChatMessages\`, \`parseOptionalChatId\`, \`parseOptionalModel\`, \`parseDocRef\`, \`parseOptionalDisplayedDoc\`, \`parseOptionalAttachedDocuments\`.
- Validate every field before any DB or LLM work; return \`400\` with the parser's \`detail\` message on the first failure.
- Downstream variable names (\`messages\`, \`chat_id\`, \`model\`, \`displayed_doc\`, \`attached_documents\`) are preserved so the rest of the handler is unchanged.

## Why
Concrete failures the previous cast permitted:
- \`messages\` as a non-array → \`[...messages].reverse()\` throws inside the handler after the chat row has already been inserted.
- \`attached_documents\` containing non-object entries → \`.map((d) => ...)\` crashes on \`d.document_id\`.
- \`displayed_doc\` as a string → the route builds a malformed user message and sends it to the model.

Matches the existing style rather than introducing a new validation library, per CONTRIBUTING.md guidance to keep changes targeted.

## Testing
- Walked each parser through the previously crash-inducing shapes and confirmed each now returns a 400 with a useful \`detail\`.

Addresses the input-validation concern from https://insights.flank.ai/where-mikeoss-falls-short.html (gap 7) for this specific route. Other routes already validate by hand.

### DIFF
--- a/backend/src/routes/projectChat.ts
+++ b/backend/src/routes/projectChat.ts
@@ -24,19 +24,129 @@ When the user wants to use an existing project document as a starting point for 
 
 export const projectChatRouter = Router({ mergeParams: true });
 
+// Mirrors the hand-written parsers in routes/chat.ts so this route doesn't
+// rely on a bare `req.body as { ... }` cast — every field flowing into the
+// LLM, the DB, or the response stream is shape-checked first.
+
+function parseChatMessages(value: unknown):
+    | { ok: true; messages: ChatMessage[] }
+    | { ok: false; detail: string } {
+    if (!Array.isArray(value) || value.length === 0) {
+        return { ok: false, detail: "messages must be a non-empty array" };
+    }
+    for (const message of value) {
+        if (!message || typeof message !== "object" || Array.isArray(message)) {
+            return { ok: false, detail: "messages must contain objects" };
+        }
+        const row = message as Record<string, unknown>;
+        if (typeof row.role !== "string") {
+            return { ok: false, detail: "message.role must be a string" };
+        }
+        if (row.content !== null && typeof row.content !== "string") {
+            return {
+                ok: false,
+                detail: "message.content must be a string or null",
+            };
+        }
+    }
+    return { ok: true, messages: value as ChatMessage[] };
+}
+
+function parseOptionalChatId(value: unknown):
+    | { ok: true; chatId: string | null }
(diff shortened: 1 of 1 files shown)


willchen96:
Closing this stale implementation during PR backlog cleanup. Request validation is still worthwhile, but this change should be refreshed against the current project-chat route before landing.

willchen96:
Reopened after review: current main still relies on an unchecked request-body cast in the project chat route. The patch needs rebasing and updating for current request fields, but the validation work remains relevant.

willchen96:
Hi @Shawnaldinho, thanks for identifying and addressing the missing request validation here. Validating the body before database writes and LLM calls is a worthwhile improvement.

`main` has changed substantially since this PR was opened. The PR is now about 119 commits behind, has merge conflicts, and the project-chat endpoint has gained support for `ask_inputs_response`.

I’m going to close this PR for now to keep the queue current, but the underlying fix remains valuable. We would welcome a fresh PR based on the latest `main` that:

- Preserves and validates newer request fields, including `ask_inputs_response`.
- Adds automated tests showing that malformed bodies return `400` before database writes or LLM calls.
- Restricts `message.role` to the roles the endpoint actually supports.
- Returns trimmed document IDs and filenames after validating them.
- Considers sharing the validation logic from `routes/chat.ts` to prevent the two endpoints from drifting apart.

Please feel free to reference #155 in the new PR. Thanks again for the contribution!

PR #156, opened 2026-05-19
feat: structured JSON logger with per-request id and access log

Shawnaldinho:
## Summary
Backend logs were all free-form \`console.log\`/\`console.error\`, so an error you see on the client can't be correlated to the line that produced it on the server, and there's no way to filter by route/status/user_id from a log aggregator. This PR puts the floor in: a tiny structured logger, a per-request id, and an access log line per response.

## Changes
- **\`backend/src/lib/logger.ts\`** — zero-dep JSON-line logger. \`logger.{debug,info,warn,error}({ fields }, msg)\` writes one JSON object per call (stdout for info/debug, stderr for warn/error). Level is controlled by \`LOG_LEVEL\` (defaults: \`info\` in production, \`debug\` elsewhere). \`logger.child({ ... })\` returns a logger that prefixes every event with the given fields. \`errFields(err)\` pulls \`err_name\`/\`err_msg\`/\`err_stack\` out of an \`Error\`. Hand-written instead of pulling pino — only ~80 lines today; the call sites won't change if it's later swapped.
- **\`backend/src/middleware/requestContext.ts\`** — assigns a per-request id (inbound \`X-Request-Id\` if present, else a UUID), echoes it back on the response, parks a child logger on \`res.locals.log\`, and on response \`close\` emits one access-log line with method/route/status/elapsed_ms/user_id. Wired into \`index.ts\` before the rate limiters.
- **Demonstrate adoption** in the most user-visible places: the two streaming error paths (\`POST /chat\`, \`POST /projects/:projectId/chat\`) and the title-generation handler now use the request-scoped logger so their error lines correlate with the access log via \`request_id\`. Debug-style \`console.log\` calls in \`chatTools.ts\` / \`documents.ts\` were left alone — that's a separate, broader cleanup.

## Why
Concrete things this enables that weren't possible before:
- \`request_id=<id> level=error\` filter in CloudWatch / Loki / Datadog pulls every line tied to one failing request.
- Frontend can surface the \`X-Request-Id\` so a user-reported \"chat broke\" comes with the exact id you grep for.
- Distinguishing client-side cancels (\`status=499\`) from server-side errors (\`status=500\`) without touching the console.

Addresses the observability concern from https://insights.flank.ai/where-mikeoss-falls-short.html (gap 14). Doesn't try to also tackle tracing / metrics / Sentry — those are reasonable follow-ups built on this foundation.

## Testing
- Hand-traced a request through \`requestContext\` → rate-limiter → handler → \`close\` and confirmed (a) inbound \`X-Request-Id\` is preserved, (b) generated UUID echoed back when absent, (c) access-log line emits with the same id, (d) the streaming error logs share that id.

### DIFF
--- a/backend/src/index.ts
+++ b/backend/src/index.ts
@@ -11,6 +11,8 @@ import { tabularRouter } from "./routes/tabular";
 import { workflowsRouter } from "./routes/workflows";
 import { userRouter } from "./routes/user";
 import { downloadsRouter } from "./routes/downloads";
+import { requestContext } from "./middleware/requestContext";
+import { logger } from "./lib/logger";
 
 const app = express();
 const PORT = process.env.PORT ?? 3001;
@@ -95,6 +97,8 @@ app.use(
   }),
 );
 
+app.use(requestContext());
+
 app.use(generalLimiter);
 
 app.use(express.json({ limit: "50mb" }));
@@ -122,5 +126,5 @@ app.use("/download", downloadsRouter);
 app.get("/health", (_req, res) => res.json({ ok: true }));
 
 app.listen(PORT, () => {
-  console.log(`Mike backend running on port ${PORT}`);
+  logger.info({ port: PORT }, "backend listening");
 });
--- a/backend/src/lib/logger.ts
+++ b/backend/src/lib/logger.ts
@@ -1,0 +1,86 @@
+/**
+ * Tiny zero-dependency structured logger. Every call emits one JSON
+ * line to stdout/stderr, which is the shape any log aggregator (CW
+ * Logs, Loki, Datadog) is happy to ingest. Fields are merged into the
+ * top-level object so a query like `{level=error route=/chat err_msg=*}`
+ * works without parsing message strings.
+ *
+ * We didn't pull in pino because we only need ~30 lines, and adding a
+ * dep + a bunch of transport configuration is overkill for the current
+ * single-process backend. If the project grows into multi-process or
+ * async transports, replace the body of this file with pino — the
+ * call sites won't have to change.
+ */
+
+type LogLevel = "debug" | "info" | "warn" | "error";
+
+type Fields = Record<string, unknown>;
+
+const LEVEL_RANK: Record<LogLevel, number> = {
+    debug: 10,
+    info: 20,
+    warn: 30,
+    error: 40,
+};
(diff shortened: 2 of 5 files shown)


amal66:
Thanks for taking this on — structured JSON logs with per-request ids are something this backend has genuinely needed, and the overall shape (child loggers, access line on close) is right. A few must-fixes before it can land:

- Error fields need to go through main's `safeErrorLog`: upstream SDK errors embed keys verbatim ("Incorrect API key provided: sk-ant-…"), and `errFields` emitting raw `err.message`/`err.stack` regresses the redaction main already does at these call sites.
- `emit()` calls `JSON.stringify` unguarded — a circular or BigInt field throws inside an async handler's catch block, which is a process-fatal unhandled rejection. Logging must never crash the request it's reporting on.
- The PR description promises 499 for client aborts, but aborts log as clean 200s. `res.writableFinished === false` at close-time is the signal.
- The access log leaks secrets: the unmatched-route fallback logs `req.originalUrl`, so a 404 on `/x?token=SECRET` writes the token in plaintext (reproduced live). Matched routes also lose their mount (`req.route.path` is router-relative) — use `req.baseUrl + req.route.path`, and `req.path` as the fallback.
- Inbound `X-Request-Id` is length-bounded but not charset-bounded; restrict it (e.g. `[A-Za-z0-9._-]{1,128}`) and regenerate otherwise, since it's echoed into logs.

Could you also add tests and rebase onto current main — the branch predates our vitest setup, so tests can only run once rebased. One admin note: the commit author email isn't linked to your GitHub account, so the CLA check can't pass until that's sorted.

To make this concrete, https://github.com/Shawnaldinho/mike/pull/1 against your branch implements all of the above with tests — merge or cherry-pick as you like. It sits directly on your branch as two focused commits; the redaction helper is vendored byte-identical to main's, so your later rebase merges clean. `tsc` and the build pass on it, and the included tests run once you're on main's vitest setup (I verified them green there).

amal66 (changes requested):
Requesting changes for the redaction and crash-safety items detailed in my comment below; there's a follow-up PR against your branch implementing them. Happy to dismiss once addressed.

willchen96:
Thanks for this Shawn, this has been superseded by newer PR #283

PR #158, opened 2026-05-20
security: spotlight untrusted content + ship an honest threat model

Shawnaldinho:
## Background

Replaces the closed [#154](https://github.com/willchen96/mike/pull/154). That PR was cosmetic — homoglyph substitution + a static `<available_documents>` tag — and I had not actually tested it. The reviewer was right to call it out. This PR is the honest version.

**Scope statement up front:** this raises the bar on casual prompt injection. It does not prevent a determined attacker. Real prevention needs output classification + capability containment + behavioural validation against the model, and those are several-week follow-ups documented as gaps in [`docs/SECURITY-MODEL.md`](https://github.com/willchen96/mike/blob/main/docs/SECURITY-MODEL.md). The LLM is not a security boundary; the docs and README now say so explicitly.

### 1. Per-request spotlighting fence — `backend/src/lib/promptFence.ts`
Each request generates a fresh 64-bit hex nonce. Untrusted spans are wrapped as

\`\`\`
«UNTRUSTED:<nonce>:<kind>»...payload...«END:<nonce>»
\`\`\`

The system prompt carries a `fenceInstructions(nonce)` block exactly once per turn telling the model that fenced content is data, not instructions, and that the nonce rotates per request and cannot be forged. **Security comes from the unguessable nonce, not from sanitising the payload** — that was the PR #154 mistake. Light hygiene strips C0 control bytes and caps label-shaped fields at 512 chars; no XML angle-bracket substitution.

### 2. Apply it to the actual attack surface — `backend/src/lib/chatTools.ts`
The real prompt-injection vector isn't filenames in the system prompt (the only thing #154 touched) — it's **document body text returned by tool calls**. This PR fences:

- `buildMessages` — filenames, folder paths, workflow titles in the system prompt and on user messages.
- `enrichWithPriorEvents` — filenames and workflow titles in the prior-turn summary attached to the last assistant message.
- `runToolCalls` for `read_document`, `find_in_document`, `fetch_documents`, `list_documents`, `list_workflows`, `read_workflow` — document body text, search excerpts, the JSON payloads of list tools, and workflow `prompt_md`.

### 3. Thread the nonce through the routes — `backend/src/routes/{chat,projectChat}.ts`
Same nonce is generated at the top of the POST handler and passed into `buildMessages`, `enrichWithPriorEvents`, and `runLLMStream` so the system-prompt convention matches the tool-result markers.

### 4. Adversarial corpus + structural test runner — `backend/tests/promptFence/`
20 attacks across the real surface: naive override in filenames, newline-break, XML-fence-break, oversize label, homoglyph, role-play in doc body, fence-close forgery with a guessed nonce, base64 payload, tool-call injection, exfiltration prompts, multi-turn drift, workflow title/prompt overrides, search-hit injection, `list_documents` filename injection, folder path injection, instruction-shaped-but-benign, plus a clean control case.

`runStructural.ts` walks every entry through the **real** `fenceLabel`/`fenceBody`/`buildMessages` code paths and asserts:
- per-request nonce shape (16 lowercase hex);
- two consecutive nonces differ (replay protection);
- the legitimate close marker appears exactly once even when the payload tries to forge it with a different static nonce;
- control bytes are stripped;
- oversize labels truncate to 512 chars + ellipsis;
- the system prompt carries the matching `fenceInstructions` block;
- the evil filename's embedded \"SYSTEM:\" string lands inside an unclosed fence, not bare in the system prompt.

\`\`\`
$ npm run test:prompt-fence --prefix backend
OK 91 structural assertions passed across 20 corpus entries.
\`\`\`

I actually ran this. Output included verbatim above.

### 5. Threat model — `docs/SECURITY-MODEL.md`
- Threat actors and surfaces table.
- What the codebase does today (this PR's mechanism).
- **What the codebase does NOT do** — explicitly: no behavioural validation against live models, no output classification, no capability containment between read and write tools in a single turn, no context-window-crowding defence, multi-turn carry-over of a compromised prior turn is unmitigated.
- What operators should do (don't upload from untrusted sources without reviewing tool calls; report via GitHub security advisories).

### 6. README — security section
One paragraph + a link to the threat model + the GitHub private-advisories link.

## What this PR explicitly does NOT claim

- It does **not** prevent prompt injection. Plausibly-shaped requests inside a fence (no fence-break attempt) will still reach the model and may still be obeyed.
- It does **not** test behavioural compliance against the model. The structural tests prove the *wrapping* is correct; whether a given model honours the spotlighting convention requires running the corpus against a live API and judging responses. Documented as a follow-up.
- It does **not** add capability containment. A turn can still chain `read_document` → `edit_document` without user confirmation.

## Testing

\`\`\`
npm run build --prefix backend       # passes
npm run test:prompt-fence --prefix backend  # 91/91 assertions pass
\`\`\`

Closes the prompt-injection concern from https://insights.flank.ai/where-mikeoss-falls-short.html (gap 12) in honest scope: structural defence-in-depth + threat-model writeup, not a claim of prevention.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,12 @@ Open `http://localhost:3000`.
 2. If you did not set provider keys in `backend/.env`, open **Account > Models & API Keys** and add an Anthropic, Gemini, or OpenAI API key.
 3. Create or open a project and start chatting with documents.
 
+## Security model and trust boundaries
+
+Mike does not treat the LLM as a security boundary. Document contents, filenames, and folder paths supplied by users (or by anyone who hands a document to a user) can attempt to instruct the model. The codebase wraps untrusted spans in a per-request spotlighting fence so the model can distinguish data from instructions; that raises the bar on casual prompt injection but does not prevent a determined attacker from getting the model to comply. **Do not upload documents from untrusted sources without reviewing the model's tool calls before accepting its output.** See [`docs/SECURITY-MODEL.md`](docs/SECURITY-MODEL.md) for the full threat model, what is and is not defended, and how to run the adversarial test corpus locally with `npm run test:prompt-fence --prefix backend`.
+
+To report a vulnerability privately, use [GitHub's security advisories](https://github.com/willchen96/mike/security/advisories/new).
+
 ## Troubleshooting
 
 **Sign-up confirmation email never arrives.** Confirmation emails are sent by Supabase Auth, not by Mike. For local development, the simplest fix is to disable email confirmation in **Supabase > Authentication > Providers > Email**. For production, configure custom SMTP in Supabase; the built-in mailer is heavily rate-limited and may be restricted on newer projects.
--- a/backend/package.json
+++ b/backend/package.json
@@ -5,7 +5,8 @@
   "scripts": {
     "dev": "tsx watch src/index.ts",
     "build": "tsc",
-    "start": "node dist/index.js"
+    "start": "node dist/index.js",
+    "test:prompt-fence": "tsx tests/promptFence/runStructural.ts"
   },
   "dependencies": {
     "@anthropic-ai/sdk": "^0.90.0",
--- a/backend/src/lib/chatTools.ts
+++ b/backend/src/lib/chatTools.ts
@@ -24,6 +24,13 @@ import {
     type LlmMessage,
     type OpenAIToolSchema,
 } from "./llm";
+import {
+    fenceBody,
+    fenceInstructions,
+    fenceLabel,
+    makeFenceNonce,
+    type FenceNonce,
+} from "./promptFence";
 
 const STANDARD_FONT_DATA_URL = (() => {
     try {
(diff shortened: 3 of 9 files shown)


b1rdmania (commented):
Just a contributor reading through the open PRs, not speaking for the maintainers — but for what it's worth, this is the strongest security PR I've seen in the current queue, and I appreciate how honestly it's scoped: it raises the bar on casual injection without pretending the LLM is a security boundary.

The design reads as the right one to me — security comes from the per-request unguessable nonce (`«UNTRUSTED:nonce:kind»…«END:nonce»`), not from mangling the payload, and the corpus test that tries to forge an `«END»` marker is exactly the case that matters. Threading a single nonce per request through both chat routes is clean.

From the outside the only thing in the way of a review looks mechanical: it currently conflicts with `main` and the CLA check is red — a rebase + CLA signature would make it land-able. Would be a shame to let this one rot.

willchen96:
Thanks for this contribution, apologies that I was not able to get to merge this in early on. The same ideas have been implemented in a new PR #258. Closing this due to conflicts with main branch.

PR #168, opened 2026-06-03
feat(i18n): add next-intl internationalization scaffold

tksuns12:
## Summary

Adds internationalization (i18n) to the frontend using **next-intl** (v4, App Router, **cookie-based — no URL changes**). All user-facing UI strings are extracted into a typed message catalog. English-only for now; adding a language is a matter of dropping in `messages/<locale>.json`.

### Infrastructure
- `next-intl@4.13` wired via `createNextIntlPlugin` in `next.config.ts`
- `src/i18n/config.ts` — single source of truth: locales, default locale, cookie name, labels
- `src/i18n/request.ts` — resolves locale from the `NEXT_LOCALE` cookie (fallback `en`)
- `src/i18n/actions.ts` — `setLocale` server action (writes the cookie)
- `NextIntlClientProvider` in the root layout + dynamic `<html lang>`
- `src/types/next-intl.d.ts` — **typed messages**: every `t('key')` is compile-checked against `en.json`
- `LanguageSwitcher` component (design-system dropdown), mounted on the Account page
- Localized `generateMetadata`, `error`, `not-found`, and `global-error` (the last self-provides its own `NextIntlClientProvider` because it replaces the root layout when active)

### String migration
- ~85 components/pages migrated from hardcoded English to `useTranslations()` / `getTranslations()`
- `messages/en.json`: **671 keys across 16 namespaces** (`Common`, `Shared`, `Documents`, `Projects`, `Assistant`, `Tabular`, `Workflows`, `Modals`, `Auth`, `Account`, …)

## Adding a new language
1. Add the code to `locales` + a label to `localeLabels` in `src/i18n/config.ts`
2. Add `messages/<locale>.json` (same shape as `en.json`)

The switcher and cookie handling pick it up automatically — no other wiring needed.

## Deliberately left as-is
- Brand name ("Mike"), model provider names (Anthropic/Google/OpenAI), and version codes (`V{n}`)
- `console.*`, API field names, route paths, and LLM prompt strings (not UI copy)
- Date formatting already uses locale-aware `toLocaleDateString`

## Test plan
- [x] `tsc --noEmit` — 0 errors (typed catalog validates all 671 keys against `t()` calls)
- [x] `next build` — all 20 routes compile
- [x] No new ESLint errors vs `main`
- [ ] Manual: switch language on `/account`, confirm `NEXT_LOCALE` cookie is set and server components re-render
- [ ] Manual: spot-check key screens (assistant, projects, tabular reviews, workflows) render expected copy

### DIFF
--- a/frontend/bun.lock
+++ b/frontend/bun.lock
@@ -30,6 +30,7 @@
         "mammoth": "^1.11.0",
         "marked": "^17.0.1",
         "next": "^16.2.6",
+        "next-intl": "^4.13.0",
         "nextjs-toploader": "^3.9.17",
         "pdfjs-dist": "4.10.38",
         "react": "19.2.0",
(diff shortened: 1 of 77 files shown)


willchen96:
Thanks for the contribution. I’m going to close this PR because the repo has moved too far along since it was opened, and the branch now conflicts with the current codebase.

This direction may still be worth revisiting later, but it would need to come back as a fresh PR against current `main`.

PR #169, opened 2026-06-05
feat: add integration tools abstraction and TrustFoundry legal research

iamgroot2285:
## Summary
- Add `ToolIntegration` interface for plugging external tool providers into assistant chat.
  The current design targets integrations backed by API keys; providers requiring OAuth or
  per-user token management would need additional key-storage infrastructure.
- Surface integration tool display names in SSE `tool_call_start` events so the frontend
  shows user-friendly labels (e.g. "Running TrustFoundry Legal Research...") — display
  names are defined per-provider, keeping everything encapsulated in a single module
- Return explicit error results for unrecognized tool calls instead of silently dropping them
- Add `docs/integrations.md` documenting optional third-party integration setup
- Implement TrustFoundry as the first integration: legal search, citation validation,
  and usage tools

## Why
Mike's chat tools were hardcoded — no way to add external tool providers without editing
`chatTools.ts` directly. The `ToolIntegration` abstraction lets new providers be added as
self-contained modules with their own tool definitions, display names, and execution logic.

## Test
- `npm run build --prefix frontend`
- Verified TrustFoundry tools appear in chat when `TRUSTFOUNDRY_ENABLED=true` and
  `TRUSTFOUNDRY_API_KEY` are set; hidden when disabled


willchen96:
Thanks but we are not accepting PRs from vendors integrating their paid services as part of our policy.

PR #170, opened 2026-06-05
fix: add SSE heartbeats to prevent stream timeout on long tool calls

iamgroot2285:
## Summary
- Extract SSE setup into `backend/src/lib/sse.ts` with 15-second keepalive heartbeats
- Disable per-request inactivity timeouts (`req.setTimeout(0)`, `res.setTimeout(0)`,
  `socket.setTimeout(0)`) and Nagle's algorithm (`socket.setNoDelay(true)`) for SSE
  connections — scoped to authenticated long-running assistant streams only
- Apply to all SSE endpoints: chat, project chat, tabular chat, and tabular generate

## Why
Long-running tool calls (e.g. external API searches, 30-60s) left the SSE connection
idle from the browser's perspective, causing "Error in input stream" when the connection
was killed. Periodic `: keepalive` comments keep the connection alive, and disabling
per-request inactivity timeouts prevents Node.js from reaping the socket during legitimate
long-running work.

## Test
- Ran multiple long-running tool calls; streams completed without error

### DIFF
--- a/backend/src/lib/sse.ts
+++ b/backend/src/lib/sse.ts
@@ -1,0 +1,53 @@
+import type { Request, Response } from "express";
+
+export function startSseStream(
+    req: Request,
+    res: Response,
+    options: { heartbeatMs?: number } = {},
+) {
+    const heartbeatMs = options.heartbeatMs ?? 15000;
+    let closed = false;
+
+    // Disable all socket-level timeouts so long-running tool calls (e.g.
+    // TrustFoundry agentic search) don't get killed mid-stream.
+    req.setTimeout(0);
+    res.setTimeout(0);
+    if (req.socket) {
+        req.socket.setTimeout(0);
+        req.socket.setNoDelay(true);
+    }
+
+    res.setHeader("Content-Type", "text/event-stream");
+    res.setHeader("Cache-Control", "no-cache");
+    res.setHeader("Connection", "keep-alive");
+    res.setHeader("X-Accel-Buffering", "no");
+    res.flushHeaders();
+
+    const write = (line: string) => {
+        if (closed || res.writableEnded) return;
+        res.write(line);
+    };
+
+    const heartbeat = setInterval(() => {
+        write(": keepalive\n\n");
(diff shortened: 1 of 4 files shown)


amal66:
Thanks for this — it fixes a real failure mode the right way: four hand-rolled SSE setups (one of which never even wired close-to-abort) become a single audited helper with a heartbeat and guarded writes. I tested the lifecycle paths fairly hard — client abort, clean completion, mid-stream error, socket-timeout scoping, frontend parsers ignoring `: keepalive` comments — and found no bugs. Nice work.

Two things before merge: rebase onto `main` (one trivial import conflict in `tabular.ts` — keep both import groups — and it puts you on the new vitest infra), and the CLA: your commit email isn't linked to a GitHub account, so the bot can't record a signature, which blocks acceptance regardless of code quality. Linking the address should be all it takes.

Hardening worth taking: guard writes with `res.destroyed` too (an async write to a destroyed socket rejects, and with no `unhandledRejection` handler that crashes the process); abort the signal only when the client actually went away, not on clean completion; unit tests for the helper's timer/abort/write lifecycle; and thread `stream.signal` into tabular generate's LLM calls so a closed tab stops the whole `Promise.all`.

One question: 15s clears nginx-style idle windows — does it also clear whatever fronts production (Cloudflare/ALB)?

I've opened https://github.com/trustfoundry-ai/mike/pull/1 against your branch with all of the above plus tests — merge or cherry-pick as you like. It sits directly on your branch as two focused commits, with the rebase left to you. `tsc` and the build are clean on it; the unit tests ship with it and run once you're on main's vitest setup (I verified them green on a main-merged copy).

amal66 (commented):
This is in good shape — details in my comment below. Mainly needs a rebase and the CLA sorted, plus some optional hardening I've PR'd against your branch.

willchen96:
@iamgroot2285 bumping

willchen96:
Thanks @iamgroot2285 closing this for now due to inactivity - feel free to submit a new PR.

PR #171, opened 2026-06-06
Add courtlistener intergration, liquid glass redesign, UI improvements, version control, various fixes

willchen96:
## Summary

This PR adds CourtListener case-law support, improves assistant streaming/tool events, expands document version management, and refreshes the app UI with liquid-glass styling, shared modals, standardized page headers, and improved project document workflows.

image

## CourtListener and Case-Law Tools

- Adds backend CourtListener integration and `/case-law` routing.
- Courtlistener integration available via both API and self-hosted replicated bulk-data.
- Adds legal-source tool support for citation verification, case fetching, keyword searching, and reading selected opinions.
- Updates assistant prompts and tool handling so case-law workflows follow the intended flow:
  verify citations → fetch cases → search keywords → read only needed opinions if required.
- Adds frontend support for case-law activity events, citation display, relevant quote panels, and case-law side panel UI.

Screenshot 2026-06-06 at 3 55 50 PM

## Assistant Streaming and Responses

- Improves assistant SSE handling for tool events, reasoning, citations, document edits, and generated/downloaded documents.
- Adds better cancellation/abort handling for LLM streams.
- Improves response rendering for tables, citation blocks, annotation text, reasoning collapse/expand behavior, and pre-response activity containers.
- Adds support for raw stream logging and expanded model/provider settings.

## Document Version Management

- Refactors document metadata toward version-based handling.
- Updates backend document/project routes to use active/latest document versions for filenames, storage paths, and metadata.
- Adds document version UI improvements, including:
  - right-side document side panel
  - version list and selected-version details
  - upload new version flows
  - delete document/version confirmation states
  - drag-and-drop save-as-new-version behavior
- Removes legacy filename/storage fallback behavior and unused structure-tree handling.

image

## Liquid Glass UI

- Adds and standardizes liquid-glass styling across shared app surfaces.
- Updates shared modal, confirm popup, warning popup, page header, and relevant quotes components.
- Standardizes modal structure, buttons, padding, translucency, and close button styling.
- Refactors page header actions into array-driven action definitions with shared styling.
- Removes the old `HeaderSearchBtn` component.
Screenshot 2026-06-06 at 4 00 16 PM
Screenshot 2026-06-06 at 4 00 28 PM
Screenshot 2026-06-06 at 4 01 22 PM

## Project, Workflow, Tabular Review, and Settings UI

- Updates project tables, skeletons, folder rows, version rows, hover states, and document row spacing.
- Improves tabular review modals, side panel tabs, chat panel styling, and table/citation presentation.
- Updates workflow modals and workflow list styling.
- Refreshes account/settings pages with boxed sections, flatter inputs, API key styling, and model preference updates.
- Applies global background and layout refinements across the app.

image

## Security

- Sanitize CourtListener-rendered opinion HTML with `DOMPurify`
- Add safe external-link handling with `rel="noopener noreferrer"`
- Store per-user API keys encrypted with AES-GCM
- Document `USER_API_KEYS_ENCRYPTION_SECRET` for API key encryption
- Revoke direct anon/authenticated table access for backend-owned tables, including API key and CourtListener bulk-data tables


backend/src/routes/caseLaw.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    try {
+        const userId = String(res.locals.userId ?? "");
+        const settings = await getUserModelSettings(userId);
+        devLog("[case-law/case-opinions] loading sidepanel opinions", {
+            clusterId,
+        });
+        const db = createServerSupabase();
+        const fetchKey = String(clusterId);
```
amal66: **Bug: one user can accidentally use another user's CourtListener credentials.**

The dedup cache key is just the case ID. If User A and User B both open the same case at the same time, User B's request joins User A's in-flight fetch — and the fetch runs with User A's API token. This means User A's CourtListener credits get spent for User B's request.

Fix: include the user ID in the cache key so each user's request is independent.
```ts
// change this:
const fetchKey = String(clusterId);
// to this:
const fetchKey = `${userId}:${clusterId}`;
```

backend/src/routes/tabular.ts
```diff
@@ -471,8 +482,19 @@ tabularRouter.patch("/:reviewId", requireAuth, async (req, res) => {
     if (req.body.title != null) updates.title = req.body.title;
     if (req.body.columns_config != null)
         updates.columns_config = req.body.columns_config;
```
amal66: **Bug: any collaborator can change the column definitions of a shared review.**

`shared_with` and `project_id` are correctly protected — only the review owner can change them (checked a bit further down in this route). But `columns_config` (which defines the entire column structure and AI prompts) has no owner check, so any collaborator with read access can silently restructure the review.

Fix: move this block to after the `ensureReviewAccess` call, then add an owner check.
```ts
if (req.body.columns_config != null) {
    if (!access.isOwner)
        return void res.status(403).json({ detail: "Only the review owner can change columns" });
    updates.columns_config = req.body.columns_config;
}
```

backend/src/routes/documents.ts
```diff
@@ -372,10 +375,204 @@ documentsRouter.get("/:documentId/versions", requireAuth, async (req, res) => {
...
+
+    const { data: sourceDoc } = await db
+      .from("documents")
+      .select("id, user_id, project_id")
+      .eq("id", sourceDocumentId)
+      .single();
+    if (!sourceDoc)
+      return void res.status(404).json({ detail: "Source document not found" });
+    const sourceAccess = await ensureDocAccess(sourceDoc, userId, userEmail, db);
+    if (!sourceAccess.ok)
```
amal66, line 424: **Bug: a collaborator with read access can permanently delete a source document they don't own.**

This line checks that the user can *access* the source document, but not that they *own* it. Further down in this route, the source document is permanently deleted when the source and target are in the same project. A read-only collaborator should be able to copy from a shared document, but they should not be able to trigger its deletion.

Fix: add an ownership check right after this.
```ts
if (!sourceAccess.isOwner) {
    return void res.status(403).json({
        detail: "Only the source document owner can move it into a version."
    });
}
```

backend/src/lib/userApiKeys.ts
```diff
@@ -29,6 +40,12 @@ function envApiKey(provider: ApiKeyProvider): string | null {
     if (provider === "openai") {
         return process.env.OPENAI_API_KEY?.trim() || null;
     }
+    if (provider === "openrouter") {
+        return process.env.OPENROUTER_API_KEY?.trim() || null;
+    }
+    if (provider === "courtlistener") {
+        return process.env.COURTLISTENER_API_TOKEN?.trim() || null;
+    }
     return process.env.GEMINI_API_KEY?.trim() || null;
```
amal66: **Two issues in this file:**

**1. Any unknown provider silently returns the Gemini API key (this line).**

This fallback `return` means any provider that doesn't match a previous `if` block gets Gemini's key instead of `null`. If a new provider is added later but its `if` block is accidentally omitted, it will silently send Gemini credentials to the wrong service.

Fix: `return null` here so unknown providers fail closed.

**2. The encryption key derivation uses SHA-256 instead of a proper KDF (around line 61).**

In `encryptionKey()`, `createHash("sha256")` is used to derive the AES key from the environment secret. SHA-256 is extremely fast, making it easy to brute-force if the secret is ever short or predictable. A key derivation function (KDF) like `scryptSync` is specifically designed to be slow for this purpose.

Fix in `encryptionKey()`:
```ts
// change this:
return crypto.createHash("sha256").update(secret).digest();
// to this:
return crypto.scryptSync(secret, "mike-user-api-keys-v1", 32);
```
Also update `.env.example` to note the secret should be at least 32 random characters, and that changing it invalidates all stored API keys (users will need to re-enter them).

backend/src/lib/llm/rawStreamLog.ts
```diff
@@ -0,0 +1,19 @@
+export function logRawLlmStream(args: {
+    provider: string;
+    model: string;
+    iteration: number;
+    label: string;
+    payload: unknown;
+}) {
+    if (
+        process.env.NODE_ENV === "production" &&
```
amal66: **Bug: raw LLM output is logged in every environment except production.**

The condition uses `&&` (AND): logging is skipped only when *both* `NODE_ENV === "production"` AND `LOG_RAW_LLM_STREAM` is not set. In practice, this means on staging, CI, or any developer's local machine, every full LLM response is printed to the logs — including sensitive content from user queries.

Fix: make logging opt-in everywhere.
```ts
// change this:
if (
    process.env.NODE_ENV === "production" &&
    process.env.LOG_RAW_LLM_STREAM !== "true"
) { return; }
// to this:
if (process.env.LOG_RAW_LLM_STREAM !== "true") return;
```

amal66 (changes requested):
Thanks for the huge amount of work here — CourtListener, version management, and the UI refresh are all well-scoped. Before this can merge, there are three categories of issues to work through.

---

## 1. Missing: testing evidence

The PR description has no testing section. For a change this size (106 files, 16K lines added), reviewers need to know:
- What did you actually run and verify by hand? (e.g. "I uploaded a new version, confirmed the version list updated, confirmed the old version is still accessible")
- What regression testing was done on existing features? (e.g. did you verify tabular reviews, document upload/download, and assistant chat still work after the refactor?)
- Before/after screenshots or screen recordings for the UI changes — the screenshots in the description show the new state, but not the old state to compare against.

Please add a **Testing** section to the PR description that covers at minimum:
1. The new CourtListener flow (end to end)
2. Document versioning (upload, switch version, delete version, delete document)
3. Tabular review creation, sharing, and column editing
4. Assistant streaming (start a chat, cancel mid-stream, start a new one)
5. Existing features you verified still work

---

## 2. Security and authorization issues (fix before merge)

See inline comments for `caseLaw.ts`, `tabular.ts`, `documents.ts`, `userApiKeys.ts`, and `rawStreamLog.ts`.

Additionally, in `tabular.ts` around **line 1377** in the `/:reviewId/chat` route:

> **Bug: a chat session from a different review can be reused here.**
>
> The check uses `||` (OR):
> ```ts
> const canUse =
>     !!existing &&
>     (existing.review_id === reviewId || existing.user_id === userId);
> ```
> This means a chat is valid if *either* it belongs to this review *or* it belongs to this user — not necessarily both. User A could pass a `chatId` from a completely different review they own and it would pass.
>
> Fix: use `&&` (AND):
> ```ts
> const canUse =
>     !!existing &&
>     existing.review_id === reviewId &&
>     existing.user_id === userId;
> ```

---

## 3. Data safety issues in `ProjectPage.tsx` (fix before merge)

**a) Single-version documents deleted with no confirmation (around line 713–719)**

`requestRemoveDoc` only shows a confirmation dialog when a document has more than one version. For single-version documents, the delete happens immediately with no "are you sure?" step. Deletion is permanent.

Fix: always show a confirmation, regardless of version count.

**b) Folder delete has no confirmation and cascades silently (around line 499–517)**

`handleDeleteFolder` immediately removes the folder and fires the delete request. If the folder contains documents or nested subfolders, everything is deleted without warning.

Fix: before calling this, show a confirmation dialog that tells the user how many subfolders and documents will be deleted (e.g. "This will permanently delete 2 subfolders and 5 documents. Continue?").

**c) Bulk document delete hides server errors (around line 857)**

`.catch(() => {})` swallows deletion errors silently. If some documents fail to delete on the server, the UI still removes them from the list — so the user thinks they're gone, but they aren't.

Fix:
```ts
const results = await Promise.allSettled(owned.map((id) => deleteDocument(id)));
const deletedIds = owned.filter((_, i) => results[i].status === "fulfilled");
const failedCount = owned.length - deletedIds.length;
// remove only deletedIds from state, show a warning if failedCount > 0
```

backend/schema.sql
```diff
@@ -50,7 +51,7 @@ create trigger on_auth_user_created
 create table if not exists public.user_api_keys (
```
amal66, line 51: **Security: `user_api_keys` has no Row-Level Security, only a REVOKE.**

This table stores every user's encrypted API keys (CourtListener, Claude, OpenAI, etc.). The only tenant boundary is `REVOKE ALL ON ... FROM anon, authenticated` (line ~402). If any endpoint accidentally queries this table using the authenticated role instead of the service role — even briefly, due to a misconfiguration — every user's encrypted keys become readable by every other user.

Row-Level Security gives a second, independent layer of protection at the database itself.

Fix:
```sql
alter table public.user_api_keys enable row level security;
create policy user_api_keys_own on public.user_api_keys
  for all using (user_id::text = auth.uid()::text);
```

backend/schema.sql
```diff
@@ -341,6 +340,41 @@ create table if not exists public.tabular_review_chat_messages (
 create index if not exists tabular_review_chat_messages_chat_idx
   on public.tabular_review_chat_messages(chat_id, created_at);
 
+-- ---------------------------------------------------------------------------
+-- CourtListener bulk-data indexes
+-- ---------------------------------------------------------------------------
+
+create table if not exists public.courtlistener_citation_index (
```
amal66, line 364: **Security: CourtListener index tables also have no Row-Level Security.**

`courtlistener_citation_index` and `courtlistener_opinion_cluster_index` use the same `REVOKE ALL` pattern as the rest of the schema, with no RLS policies. These tables contain bulk case-law data imported from CourtListener — while less sensitive than API keys, they should still be protected with RLS so that the REVOKE is not the only guard.

Fix:
```sql
alter table public.courtlistener_citation_index enable row level security;
alter table public.courtlistener_opinion_cluster_index enable row level security;
```
Since these are shared read-only reference tables (not per-user data), a simple allow-all-reads policy is sufficient:
```sql
create policy cl_citation_read on public.courtlistener_citation_index for select using (true);
create policy cl_cluster_read on public.courtlistener_opinion_cluster_index for select using (true);
```

backend/src/lib/courtlistener.ts
```diff
@@ -0,0 +1,1008 @@
...
+    apiToken?: string | null,
+): Promise<T> {
+    const url = pathOrUrl.startsWith("http")
+        ? pathOrUrl
+        : `${COURTLISTENER_BASE}${pathOrUrl}`;
+    devLog("[courtlistener/api] request", {
+        method: init?.method ?? "GET",
+        path: pathOrUrl,
+        url,
+    });
```
amal66, line 90: **Reliability: `courtlistenerFetch` has no timeout.**

This `fetch()` call has no `AbortSignal.timeout()` and no abort signal from any caller. If CourtListener is slow or unresponsive, this request hangs indefinitely — blocking the server request, consuming a connection, and eventually timing out only at the infrastructure level (if at all).

Fix: add a default timeout.
```ts
const response = await fetch(url, {
    ...init,
    signal: init?.signal ?? AbortSignal.timeout(15_000),
    headers: {
        ...courtlistenerHeaders(apiToken),
        ...(init?.headers ?? {}),
    },
});
```

backend/src/lib/courtlistener.ts
```diff
@@ -0,0 +1,1008 @@
...
+    clusterId: number;
+    maxChars: number;
+    includeFullText?: boolean;
+    apiToken?: string | null;
+}) {
+    const opinions: ReturnType<typeof compactOpinion>[] = [];
+    const rawOpinions: JsonRecord[] = [];
+    let nextUrl: string | null = `/opinions/?cluster=${args.clusterId}`;
+
+    while (nextUrl) {
```
amal66: **Reliability: pagination loop has no page cap.**

This `while (nextUrl)` loop terminates only when CourtListener stops returning a `next` URL. A cluster with hundreds of opinions will fetch all of them, making an unbounded number of API calls and accumulating all results in memory. This can exhaust serverless function timeouts and memory limits.

Fix: add a hard page limit and track a running character budget.
```ts
const MAX_PAGES = 10;
let pages = 0;
let remainingChars = args.maxChars ?? 50_000;

while (nextUrl && pages < MAX_PAGES && remainingChars > 0) {
    pages++;
    // ... fetch page, add opinions, decrement remainingChars
}
```

backend/src/routes/documents.ts
```diff
@@ -482,10 +693,10 @@ documentsRouter.post(
     const nextVersionNumber =
       ((maxRow?.version_number as number | null) ?? 1) + 1;
```
amal66, line 699: **Data integrity: race condition can produce duplicate version numbers.**

This follows the classic SELECT max → compute → INSERT pattern with no transaction or row-level lock. Two concurrent uploads to the same document will both read the same `max(version_number)`, both compute the same `nextVersionNumber`, and both insert it successfully — because there is no UNIQUE constraint on `(document_id, version_number)` in the schema.

The result is two version rows with the same version number, which breaks the UI display and any logic that assumes version numbers are unique per document.

Fix: either add a UNIQUE constraint so the database rejects duplicates:
```sql
alter table public.document_versions
  add constraint document_versions_doc_version_unique unique (document_id, version_number);
```
Or use a Postgres sequence / a database function that assigns the number atomically inside a single transaction.

backend/src/routes/documents.ts
```diff
@@ -507,39 +721,19 @@ documentsRouter.post(
...
-    if (providedDisplayName) {
-      const hasExt = /\.[a-z0-9]{1,6}$/i.test(providedDisplayName);
-      const existingExt = (doc.filename as string | null)?.match(
-        /\.[a-z0-9]{1,6}$/i,
-      )?.[0];
-      const uploadedExt = suffix ? `.${suffix}` : "";
-      const ext = hasExt ? "" : uploadedExt || existingExt || "";
-      documentsUpdate.filename = `${providedDisplayName}${ext}`;
-    }
     await db
```
amal66: **Bug: the result of this update is never checked.**

If this `.update({ current_version_id })` call fails silently — network blip, RLS issue, anything — the new version row has already been inserted and committed to the database, but the document still points to the old version. The caller receives a 201 response and believes the upload succeeded, but the document's active version never changed.

For comparison, the `from-document` route (around line 538) checks the same update correctly:
```ts
const { error: updateDocErr } = await db.from("documents").update(...).eq("id", documentId);
if (updateDocErr) {
    return void res.status(500).json({ detail: "Failed to update document current version." });
}
```
Apply the same pattern here.

amal66 (commented):
Follow-up round of bugs found after the first review — all six were verified independently by two separate reviewers before being posted.

---

**In `useAssistantChat.ts` (cancel() / abort controller — lines ~164–171 and ~1187–1189)**

The `cancel()` function nullifies `abortControllerRef.current` immediately and synchronously (line 167), before the in-flight request finishes. The request's `finally` block (line 1188) also nullifies the ref unconditionally, with no identity check. The failure sequence is:

1. Request A is streaming.
2. User calls `cancel()` → ref is set to `null`.
3. New Request B starts → a new controller is created and assigned to `abortControllerRef.current`.
4. Request A's `finally` runs and sets `abortControllerRef.current = null`.
5. Request B's controller is now orphaned — cancelling is impossible for the duration of that stream.

Fix: in the `finally` block, only null the ref if it still equals the controller that block owns:
```ts
} finally {
    if (abortControllerRef.current === controller) {
        abortControllerRef.current = null;
    }
}
```
And remove the `abortControllerRef.current = null` line from `cancel()` — `abort()` is enough; the finally block handles cleanup.

---

**In `ShareWorkflowModal.tsx` `handleConfirm` (lines ~63–64)**

The catch block is literally `// ignore`. If `shareWorkflow()` fails for any reason (network error, permission denied, validation failure), the button returns to its normal state with no feedback and no error shown. The user has no way to know the share did not happen.

Fix: set an error state in the catch block and display it in the modal:
```ts
catch (err) {
    setError("Could not save sharing settings. Please try again.");
}
```

backend/schema.sql
willchen96: RLS enabled but no RLS policies need to be included as tables are accessed from the backend not frontend and nothing will be returned from client queries when no RLS policies are set.

willchen96:
Thanks at @amal66 for the review. All comments are incorporated into the PR with the exception of the inclusion of RLS policies since tables are only accessed from backend as commented above.

willchen96:
Additional changes were made incorporating @amal66 review and other changes.

- Refine CourtListener citation verification, bulk lookup logging, and API fallback behavior
- Persist cancelled chat stream output and render cancellation as the final assistant message
- Add document/version deletion safety fixes and shared warning/modal UI updates
- Sync document panel, case law panel, and response UI styling refinements
- Harden OSS sync script to preserve local env, dependency, and generated files

PR #173, opened 2026-06-10
Fix frontend build without Supabase env

lawyered0:
## Summary
- avoid constructing the Supabase browser client with empty env values during static prerender
- make token validation fail closed with a clear log when Supabase env is missing
- pin Turbopack's root to the frontend app to avoid workspace-root misdetection

## Verification
- npm run build --prefix frontend

## Notes
- This fixes a fresh-checkout build failure where `npm run build --prefix frontend` crashed on `/account/api-keys` with `supabaseUrl is required` before any runtime Supabase call was made.

### DIFF
--- a/frontend/next.config.ts
+++ b/frontend/next.config.ts
@@ -1,21 +1,24 @@
 import type { NextConfig } from "next";
+import path from "node:path";
 
 const nextConfig: NextConfig = {
-    /* config options here */
-    reactCompiler: true,
-    async rewrites() {
-        return [
-            {
-                source: "/sitemap.xml",
-                destination: "/api/sitemap/sitemap.xml",
-            },
-            {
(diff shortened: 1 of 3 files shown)


willchen96:
Thanks for the fix. I’m requesting changes because the fallback in frontend/src/lib/supabase.ts allows a production build without Supabase configuration and then points the browser client at http://localhost:54321. In a misconfigured deployment, user authentication or data requests could therefore be sent to a service running on the user’s own machine.
Please keep the fail-closed check in auth.ts and the Turbopack root fix, but remove the localhost fallback. CI builds should provide explicit placeholder environment variables—as the new CI workflow already does—or client creation should be deferred and fail clearly when configuration is missing.
The branch also currently conflicts with main, so it will need to be rebased and the frontend build reverified before approval.

PR #174, opened 2026-06-10
Confirm destructive folder deletes in chat explorer

lawyered0:
## Summary
- route assistant chat explorer folder deletes through a confirmation popup
- show how many nested folders and documents will be permanently deleted
- remove deleted documents from local chat explorer state and close stale document tabs after deletion

## Bug
The assistant chat project explorer called the recursive folder-delete endpoint immediately. That backend endpoint permanently deletes documents in the folder tree, but this view updated local state as if those documents were only moved out of the folder. A user could delete documents from the chat explorer without a destructive-action warning and keep seeing stale document entries until refresh.

## Verification
- NEXT_PUBLIC_SUPABASE_URL=http://localhost:54321 NEXT_PUBLIC_SUPABASE_PUBLISHABLE_DEFAULT_KEY=dummy npm run build --prefix frontend

Note: running the frontend build without placeholder Supabase public env values still hits the existing `supabaseUrl is required` prerender failure fixed separately in #173.


b1rdmania (approved):

This fixes a real bug, not just UX. The old `handleDeleteFolder` called the permanent recursive-delete endpoint but then set `folder_id: null` on the affected docs locally — so the UI pretended they'd been un-foldered while the backend had actually deleted them, leaving zombie doc entries and stale tabs. The new path removes the deleted docs from state, closes their tabs, and clears `activeTabId`/`selectedDocId` if they pointed at one. Confirmation popup reuses the existing `ConfirmPopup` and the impact counts (folders + nested docs) are computed correctly. Nice one.

amal66:
Could we please rebase latest main to resolve merge conflicts?

willchen96 (approved):
Made some changes:
- Added tests
- Edited document tab state manipulation to ensure that react has the latest selected document tab when resetting the tab after the delete api call
- Fixed merge conflict

PR #175, opened 2026-06-10
Require review owner for tabular settings edits

lawyered0:
## Summary
- require tabular review ownership before changing the review title
- require tabular review ownership before changing the review document list
- keep existing owner-only checks for columns, sharing, and project moves unchanged

## Bug
`PATCH /tabular-review/:reviewId` already restricted columns, sharing, and project moves to the review owner, but it allowed any user with review access to change `title` or `document_ids`. That let collaborators rename another user’s review or add/remove documents and cells by calling the API directly, even though the UI treats these as owner-controlled structural settings.

## Verification


willchen96 (commented):
LGTM

willchen96 (approved):

PR #176, opened 2026-06-10
Restrict project folder deletion to project owners

lawyered0:
## Summary
- block shared project members from calling `DELETE /projects/:projectId/folders/:folderId`
- keep folder-tree and document deletion behind project-owner authorization

## Why
The folder delete route previously accepted any user with project access. Because that endpoint recursively deletes the folder tree and removes every document inside it, a shared project member could permanently delete documents they do not own.

## Test


willchen96:
Reject in favor of [#193](https://github.com/willchen96/mike/pull/193), which is the newer/narrower version of the same project-folder owner restriction.

PR #177, opened 2026-06-10
Filter project documents before tabular generation

lawyered0:
## Summary
- filter fallback project documents through `filterAccessibleDocumentIds` before tabular generation
- prevent directly shared reviews from expanding into private project document access

## Why
`POST /tabular-review/:reviewId/generate` allows access through `ensureReviewAccess`, including direct `shared_with` review sharing. When a linked project review has no existing cells, the route falls back to loading every document in `review.project_id` without checking whether the requester can access those documents. A user who can access only the review could therefore trigger extraction over private project documents.

## Test


willchen96:
Thanks for the contribution and for spotting the issue. The intended behaviour is that non-members of projects should not have access to project specific tabular reviews and direct sharing of project chats and TRs should not be allowed. This will be fixed.

PR #185, opened 2026-06-19
fix: Google Workspace MCP connectors — durable refresh tokens + reliable OAuth popup

hordruma:
## Summary

Google Workspace MCP connectors (Drive / Gmail / Calendar on `*.googleapis.com`) were unusable past the first hour: they authorized once and then broke when the short-lived access token expired, and the OAuth connect popup could report a false failure. This PR makes the connection durable and the connect flow reliable.

## Problems

1. **No refresh token was ever obtained.** Google only issues a refresh token when the authorization request opts into offline access (`access_type=offline`) and is forced to re-prompt for consent (`prompt=consent`), and it does **not** implement the OIDC `offline_access` scope the MCP SDK would otherwise use. The SDK builds a spec-compliant authorization URL and exposes no hook for these proprietary params, so a Google connector got only an access token and broke as soon as it expired (the `encrypted_refresh_token` column stayed null, so there was nothing to renew with).

2. **OAuth popup completion was mis-detected under COOP.** Google's consent page is served with `Cross-Origin-Opener-Policy: same-origin`, which severs `window.opener` and makes `popup.closed` unreadable from the opener. As a result the callback's `window.opener.postMessage` never reached the app, and a COOP-blocked `popup.closed` read reports a **false** "closed" — surfacing a spurious *"OAuth authorization window was closed"* error even though the backend had already exchanged the code and stored the tokens.

## Changes

**Backend — `backend/src/lib/mcp/oauth.ts`** (commit `f9b5bf1`)
- Add `providerAuthorizationParams()` policy returning `{ access_type: "offline", prompt: "consent" }` for Google hosts, applied generically in `redirectToAuthorization()` (the SDK's only authorization-URL seam) — the redirect path carries no Google specifics, so future providers are a one-line policy change.
- Add/share `isGoogleOAuthHost()` with tightened matching (`=== googleapis.com` or `.googleapis.com`), so look-alikes like `notgoogleapis.com` are no longer treated as Google.
- Leave scope resolution to the SDK (documented): it falls back to the scopes the MCP server advertises; passing auth-server scopes would request the wrong (generic OIDC) scopes for Google.
- Align the frontend `isGoogleMcpConnector()` host check.
- Add backend tests (`oauth.test.ts`), an `npm test` script, and exclude test files from the build.

**Frontend — `connectors/page.tsx`** (commit `bfb5c11`)
- Detect OAuth completion by **polling the backend's `oauthConnected` flag** (the source of truth) instead of relying on COOP-broken `postMessage` / `popup.closed`. Keep `postMessage` as a fast path for providers that don't sever the opener, and drop the false-positive `popup.closed` rejection.

## Testing

- Backend unit tests: **7/7 pass**.
- Verified **end-to-end against a real Google Workspace account**:
  - The live backend builds the Google authorization URL with `access_type=offline` + `prompt=consent`, the correct Drive/Gmail scopes, PKCE (`S256`), and the right redirect URI.
  - Completed real consent → **refresh token obtained and stored** (previously-null `encrypted_refresh_token` now populated) — confirmed via both the API and the UI.
  - Live `tools/list` against `drivemcp.googleapis.com` returns the Drive tools.
  - The UI connect flow now completes cleanly with the popup fix (no false "window closed").

## Notes

- Write / destructive tools (e.g. Gmail `create_draft`) remain intentionally gated behind the existing "confirmation required" flag; this PR does not change that behavior.

🤖 Generated with [claude-flow](https://github.com/ruvnet/claude-flow)

https://claude.ai/code/session_01HSWSVK1PQozcrmZGgBhFXS


b1rdmania (approved):

Read through this properly — it's a clean fix and the reasoning in the comments is genuinely good. Two strengths worth calling out: the `access_type=offline` + `prompt=consent` explanation is correct (Google won't mint a refresh token otherwise, and doesn't honour the OIDC `offline_access` scope the SDK would use), and hardening `isGoogleOAuthHost` to match `googleapis.com` exactly / `.googleapis.com` as a suffix — so `notgoogleapis.com` and `googleapis.com.evil.test` no longer slip through — is a nice catch the old `endsWith("googleapis.com")` missed. Treating the backend `oauthConnected` flag as source of truth is also the right move given COOP severs `window.opener`. Tests cover the host-matching and param cases. LGTM.

amal66:
Thanks for a genuinely well-researched fix. Both root causes are correctly diagnosed, and I verified them against the installed SDK: Google really does ignore `offline_access` (you need `access_type=offline` + `prompt=consent`), and COOP on the consent page severs `window.opener`, so `popup.closed` false-positives. Nice bonus: the `isGoogleOAuthHost` tightening quietly closes a real bug — a lookalike domain like `notgoogleapis.com` previously matched the `endsWith` check and would have been handed the deployment's Google client secret.

One must-fix: reconnecting an already-broken connector (expired access token, no refresh token — the exact fleet this PR targets) resolves as a phantom success. `oauthConnected` only checks that an encrypted access token exists, and the SDK returns REDIRECT without invalidating the stale row, so the first poll resolves, the popup closes mid-consent, and the user loops forever. The fix is ~3 lines: invalidate stored tokens whenever the flow returns an auth URL.

Smaller items: closing the popup mid-consent now hangs the modal until the 5-minute timeout (needs a cancel affordance); the 1.5s poll has no unmount cleanup or backoff (~200 authenticated requests); the new error path in `servers.ts` returns remote-authored text without the untrusted-content wrapper; and the `errors.ts`/`servers.ts` commits are unrelated to OAuth — worth a line in the description.

I've opened https://github.com/hordruma/mike/pull/1 against your branch implementing all of this plus tests — merge or cherry-pick as you prefer. It sits directly on your branch, three focused commits, with rebasing onto current `main` left to you; both suites pass on it (backend 279, frontend 40) and both typechecks are clean.

amal66 (commented):
Full write-up in my comment below — one must-fix (reconnecting an already-broken connector resolves as a phantom success) plus a few smaller items, with a follow-up PR against your branch implementing them.

amal66:
Closing as it is stale and superseded by https://github.com/Open-Legal-Products/mike/pull/285

PR #193, opened 2026-06-28
Restrict project folder deletion to owners

lawyered0:
## Summary
- Require project ownership before deleting a project folder
- Prevent shared project members from deleting folders and nested documents

## Verification
- git diff --check


b1rdmania (approved):

Correct and minimal. Shared members previously passed the `access.ok` check and could delete an owner's folders plus every nested doc; the added `isOwner` guard closes that, and returning 404 (not 403) matches the existing pattern in this file. One follow-up worth doing in a separate pass: a quick grep of the sibling mutate/delete routes for the same missing `isOwner` guard — this is unlikely to be the only one.

willchen96 (approved):

willchen96:
Thanks good spot.

PR #194, opened 2026-06-28
Persist MCP OAuth discovery metadata

lawyered0:
## Summary
- Persist MCP OAuth discovery metadata on token rows
- Preserve token endpoint, authorization server, and resource values across client registration and token saves
- Allows OAuth token refresh after access-token expiry without forcing reconnect

## Verification
- git diff --check
- Targeted TypeScript check: ./node_modules/.bin/tsc --noEmit --pretty false --skipLibCheck --moduleResolution node16 --module node16 --target es2022 src/lib/mcp/oauth.ts

### DIFF
--- a/backend/src/lib/mcp/oauth.ts
+++ b/backend/src/lib/mcp/oauth.ts
@@ -2,6 +2,7 @@ import crypto from "crypto";
 import {
     auth as runMcpOAuth,
     type OAuthClientProvider,
+    type OAuthDiscoveryState,
 } from "@modelcontextprotocol/sdk/client/auth.js";
 import type {
     OAuthClientInformationMixed,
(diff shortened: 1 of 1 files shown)


amal66:
Thanks — this fixes a real bug. Before this PR, `saveTokens` never persisted `authorization_server`/`token_endpoint`, so refresh always threw and every access-token expiry forced a full reconnect. The implementation is careful, too: `saveDiscoveryState` is a real SDK hook, the `/token` fallback matches the SDK's own, and persisted URLs are validated on both write and read.

One thing I need fixed before merging. Since the provider has no `discoveryState()` cache, every OAuth flow re-runs discovery, and `saveDiscoveryState` silently patches `authorization_server`/`token_endpoint` while leaving the existing refresh token and client secret in place. A later-compromised MCP server can therefore point discovery at an attacker-controlled authorization server, and the next refresh POSTs the real refresh token plus the decrypted client secret straight to it. This path didn't exist before the PR — refresh never fired. Fix: if the stored authorization-server origin changes during re-discovery, clear the stored tokens/secret and force a re-authorize instead of swapping endpoints.

Smaller items: please add tests for the persistence and refresh behavior (main now has vitest, and #185 adds an `oauth.test.ts` you could extend), rebase onto current main so CI runs them, make `patchOAuthTokenRow` an atomic upsert rather than read-then-insert, and persist the sanitized URL `validateRemoteMcpUrl` returns, not the raw input.

I've put all of this in https://github.com/lawyered0/mike/pull/1 against your branch — fixes plus tests, two focused commits sitting directly on your branch. Merge it in or cherry-pick what you agree with; rebasing onto current `main` is left to you. Sorry it appeared earlier without any context — this comment is the explanation it was missing. `tsc` and the build are clean on it; the tests ship with it and run once you're on main's vitest setup (I verified them green there).

willchen96:
Closing this as it is covered in newly PR #285

PR #196, opened 2026-06-29
Self-hosted Docker stack (embedded Supabase + storage + email) and local Ollama models

thijsdezoete:
## Summary

A single `docker compose up` brings up the entire app with **no external accounts** — a more automated alternative to the CLI-based approach in #44, plus local LLM support.

### Self-contained stack
- **Embedded Supabase** in compose: Postgres + GoTrue auth + PostgREST + an nginx gateway. Schema auto-loads on first boot; no `supabase` CLI needed.
- **RustFS** for S3-compatible object storage, replacing Cloudflare R2 (the backend already uses path-style S3, so it's drop-in).
- **Mailpit** catches auth emails (signup/confirmation) in a local inbox at `:8025`.
- Production-style image builds (not bind-mount + `npm install` on start).

### Local models via Ollama
- New provider speaking Ollama's OpenAI-compatible chat API, with the agentic tool-call loop and a graceful fallback for models that don't support tools.
- Models **detected dynamically** (`GET /models/ollama`) and shown in every picker (chat, title, tabular) under a **Local** group — no API key. Any `ollama/<tag>` id is valid by prefix.
- A **Refresh** button on the API Keys page re-detects installed models and re-checks key status, live to the pickers.

### How this differs from #44
- One command, all-in-one — no `supabase start`, no fork of the Supabase CLI stack.
- Embedded Supabase + RustFS + Mailpit vs. relying on the CLI's Supabase + Storage.
- Adds local LLM (Ollama) support, which #44 doesn't have.

## Test plan
- [x] `docker compose up --build` brings up all services; schema auto-applies (23 public tables)
- [x] End-to-end signup → Mailpit confirmation email → confirm → login
- [x] Chat streams from `ollama/qwen3.6`; `phi3:mini` falls back to no-tools
- [x] `GET /models/ollama` returns installed models (401 unauthenticated)
- [x] Ollama models selectable as chat + tabular review models


amal66:
Good idea. For future reference, we should keep PRs atomic to just one change so that we can better review and merge each change. 

Before reviewing, we should fix all e2e tests before marketing it ready for review. Helped with that here, could you please take a look and merge if ready? https://github.com/thijsdezoete/mike/pull/1

I'll review this once that's merged and e2e passes here

amal66:
Since this has been stalled on the CI failure (fix proposed at thijsdezoete/mike#1 but not merged there), I've re-submitted this stack from a maintainable branch as #282 — all six commits carried over verbatim with @thijsdezoete's authorship preserved, plus that one e2e fix on top. If you're back, happy to close #282 in favour of applying the fix here instead.

willchen96:
Thanks both - closing this as #282 has been merged.

PR #220, opened 2026-07-20
[Testing 12] test: Playwright e2e suite + CI merge gate (auth, chat, projects, tabular, workflows)

amal66:
> **Merge-order note:** please merge #212 (the missing `chat_messages.workflow` column fix) before this. The cold-load chat spec currently has to test around that bug — messages silently fail to persist on a fresh database — by asserting the empty-chat redirect instead of rendered messages. Once #212 lands, that spec will be strengthened to assert real message rendering.

## Summary

Adds a Playwright end-to-end suite plus the CI workflow that runs it on every PR, so changes can be gated on a real browser-driven pass. 27 specs across 7 files cover auth flows, the critical path (create project → upload PDF → ask a question → streamed answer), chat management, project management, tabular reviews, and workflows/account settings.

**Verified green against this repo's own stack**: the workflow builds a full local environment from `backend/schema.sql` + `backend/migrations/` and runs the suite — latest run **23 passed / 4 skipped / 0 failed in ~7 min**.

## Changes

- `e2e/` — the specs plus `auth.setup.ts` (idempotently bootstraps a confirmed `e2e@mike.local` via the Supabase admin API and saves Playwright `storageState`) and `llm.ts` (skip-guard helper).
- `playwright.config.ts` — single worker (shared backend + one test user), a `setup` project feeding a `chromium` project; local `webServer` boots `backend/` and `frontend/`, disabled under CI (the workflow owns the servers).
- `.github/workflows/e2e.yml` — on every `pull_request`, boots MinIO + local Supabase (loads `schema.sql`, applies migrations, grants `service_role`, reloads PostgREST), builds and serves a **production** web build, then runs Playwright and uploads the report/traces.
- Root `package.json` / `package-lock.json` / `tsconfig.json` — Playwright dev-deps, `test:e2e*` scripts, and typechecking scoped to `e2e/`. `docs/e2e-ci.md` documents setup and the branch-protection step. `.gitignore` entries for Playwright artifacts.

## Why

The repo has no end-to-end coverage today; a regression in auth, chat, projects, tabular, or workflows ships silently. This makes those flows a merge gate. Two decisions worth calling out:

- **Production build, not `next dev`** — dev's on-demand compilation and hydration-error overlay make the suite flaky; a production build is deterministic.
- **Schema + migrations + `service_role` grant** — `schema.sql` alone lags the migrations (e.g. `GET /workflows/:id` queries `workflow_open_source_submissions`, only a migration creates it) and, loaded via raw `psql`, leaves `service_role` without table grants; the workflow reconciles both so CI matches production behavior.

## Testing

- CI: `e2e / playwright` — **23 passed, 4 skipped, 0 failed** (the 4 skips are the LLM-dependent specs; see below).
- `npx tsc --noEmit` typechecks the suite.
- Locally: `npm ci && npx playwright install --with-deps chromium && npm run test:e2e` against a running stack.

**Optional secret** — the 4 LLM-dependent specs (chat send / critical-path) `test.skip` when `ANTHROPIC_API_KEY` is unset, so the suite is green without any secret; set the key to also run and enforce them.

**To make it merge-blocking**, mark `e2e / playwright` a required status check in branch protection — steps and a `gh api` one-liner are in `docs/e2e-ci.md`. The workflow reports pass/fail on its own; only branch protection turns a red run into a blocked merge.

## Flakiness root-cause fix (update)

An earlier revision of these specs carried retry loops and settle-waits around intermittent local-Supabase gateway 5xxs. That flakiness has now been root-caused and fixed, and the scaffolding removed — the suite passes with none of it.

**Diagnosis.** `useDirectoryData` fired one `GET /projects/:id` per existing project every time a directory modal opened (Add Documents / New Project / New Review / Use Workflow). Each request costs a GoTrue auth check plus ~6 PostgREST queries, so the burst scaled ~8x with project count. Reproduced against the Supabase CLI stack with a 43-project account: overlapping modal-open storms drove GoTrue into Postgres connection exhaustion (`failed to connect ... user=supabase_auth_admin ... context deadline exceeded`), yielding 467x 500 + 641x 504 on `/auth/v1/user` in one run — 38/48 creates and ~70% of the GETs failed. On CI-class hardware Kong surfaces the same collapse as the 502s the retries papered over.

**Fix.** #238 (`olp-pr/directory-fetch-storm`) — `GET /projects?include=documents` returns each project's documents from one batched query, and `useDirectoryData` uses it: one request per modal open regardless of project count. After the fix the identical storm harness produced zero failures (96/96 OK). The same commit is cherry-picked onto this branch so the suite passes here; git dedupes it when both merge.

**Retry scaffolding removed** (`test(e2e): drop transient-5xx retry scaffolding...`): create-with-retry loops (project x2, chat x2, workflow, tabular-review, review add-document), reload-guards for transient "Project not found", the networkidle settle-waits that let the fetch burst drain before submitting, and the storm-sized 180s test timeouts. **Kept**: the title-generation settle logic, two bounded retries that guard documented client-side hydration races unrelated to the gateway (chat submit model seeding; account display-name), and Playwright's own `retries: 2` on CI.

**Evidence.** With the fix and the scaffolding stripped, 3 consecutive full-suite runs (after 2 pre-de-retry runs with the fix) were green — 23 passed / 4 LLM-skipped each time, zero Playwright retries — against a production build and an account seeded with ~40 projects, a larger directory fan-out than CI's fresh database ever produces.

## LLM specs verified (update)

The 4 LLM-gated specs previously selected a "Demo (no key needed)" model that exists only in a fork's demo provider — not in this repo's `ModelToggle.MODELS` — so setting the `ANTHROPIC_API_KEY` secret unskipped them and they then failed at model selection. Fixed in `test(e2e): select a real Claude model in LLM-gated specs`: the helper now picks **Claude Sonnet 4.6** (the cheapest Anthropic entry in the toggle), and critical-path Step 8 asserts a nonempty streamed assistant answer (presence-based, deterministic against real LLM output) instead of the fork's canned "Demo mode" reply.

Verified locally against a production build + Supabase CLI stack + MinIO with only `ANTHROPIC_API_KEY` configured:

- keyless Playwright env: **23 passed / 4 skipped** (skip guards intact)
- `ANTHROPIC_API_KEY` exported: **27 passed / 0 skipped** (1.2m, zero retries)
- `npx tsc --noEmit` clean

With the secret set, a full run makes ~4 Sonnet 4.6 chat turns plus up to 4 Haiku 4.5 title generations — a few cents per run.


willchen96:
### Notes on the latest push (`d4448fa..fb3ec2d`)

Beyond rebasing on current `main` (clean merge, `c196f9b` — picks up the backend integration tests, lint burndown, and the narrowed `service_role` schema grants), this adds:

- **`62da94b` — self-bootstrapping local e2e stack.** `scripts/e2e-local-stack.sh` (exposed as `npm run test:e2e:local`, and also wired into the Playwright `webServer` hook so plain `npm run test:e2e` works too) checks Docker, starts local Supabase, loads `schema.sql` + migrations + grants, and points the env files at the local stack. It also raises the backend `RATE_LIMIT_*` caps the way CI already does: a full suite run exceeds the default 300-requests/15-min general cap, after which every API call 429s and profile/list waits time out — this was the root cause of an intermittent `display name` spec failure that only reproduced in full-suite runs.
- **`f3173a9` — narrowed post-migration grants.** `main`'s `schema.sql` now self-grants `service_role` a deliberately narrowed privilege set, so the old `GRANT ALL` workaround in `e2e.yml` would have widened the test database beyond what prod gets (letting e2e pass on privileges a real deployment lacks). The post-migration re-grant stays — `schema.sql`'s `GRANT ... ON ALL` can't cover tables the dated migrations create — but now uses the identical narrowed set. Verified on a wiped database: no permission-denied anywhere.
- **`cb2306c` — hydration-mismatch fix.** The provider Suspense fallback, MFA gate loader, and pages-layout auth loader were three near-identical spinners with different Tailwind classes; server and client can resolve those gates differently on first paint, which logged a hydration mismatch on nearly every dev page load (noise in e2e logs). Unified into one shared `FullScreenLoader`.
- **`4728fd1` / `fb3ec2d` — housekeeping.** Pin `turbopack.root` (the root + frontend lockfile pair made Next.js guess the workspace root and warn on every dev-server start), and gitignore the `supabase init` scaffold that the local stack creates under `backend/supabase/`.

Verification: full suite against a fresh `supabase db reset` local stack — **23 passed / 4 skipped (LLM-gated, no key) / 0 failed in 46.5s**.

willchen96 (approved):
Tested and made changes in additional commits to solve some errors impeding the tests.

PR #221, opened 2026-07-20
[Feature 14] feat: Word add-in — chat, tracked-changes rewrites, workflows and projects in Word

amal66:
## Summary

Adds a Microsoft Word add-in — an Office.js task pane that brings Mike's core surfaces directly into Word: chat with the assistant, apply tracked-changes redlines to the open document, run document actions (proofread, anonymise, improve writing, draft clauses), and use workflows and projects without leaving the editor.

The guiding rule for this PR is **the add-in should be the web app, shaped to a task pane**. Where the web app has a pattern — visual or architectural — the add-in copies it rather than inventing its own. Shared pieces are deliberately **duplicated** into `word-addin/src/` for now (the add-in ships standalone, with no Next.js); the copies are kept faithful precisely so a later PR can extract them into a shared package mechanically.

Everything lives under a single new top-level directory, **`word-addin/`**, and talks to the existing backend API. One deliberate exception to "purely additive": a small backend change (the `document_context` field on `POST /chat` + spotlighting of it) — without it the add-in's document-context feature silently does not work. Details below.

## How close is it to the main app?

Since you built the frontend, here is the parity map, honest about both directions.

**Identical or copied verbatim:**
- `ui/button.tsx`, `ui/input.tsx`, `lib/utils.ts`, `chat/mike-icon.tsx` — byte-identical copies (only the `cn` import path differs).
- `ui/toggle-switch.tsx`, `ui/tab-pill-button.tsx` — verbatim copies of your components; the pane's toggles are the same blue-track switch and the tab bar is the same glass pills as the web.
- `tokens.css` — a verbatim subset of `globals.css` (same oklch values, same `@theme` mapping, same font variables; EB Garamond serif answers, Inter chrome).
- The transcript surface — `EventBlocks` (dot/connector), `PreResponseWrapper` ("Working… / Completed in N steps" strip), `UserMessage` bubble, `PillButton`, `EditCard`/`EditCardsSection` (green replacement / red strikethrough, collapsible "N tracked changes"), `messageStyles` glass constants — same Tailwind class strings as `frontend/src/app/components/assistant/`, trimmed to what the pane can honestly report. Each file's header says exactly what it was duplicated from.
- The composer is the web composer: same liquid-glass container, same gradient-black square send/stop button.

**Architecturally the same:**
- Same endpoints, same `Bearer` auth, same SSE framing rules (`data:` frames, `[DONE]` terminal, trailing-error tolerance) as `useAssistantChat`.
- **Workflows run exactly like the web:** the message carries a `workflow: { id, title }` reference and the backend resolves `skill_md` server-side (via the workflow store + `read_workflow`), inside the semi-trusted `<workflow-instructions>` fence. The pane never pastes a workflow body into the user turn — an earlier iteration did, which both bypassed the fence (a shared workflow's body arrived fully trusted) and guaranteed drift (server-side prompt improvements would never reach Word). Fixed before this lands.
- **Document text never rides in the prompt:** the Chat tab's "Use document as context" toggle, the Workflows tab, and the Proofread/Anonymise actions all send the open document via `document_context`, which the backend nonce-fences as untrusted reference content — the same spotlighting treatment your document flows get.

**Word-specific, no web counterpart (new code, not divergence):**
- `useWordDoc` (Office.js read / insert / tracked-edit application), `lib/redline.ts` + `lib/wordText.ts` (ORIGINAL/REPLACEMENT/REASON parsing and Word text matching). Applying puts genuine tracked changes into the document; Word's Review ribbon owns accept/reject.
- Auth is a password-grant + refresh-token store on `OfficeRuntime.storage` — a webview pane can't host supabase-js's browser session model — feeding the same `Authorization` header the web sends.
- Manifest, commands, webpack toolchain, Office-mock e2e harness, live Word-online recorders.

**Known deltas (deliberate, with the follow-up named):**
- The Actions tab (Improve Writing / Anonymise / Draft Clause) uses pane-local prompts rather than builtin workflows. Backing them with builtins is the right end state, but new builtins live in the external `mike-workflows` repo (`systemWorkflows.ts` is generated from it) — and `builtin-proofread` outputs an issues *table*, while the pane's Proofread must produce redline edits it can apply as tracked changes. Follow-up: Word-flavored builtins in `mike-workflows`, then the pane switches to references like the Workflows tab. The prompts already send the document via fenced `document_context`, so only the instruction source moves.
- Some pane chrome uses shadcn semantic tokens (`text-muted-foreground` etc.) where the web uses literal `gray-*`; both resolve to the same neutrals. Worth unifying on one vocabulary in the shared-package PR — happy to go either direction.
- The markdown renderer is a trimmed adaptation of `MarkdownContent` (no KaTeX/citations — the pane doesn't render those yet).
- No `/slash-command` workflow autocomplete in the pane's composer yet; that wants the shared `ChatInput` extraction.

## Backend change (small, no-op for the web app)

The add-in posts the open document's text to `POST /chat` as **`document_context`** (snake_case, matching `chat_id` / `displayed_doc` / `attached_documents`; `documentContext` is accepted as a deprecated alias for older pane builds):

- `backend/src/routes/chat.ts` parses the optional field (400 on non-strings; trimmed; capped at 200k chars) and passes it into the system prompt via `buildMessages`'s existing `systemPromptExtra` parameter — the same mechanism project chat uses.
- `backend/src/lib/chat/contextBuilders.ts` builds the block. The document body is user-controlled text and a prompt-injection vector, so it enters the prompt nonce-fenced ("spotlighting": smuggled fence tokens HTML-encoded, echoed nonces redacted) behind an instruction that it is reference content, not instructions. The fence uses the **same per-request nonce** as filenames and tool outputs — one nonce per request, as the system-prompt policy states.
- Integration tests pin the 400 shape, the alias, and that the document fence carries the request nonce; unit tests cover parsing, the fence, and injection through `buildMessages`.

Requests that don't send the field (the web app) are unaffected.

## Demo — every button, against real Word on the web

`word-addin/e2e-live/word-web-full-demo.mjs` drives the sideloaded pane in **real Word online** (word.cloud.microsoft) and records video: it types a five-clause contract with one deliberate flaw per feature, then exercises **every actionable control** — chat "Apply 2 tracked edits" / "Insert below cursor" / "Insert below (tracked)", Improve Writing (tracked + plain replace), Proofread → apply corrections, Anonymise → apply redactions, Draft Clause (both inserts), Workflows run + insert, Projects, and Sign out. Every apply lands genuine Word tracked changes (strikethrough + insertion, accept/rejectable in the Review ribbon).

To reproduce (no funded LLM key needed — `e2e-live/anthropic-stub.mjs` speaks the Anthropic streaming protocol with scripted responses; everything else is real: Supabase auth, this backend, SSE, the Word JS API):

```
node word-addin/e2e-live/anthropic-stub.mjs &            # :4141
ANTHROPIC_BASE_URL=http://127.0.0.1:4141 npm run dev     # backend, in backend/
npm run dev                                              # add-in dev server, in word-addin/
node word-addin/e2e-live/word-web-session.mjs --login    # one-time Microsoft sign-in
node word-addin/e2e-live/word-web-full-demo.mjs          # records video + per-step stills
```

Demo videos are attached in the PR comments.

## Testing

- `backend/`: `npx tsc --noEmit` clean; full vitest suite green on this branch (440 passed / 14 skipped), including the `document_context` unit + integration coverage.
- `word-addin/`: `tsc --noEmit` clean; **64 hermetic browser scenarios** (mocked Office.js host + backend, own Playwright harness under `word-addin/`) all pass — including request-body assertions that workflows travel as `{ id, title }` references and that `document_context` is sent when the toggle is on and omitted when off. The restyle changed no role, aria-label, or status string, so the suite exercises the new UI unmodified.
- Live: the full-demo recording ran end-to-end against real Word online with zero step failures.

Happy to split this further or adjust where it lives if you'd prefer a different layout — it's deliberately isolated so that's easy.


amal66:
Pushed `dfee7c1`, which closes the biggest functional gap: **the add-in can now make tracked changes to existing document text**, not just to a manually selected range or via new-paragraph insertion.

- **Proofread** and **Anonymise** now stream results as `ORIGINAL` / `REPLACEMENT` / `REASON` blocks (readable while streaming, parseable when done) and offer **Apply N corrections/redactions (tracked)**.
- New `applyTrackedEdits()` locates each original snippet with Word's search API (exact, case-sensitive, every occurrence) and replaces it under `changeTrackingMode = TrackAll`, restoring the user's prior tracking mode — so every AI edit lands as a genuine accept/rejectable redline in the Review tab.
- Fail-honest: snippets that are unsearchable (multi-paragraph or >255 chars, Word search limits) or no longer present (document edited after the scan) are skipped and reported ("Applied 2 of 3 … 1 skipped"), never fuzzily matched.

Verification: TypeScript clean, production webpack build green, hermetic Playwright suite 61/61 (4 new specs covering tracked application, stale-text skip reporting, the no-issues path, and tracked PII redaction against the recorded Office.js shim calls).

amal66:
Demo — every button exercised in real Word on the web (recorded 2026-08-03, unified UI):

[mike-word-addin-full-demo.webm](https://github.com/user-attachments/assets/19bb0fb3-962a-48a5-915d-89deccd05ed1)

willchen96 (approved):

willchen96:
Fantastic addition!

PR #242, opened 2026-07-23
[Security 1/9] SSRF guardrails for server-side connector fetches

amal66:
## [Security 1/9] SSRF guardrails for server-side connector fetches

> Part of the split of #227 into reviewable, single-topic PRs. Each stands alone and is scoped to one attack class. Index: see the tracking comment on #227.

### TL;DR
When a user registers an MCP connector, **our server** fetches the URL they gave us. This PR stops that fetch from being pointed at internal infrastructure (the cloud metadata service, the database, admin panels) by validating the *resolved IP address* against every private/reserved range and failing closed. It also closes the DNS-rebinding race by pinning the connection to the address we validated.

---

### Risk to user data
**Severity: critical.** A successful SSRF here is a direct path to *total* data compromise, not just one user's records:
- The cloud metadata endpoint (`http://169.254.169.254/…`) hands out the server's IAM credentials. With those, an attacker reads the entire object store and database — **every tenant's documents**, not their own.
- Internal addresses (`localhost:5432`, private admin panels) are reachable from the server but not from the public internet, so this bypasses the network perimeter entirely.

The attacker only needs an ordinary app account and the "add a connector" feature.

### Flows affected
- MCP connector registration / validation (`validateRemoteMcpUrl`)
- Every outbound connector transport fetch
- OAuth discovery, dynamic client registration, and token refresh for connectors (`mcp/oauth.ts`) — previously these used raw `fetch` and were unguarded SSRF sinks.

### Attack precedent
- **Capital One, 2019** — 106M records. An SSRF reached `169.254.169.254`, retrieved IAM role credentials, and used them to read S3. This exact primitive. ([Krebs](https://krebsonsecurity.com/2019/07/capital-one-data-theft-impacts-106m-people/))
- SSRF is its own category in the [OWASP Top 10 (A10:2021)](https://owasp.org/Top10/A10_2021-Server-Side_Request_Forgery_%28SSRF%29/).

```mermaid
sequenceDiagram
    participant A as Attacker (ordinary user)
    participant S as Mike backend (inside the VPC)
    participant M as Metadata service<br/>169.254.169.254
    A->>S: Register connector<br/>url = http://169.254.169.254/latest/meta-data/
    S->>M: server-side GET (from inside the VPC)
    M-->>S: IAM credentials
    S-->>A: response relayed back
    Note over S,M: With this PR: 169.254.169.254 is link-local →<br/>refused before a socket opens
```

| Option | Why not / why |
|---|---|
| Blocklist hostnames (`localhost`, `metadata`) | Trivially bypassed: raw IPs, DNS names that resolve to internal IPs, `0x7f.0.0.1`, IPv6 forms. Rejected. |
| Allowlist of known-good connector domains | Safest, but MCP is explicitly "bring your own server" — an allowlist defeats the feature. Rejected for now (a good future opt-in for locked-down tenants). |
| **Validate the resolved IP against all private/reserved ranges, fail closed** | **Chosen.** Works for arbitrary user URLs, no maintenance of a domain list. The subtlety is doing it *correctly* — see below. |

**The correct-implementation details that matter:**
- Validate the **resolved IP**, not the URL string. `http://internal.evil.com/` looks public but resolves to `10.0.0.5`.
- Cover **every** reserved range: loopback, RFC1918, link-local, CGNAT (`100.64/10`), multicast, and the **IPv6 forms of all of them** — including IPv4 addresses *embedded in* IPv6 literals (`::ffff:10.0.0.1`, `::7f00:1`), the classic filter bypass. `privateIp.ts` expands these.
- **Fail closed:** anything unparseable or unrecognized is treated as blocked.
- **Never auto-follow redirects.** A public URL that `302`s to `http://localhost/` defeats a naive check — so `guardedFetch` sets `redirect: "manual"`.

The tests in `client.ssrf.test.ts` encode the bypass encodings, not just the happy path.

### The subtle part: DNS rebinding (a TOCTOU race)
Validating the IP and *then* fetching the URL is two separate DNS lookups. An attacker who controls DNS for `evil.example` (TTL 0) answers the first lookup — the validation — with a harmless public IP, and the second lookup — the real fetch — with `127.0.0.1`. This is a classic [time-of-check-to-time-of-use](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) race: **the thing you checked is not the thing you used.**

```mermaid
sequenceDiagram
    participant D as Attacker DNS (TTL 0)
    participant S as Naive server
    S->>D: resolve evil.example (validation)
    D-->>S: 93.184.216.34 — public, check passes ✅
    S->>D: resolve evil.example (actual fetch)
    D-->>S: 127.0.0.1 — check already passed ❌
    S->>S: connects to 127.0.0.1
```

**Fix:** resolve **once**, validate those addresses, and connect to **exactly** them. This PR does it with a per-request [undici](https://undici.nodejs.org/) `Agent` whose `connect.lookup` hook runs the private-IP guard *at the moment the socket opens* and hands undici only validated addresses — so the address we validate is the address we connect to. The URL hostname is untouched, so the `Host` header and TLS SNI still reflect the real host and certificate verification works normally against legitimate public servers.

### What's in this PR
- `backend/src/lib/privateIp.ts` — conservative IP classifier (new, shared).
- `backend/src/lib/mcp/client.ts` — `validateRemoteMcpUrl`, `guardedFetch`, and the connect-time-pinned undici dispatcher.
- `backend/src/lib/mcp/oauth.ts` — routes all discovery/registration/refresh through `guardedFetch`.
- `undici` dependency; tests in `client.ssrf.test.ts` (9 tests, incl. embedded-IPv4-in-IPv6 bypasses).

### Reading
[PortSwigger: SSRF](https://portswigger.net/web-security/ssrf) · [OWASP SSRF Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Server_Side_Request_Forgery_Prevention_Cheat_Sheet.html)

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -29,6 +29,7 @@
         "multer": "^1.4.5-lts.2",
         "pdfjs-dist": "^4.10.38",
         "resend": "^4.5.1",
+        "undici": "^6.27.0",
         "xlsx": "https://cdn.sheetjs.com/xlsx-0.20.3/xlsx-0.20.3.tgz",
         "zod": "^3.25.76"
       },
@@ -6953,6 +6954,15 @@
       "resolved": "https://registry.npmjs.org/underscore/-/underscore-1.13.8.tgz",
       "integrity": "sha512-DXtD3ZtEQzc7M8m4cXotyHR+FAS18C64asBYY5vqZexfYryNNnDc02W4hKg3rdQuqOYas1jkseX0+nZXjTXnvQ==",
       "license": "MIT"
+    },
+    "node_modules/undici": {
+      "version": "6.27.0",
+      "resolved": "https://registry.npmjs.org/undici/-/undici-6.27.0.tgz",
+      "integrity": "sha512-YmfV3YnEDzXRC5lZ2jWtWWHKGUm1zIt8AhesR1tens+HTNv+YZlN/dp6G727LOvMJ8xjP9Be7Y2Sdr96LDm+pg==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=18.17"
+      }
     },
     "node_modules/undici-types": {
       "version": "6.21.0",
--- a/backend/package.json
+++ b/backend/package.json
@@ -31,6 +31,7 @@
     "multer": "^1.4.5-lts.2",
     "pdfjs-dist": "^4.10.38",
     "resend": "^4.5.1",
+    "undici": "^6.27.0",
     "xlsx": "https://cdn.sheetjs.com/xlsx-0.20.3/xlsx-0.20.3.tgz",
     "zod": "^3.25.76"
   },
--- a/backend/src/lib/__tests__/privateIp.test.ts
+++ b/backend/src/lib/__tests__/privateIp.test.ts
@@ -1,0 +1,76 @@
+import { describe, expect, it } from "vitest";
+import { isBlockedIp, isPrivateIpv4, isPrivateIpv6 } from "../privateIp";
+
+describe("private/reserved IP classification", () => {
+    it.each([
+        "0.0.0.0",
+        "10.0.0.1",
+        "100.64.0.1",
+        "127.0.0.1",
+        "169.254.169.254",
+        "172.16.0.1",
+        "192.0.0.1",
+        "192.0.2.1",
+        "192.88.99.1",
+        "192.168.0.1",
+        "198.18.0.1",
+        "198.51.100.1",
+        "203.0.113.1",
+        "224.0.0.1",
+        "240.0.0.1",
+        "255.255.255.255",
+    ])("blocks non-global IPv4 address %s", (ip) => {
+        expect(isPrivateIpv4(ip)).toBe(true);
+        expect(isBlockedIp(ip)).toBe(true);
+    });
+
+    it.each(["8.8.8.8", "93.184.216.34", "192.31.196.1"])(
+        "allows globally reachable IPv4 address %s",
+        (ip) => {
+            expect(isPrivateIpv4(ip)).toBe(false);
+            expect(isBlockedIp(ip)).toBe(false);
+        },
+    );
+
+    it.each([
+        "::",
+        "::1",
+        "::ffff:8.8.8.8",
+        "::8.8.8.8",
+        "100::1",
+        "2001::1",
+        "2001:2::1",
+        "2001:10::1",
+        "2001:db8::1",
+        "2002::1",
+        "3fff::1",
+        "5f00::1",
+        "fc00::1",
+        "fd00::1",
+        "fe80::1",
+        "fec0::1",
+        "ff02::1",
+        "64:ff9b::10.0.0.1",
+        "64:ff9b:1::1",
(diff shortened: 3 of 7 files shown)


willchen96:
Implemented the requested review fixes in commit [`[65cbf0e](https://github.com/Open-Legal-Products/mike/pull/242/commits/65cbf0eef60e88dd584af2d85f94fbec1dcc4ebf)`](https://github.com/Open-Legal-Products/mike/pull/242/commits/65cbf0eef60e88dd584af2d85f94fbec1dcc4ebf):

- Hardened IPv4/IPv6 classification to fail closed and block non-global ranges, including `fec0::/10`, multicast, documentation, transition, and reserved addresses.
- Replaced the per-request Undici `Agent` with a shared guarded agent, preserving connection-time DNS validation while enabling connection pooling and avoiding unmanaged agent/socket churn.
- Added regression coverage for private/reserved address classification, malformed addresses, NAT64 cases, and dispatcher reuse.

Verification completed:

- TypeScript build passed.
- Focused SSRF tests: 53 passed.
- Full backend suite: 312 passed, 5 skipped.
- Live guarded HTTPS smoke test returned 200.

willchen96 (approved):
LGTM

amal66:
Retrospective (2026-08-27), for anyone reading this PR's history:

My original `isPrivateIpv6` in this PR was structured as a **blocklist and failed open**. It ended in `return false` and treated unparseable input as safe (`if (!groups) return false`), so any IPv6 range I hadn't explicitly enumerated — site-local `fec0::/10` and multicast `ff00::/8` among them — was classified as globally reachable and allowed, while the file's own doc comment claimed "anything unparseable or unrecognized is treated as blocked". I also shipped the classifier with no direct unit tests; my nine integration tests all probed addresses I *had* enumerated, so they passed against the fail-open version.

@willchen96's pre-merge commit (65cbf0ee, minutes before the merge) is what made the shipped version sound: it inverted the classifier to a `2000::/3` global-unicast **allowlist** over `net.BlockList` RFC tables, made every unrecognized or unparseable input fail **closed**, and added `privateIp.test.ts`. The parts of the PR that survived that rewrite — the connect-time DNS-pinning dispatcher and the embedded-IPv4/NAT64 expansion — are mine; the fail-closed classifier design is his. Credit where due.

What I took from it, concretely:

- **For guard code, the default disposition *is* the design.** An incomplete enumeration is expected; the default catching what you missed is the entire discipline. Blocklist-with-`return false` means every range you didn't think of is a hole.
- **Tests for a guard have to attack the unenumerated case**, not confirm the enumerated ones. One test asserting `isPrivateIpv6("fec0::1")` or even `isPrivateIpv6("not-an-ip")` blocks would have caught this before review.
- Follow-through: #394 puts `privateIp.ts` under mutation testing (it had been exempt from the harness I wrote) — the first measured run scored it 57.9 with 66 surviving mutants, including NAT64 prefix comparisons that could each be loosened without any test noticing. That PR adds the adversarial cases (near-NAT64 prefixes with *public* embedded IPv4s) that kill them. #392 applies the same lesson to the citation verifier.

PR #244, opened 2026-07-23
[Security 2/9] Encrypt MCP connector secrets with per-record derived keys

amal66:
## [Security 2/9] Encrypt MCP connector secrets with per-record derived keys

### TL;DR
OAuth access/refresh tokens and client secrets for third-party connectors are stored in our database. This PR encrypts them at rest with **AES-256-GCM**, a **fresh random IV** per encryption, and a **per-record derived key** (random salt per row, key derived from a master secret via HKDF). Legacy rows still decrypt, so there is no migration flag-day.

### Risk to user data
**Severity: high.** These secrets are live credentials to a user's *other* accounts — Gmail, Slack, GitHub, whatever they connected. Stored in plaintext, any single read-path leak (a stolen backup, a mis-permissioned replica, a SQL-injection read) turns into a compromise of every connected third-party account for every user. Encryption at rest means the leaked bytes are useless without the master key, which lives outside the database.

### Flows affected
- MCP connector creation and auth-config storage (`encryptString`, `encryptJson`)
- Connector use / token refresh (`decryptString`, `decryptAuthConfig`)
- No API or behavior change for users; purely at-rest representation.

### Attack precedent
Read-path leaks of at-rest secrets are among the most common breach mechanics: exposed backups and public database snapshots, over-permissioned read replicas, and SQL injection that exfiltrates rows. The [OWASP Cryptographic Storage Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Cryptographic_Storage_Cheat_Sheet.html) exists precisely because "the DB is behind a firewall" has failed so many times.

| Option | Verdict |
|---|---|
| Store plaintext, rely on DB access controls | The status quo we're removing. One leak = full compromise. |
| Single static key for all rows | Better, but a dump can be attacked in bulk, and **equal plaintexts produce equal ciphertexts** (leaks structure). |
| AES-256-CBC / encryption without authentication | Rejected — no tamper detection; an attacker who can write rows can flip bits undetectably. |
| **AES-256-GCM + per-record HKDF-derived key + random IV** | **Chosen.** Authenticated (GCM tag detects tampering), semantically secure (random IV, per-row salt), and bulk-attack-resistant (one key's compromise ≠ all rows). |

**The composition is the whole game** — you never invent primitives, you compose standard ones correctly:
- **AES-256-GCM** — authenticated encryption; the auth tag detects tampering (encryption alone does not).
- **Random 12-byte IV per encryption** — GCM fails catastrophically on IV reuse, so this is non-negotiable.
- **Per-record key via HKDF (RFC 5869)** — random 16-byte salt per row → unique key. Bulk dump can't be attacked with one key, and equal plaintexts don't collide.

```mermaid
flowchart LR
    P["secret plaintext"] --> E
    MK["master secret<br/>(env, outside DB)"] --> HKDF
    Salt["random 16-byte salt<br/>(per record)"] --> HKDF
    HKDF["HKDF-SHA256"] --> K["per-record key"]
    K --> E["AES-256-GCM<br/>+ random 12-byte IV"]
    P --> E
    E --> C["v2. || base64(salt ‖ ciphertext)"]
    E --> Tag["auth tag"]
    C --> DB[(database)]
    Tag --> DB
```

**Migration without a flag-day:** the connector tables have no salt column, so the salt is packed into the stored value: `v2.` + `base64(salt(16) ‖ ciphertext)`. Legacy rows have no `v2.` prefix and decrypt with the old static key. A wrong/forged salt derives a wrong key, so GCM auth *fails closed* on decrypt. `crypto.test.ts` covers round-trips and tamper detection.

### What's in this PR
- `backend/src/lib/mcp/client.ts` — HKDF key derivation, packed salt format, GCM encrypt/decrypt with legacy fallback.
- Tests: `crypto.test.ts` (4 tests: round-trip, legacy decrypt, tamper detection).

### Reading
[OWASP Cryptographic Storage Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Cryptographic_Storage_Cheat_Sheet.html) · [Galois/Counter Mode](https://en.wikipedia.org/wiki/Galois/Counter_Mode) · [RFC 5869 (HKDF)](https://www.rfc-editor.org/rfc/rfc5869)

### DIFF
--- a/backend/src/lib/mcp/__tests__/crypto.test.ts
+++ b/backend/src/lib/mcp/__tests__/crypto.test.ts
@@ -1,0 +1,65 @@
+import crypto from "crypto";
+import { beforeAll, describe, expect, it } from "vitest";
+
+// The MCP secret crypto reads its master secret from the environment lazily
+// (per call), so setting it before importing the module under test is enough.
+const SECRET = "test-mcp-master-secret-at-least-32-chars-long";
+
+let encryptString: typeof import("../client").encryptString;
+let decryptString: typeof import("../client").decryptString;
+
+beforeAll(async () => {
+    process.env.MCP_CONNECTORS_ENCRYPTION_SECRET = SECRET;
+    const mod = await import("../client");
+    encryptString = mod.encryptString;
+    decryptString = mod.decryptString;
+});
+
+// Reproduce the pre-HKDF format: one static-salt scrypt key for every secret,
+// ciphertext stored as bare base64 with no version prefix.
+function legacyEncrypt(value: string) {
+    const key = crypto.scryptSync(SECRET, "mike-user-mcp-v1", 32);
+    const iv = crypto.randomBytes(12);
+    const cipher = crypto.createCipheriv("aes-256-gcm", key, iv);
+    const encrypted = Buffer.concat([cipher.update(value, "utf8"), cipher.final()]);
+    return {
+        encrypted: encrypted.toString("base64"),
+        iv: iv.toString("base64"),
+        tag: cipher.getAuthTag().toString("base64"),
+    };
+}
+
+describe("mcp secret crypto", () => {
+    it("round-trips a value through the versioned per-row scheme", () => {
+        const secret = "sk-connector-abc123";
+        const row = encryptString(secret);
+        expect(row.encrypted.startsWith("v2.")).toBe(true);
+        expect(decryptString(row.encrypted, row.iv, row.tag)).toBe(secret);
+    });
+
+    it("derives a fresh salt per encryption (no shared key across rows)", () => {
+        const a = encryptString("same-value");
+        const b = encryptString("same-value");
+        // Different salt (packed in `encrypted`) and IV → different ciphertext,
+        // yet both decrypt back to the same plaintext.
+        expect(a.encrypted).not.toBe(b.encrypted);
+        expect(decryptString(a.encrypted, a.iv, a.tag)).toBe("same-value");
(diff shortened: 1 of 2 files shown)


willchen96:
Requesting changes before merge.

The per-record HKDF design provides useful key separation, but the security improvement is modest because the existing implementation already uses AES-256-GCM with a random IV for every encryption. A database-only leak is already protected as long as the master secret remains secure.

I see two issues that should be addressed:

1. **Legacy-key compatibility:** Existing MCP records may have been encrypted using `USER_API_KEYS_ENCRYPTION_SECRET`. If a deployment introduces `MCP_CONNECTORS_ENCRYPTION_SECRET`, the new variable takes precedence and legacy ciphertext is decrypted with the wrong key. Those tokens will silently become unreadable. Please support the previous key as a fallback or provide a controlled migration, with a test covering this two-key transition.

2. **Weak-master-secret regression:** Legacy encryption uses scrypt, which makes guessing a weak environment secret expensive. The new scheme feeds the raw environment value directly into fast HKDF without enforcing that it is a high-entropy random key. Please either derive the HKDF root through scrypt or require and validate a randomly generated key of at least 32 bytes.

Please also add coverage for legacy encrypted auth-config objects, not only string tokens.

With those changes, the versioned format and per-record key separation would be sound defence-in-depth. As written, the PR introduces configuration-dependent compatibility and offline-guessing risks that outweigh its incremental benefit.

PR #246, opened 2026-07-23
[Security 19] security posture: SECURITY.md, CodeQL, secret scan, Scorecard

amal66:
This picks up the security-posture pieces that #227 (npm audit gate, security ESLint, dependabot) doesn't cover. No overlap: #227 gates *dependencies and code patterns on each PR*; this PR adds a *disclosure policy* and three *scanners*.

**What's in it**

- **SECURITY.md** — tells a researcher how to reach you privately (GitHub's built-in reporting, no email needed), what to expect (acknowledgment within 7 days — it says plainly you're a solo maintainer), and what's in scope. Since Mike is an LLM legal product, prompt-injection reports are explicitly invited.
- **CodeQL** (`codeql.yml`) — GitHub's own static analyzer, on every PR, push to main, and weekly. Both backend/ and frontend/ are interpreted TypeScript, so it needs no build step. Findings appear under the Security tab, not as PR noise.
- **Secret scan** (`gitleaks.yml` + `.gitleaks.toml`) — scans the *entire git history* for committed credentials on every PR and push. It deliberately does **not** use the official gitleaks-action, which requires a paid license key for organization repos — instead it downloads the pinned 8.30.1 binary and verifies its sha256 before running. I ran it over all 551 commits locally: the only hits were fake keys in tests and docs examples (verified by hand, each one allowlisted with a comment in `.gitleaks.toml`), so the check starts green and any *new* secret fails CI.
- **OpenSSF Scorecard** (`scorecard.yml`) — a weekly automated report card on the repo's own security hygiene (branch protection, workflow permissions, etc.), published to the public OpenSSF database.

**Maintainer action required** (repo Settings, ~2 minutes)

1. **Enable private vulnerability reporting**: Settings → Code security → "Private vulnerability reporting" → Enable. Without this, the "Report a vulnerability" button SECURITY.md points to doesn't exist.
2. **Enable secret scanning + push protection**: Settings → Code security → enable both "Secret scanning" and "Push protection". This is GitHub's server-side layer; the gitleaks workflow complements it, but push protection is the one that stops a secret *before* it lands in history.
3. **Optional, after the first Scorecard run on main**: add the badge to README —
   `[OpenSSF Scorecard](https://scorecard.dev/viewer/?uri=github.com/Open-Legal-Products/mike)`

**Notes**

- CodeQL and Scorecard results go to the Security tab (code scanning); only the gitleaks job blocks PRs directly.
- codeql-action is pinned to the v3 major; Scorecard to v2.4.3.
- Merge order vs #227 doesn't matter — the two touch disjoint files.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,41 @@
+# CodeQL: static analysis for the whole tree (backend/ + frontend/).
+#
+# Both halves are interpreted TypeScript, so CodeQL's `build-mode: none`
+# extracts them straight from source — no npm ci, no build step, and no
+# autobuild guessing at the two-package layout. One analyze job covers
+# backend and frontend together. The weekly cron re-runs analysis on an
+# otherwise-quiet tree so newly shipped CodeQL queries still get applied.
+# Findings land in the Security tab (code scanning), not as CI noise.
+name: CodeQL
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  schedule:
+    # Mondays 04:30 UTC.
+    - cron: "30 4 * * 1"
+
+permissions:
+  contents: read
+
+jobs:
+  analyze:
+    name: Analyze (javascript-typescript)
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      # Required to upload results to code scanning.
+      security-events: write
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: github/codeql-action/init@v3
+        with:
+          languages: javascript-typescript
+          build-mode: none
+
+      - uses: github/codeql-action/analyze@v3
+        with:
+          category: "/language:javascript-typescript"
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -29,13 +29,13 @@ jobs:
       # Required to upload results to code scanning.
       security-events: write
     steps:
-      - uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


willchen96:
Thanks for adding this security baseline. Private vulnerability reporting is now enabled, so that concern is resolved. I found four remaining issues before merge:

1. **Hosted-service scope:** `SECURITY.md` says there is no hosted service, but the official Mike website offers a cloud version at `app.mikeoss.com`. The third-party-deployment exclusion should apply only to independent self-hosted installations.

2. **Workflow action pinning:** The new workflows reference actions through movable tags such as `@v3`, `@v4`, and `@v2.4.3`. This is particularly concerning in the Scorecard job because it receives `id-token: write` and `security-events: write`. GitHub recommends pinning actions to verified full commit SHAs, the only immutable reference format. Please pin all new `uses:` entries and retain the release version in a comment for Dependabot.

3. **Overbroad gitleaks exclusions:** The path allowlists suppress every future secret found in those files, not just the known fake fixtures. A real credential added later to one of the security tests or documentation files would pass the scan. Please allowlist the exact fixture strings or specific historical commits instead of entire paths.

4. **Incorrect release statement:** The repository already has tags `v0.1.0` through `v0.4.0`, all predating this PR. If only `main` receives security fixes, the policy should explicitly mark those tagged versions as unsupported rather than saying no tagged releases exist.

The branch is behind current `main`, but I confirmed that it merges cleanly. I also verified the gitleaks 8.30.1 checksum and successfully scanned the merged repository with no existing leaks.

willchen96 (approved):

willchen96 (approved):

PR #249, opened 2026-07-23
[Security 6/9] Server-side citation verification against source text

amal66:
## [Security 6/9] Server-side citation verification against source text

### TL;DR
Every quoted citation the model produces is checked **server-side against the actual document text** — pure deterministic string matching, no extra LLM calls. Quotes that match exactly are `verified`; quotes that match under whitespace/case/punctuation tolerance are `repaired` (and the exact source text is swapped in); quotes that don't appear are `unverified`. A missing status is treated as untrusted by the UI.

### Risk to user data
**Severity: high — integrity, not confidentiality.** The adversary here is the model itself. LLMs fabricate plausible-looking quotes and attributions. When the app *presents a quote as coming from a document*, that is an integrity claim, and in legal work a fabricated citation is a **sanctionable harm**. This is also the third defense-in-depth layer against prompt injection ([5/9]): even if an injection makes the model assert something, a quote it attributes to a document must actually be in that document to be shown as verified.

### Flows affected
- Chat / project-chat citation finalization (`streaming.ts`).
- New `verifyCitations.ts` runs after the model's `<CITATIONS>` block is parsed.
- Air-gap preserved: source text is read from storage with `emitEvents:false`, memoized per `doc_id`, at most one read per document per turn. Case-law citations (verified upstream via CourtListener) pass through untouched.

### Attack precedent
- ***Mata v. Avianca, Inc.*** (2023) — lawyers filed a brief citing ChatGPT-invented case law and were sanctioned. ([overview](https://en.wikipedia.org/wiki/Mata_v._Avianca,_Inc.)) The canonical example of hallucinated citations causing real-world harm.

| Option | Verdict |
|---|---|
| Trust the model's quotes | The status quo we're fixing. Hallucinations ship to users. |
| Ask a second LLM "did the first hallucinate?" | **Rejected** — inherits the same failure mode (a model judging a model), costs tokens/latency, and is itself injectable. |
| **Deterministic string matching against the extracted source text** | **Chosen.** A quote either appears in the source or it doesn't — no model judgment, no LLM calls, not injectable. |

```mermaid
flowchart TD
    Q["model's quoted citation"] --> M{exact substring<br/>in source text?}
    M -- yes --> V["verified"]
    M -- no --> T{match under whitespace/<br/>case/punctuation tolerance?}
    T -- yes --> R["repaired<br/>(swap in exact source excerpt)"]
    T -- no --> U["unverified<br/>(preserve model text, flag untrusted)"]
```

**Design subtleties:** a *repaired* quote has the exact source excerpt substituted in, so drifted text is never surfaced as the source's words. Cross-page quotes (joined by a `[[PAGE_BREAK]]` sentinel) are split and each segment verified independently. Unreadable-source sentinels are treated as "no source" so a quote falls back to `unverified` rather than false-matching the error string. Custom builders (tabular) that bypass verification carry no status, and the UI treats a missing status as untrusted.

### What's in this PR
- `backend/src/lib/chat/verifyCitations.ts` — the tiered matcher (new).
- `backend/src/lib/chat/types.ts` — `CitationVerificationStatus` / `QuoteVerification`.
- `backend/src/lib/chat/streaming.ts`, `tools/documentOps.ts` — wire verification into finalization.
- Tests: `verifyCitations.test.ts` (16 tests: exact/repaired/unverified, cross-page, unreadable source).

### Reading
[Mata v. Avianca](https://en.wikipedia.org/wiki/Mata_v._Avianca,_Inc.) · [OWASP Top 10 for LLM Apps — overreliance / hallucination](https://owasp.org/www-project-top-10-for-large-language-model-applications/)
## Known limitation (from a post-open adversarial re-review)

Tier-3 tolerant matching strips punctuation without inserting a separator, so digit tokens can collapse (`1.2` → `12`): a quote citing "Section 1.2" can tier-3 match a source that only contains "Section 12". Tiers 1–2 (exact and whitespace/case-normalized) are unaffected. The tight fix — inserting a separator when punctuation sits between digits — touches the shared `normalizeWithMap` used by `find_in_document`, so I'd rather land it as a small follow-up with its own tests than widen this diff. Flagging it now so the tier-3 tolerance is reviewed with eyes open.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,4 +25,5 @@ e2e/.auth/
 
 # Local Supabase scaffold (supabase init output) — the local stack is a test
 # harness only; CI scaffolds its own fresh copy on every run
+/supabase/
 backend/supabase/
--- a/backend/src/lib/chat/prompts.ts
+++ b/backend/src/lib/chat/prompts.ts
@@ -35,6 +35,7 @@ Citation rules:
 - A cell tagged "⟨merged A1:C1⟩" spans that whole range: its value belongs to the anchor cell and the other covered cells are shown blank. When citing anything in a merged range, set "cell" to the full range from the tag (e.g. "A1:C1"), not a covered cell like "B1". Do not include the "⟨merged ...⟩" tag text in "quote".
 - For a continuous quote crossing two pages, set "page" to "N-M" and include [[PAGE_BREAK]] at the page break. Otherwise, use separate quote objects.
 - For legacy compatibility, you may also include top-level "page" and "quote" matching the first quote.
+- TESTING ONLY: Make 50% of document citation quotes exact source text and 50% deliberately false text that does not appear in the source, so citation verification produces both outcomes. Use an even number of document quotes. Do not alter case-law citations.
 - Omit the <CITATIONS> block when there are no citations.
 
 DOCX GENERATION:
--- a/backend/src/lib/chat/streaming.ts
+++ b/backend/src/lib/chat/streaming.ts
@@ -25,6 +25,7 @@ import {
   type AskInputsEvent,
   type EditAnnotation,
   devLog,
+  resolveDocLabel,
 } from "./types";
 import { TOOLS, WORKFLOW_TOOLS } from "./tools/toolSchemas";
 import {
@@ -38,9 +39,11 @@ import {
   type CourtlistenerTurnState,
 } from "./tools/toolDispatcher";
 import {
+  readDocumentContent,
   type TurnEditState,
   type TurnReadState,
 } from "./tools/documentOps";
+import { verifyDocumentCitations } from "./verifyCitations";
 
 
 export type AssistantEvent =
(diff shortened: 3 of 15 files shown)


willchen96 (approved):
Made changes

willchen96:
Implemented the citation-verification follow-up changes:

- Simplified persisted verification to booleans: `citation.verified` and `quote.verification.verified`. There is no `repaired` state or legacy string-status compatibility.
- Tolerant matches are replaced with the exact source excerpt and stored as verified; unmatched quotes remain unverified.
- Added verification support for `...`, `…`, and `[[PAGE_BREAK]]` segmented quotes.
- Added frontend treatment for unverified citations: red borderless citation markers, a liquid-glass warning badge beside the Cite button, and disabled/unselected unverified quote cards. Verified citations retain the original gray presentation with no marker.
- Added a testing-only prompt instruction that produces an even 50/50 mix of genuine and false document quotes.
- Added frontend/backend verification tests and ignored the root local `supabase/` scaffold.

Validation completed: 395 backend tests, 113 frontend tests, and both production builds passed.

Assistant response unverified citation marker: 
Screenshot 2026-08-04 at 7 20 22 PM

Assistant sidepanel citationQuoteHeader: 
Screenshot 2026-08-04 at 7 17 16 PM

PR #250, opened 2026-07-23
[Security 7/9] Download-token expiry (with legacy back-compat)

amal66:
## [Security 7/9] Download-token expiry (with legacy back-compat)

### TL;DR
Download links are HMAC-signed tokens encoding an R2 path + filename. This PR adds a **mandatory `exp` claim to every newly issued token** (30-day default TTL) and rejects expired or malformed-`exp` tokens. Well-signed *legacy* tokens without an `exp` claim are accepted **transitionally** so historical links baked into chat history don't 404 on deploy.

### Risk to user data
**Severity: medium–high.** A non-expiring download token is a **permanent capability**. In a legal product it points at a confidential document, and it leaks the way URLs always leak — chat logs, browser history, forwarded emails, referrer headers. A leaked eternal link is a confidential document exposed forever, to anyone who ever sees the URL. Time-boxing bounds that exposure window.

### Flows affected
- Download link issuance (`signDownload` / `signDownloadPayload`).
- `/download/:token` verification (`verifyDownload` / `verifyDownloadPayload`).
- Links persisted in historical chat messages (the back-compat concern).

### Attack precedent
The general principle — **credentials must be time-boxed** — is codified for signed tokens in [RFC 8725 (JWT BCP)](https://www.rfc-editor.org/rfc/rfc8725): always set and validate `exp`. Long-lived bearer URLs are a recurring source of data exposure (leaked pre-signed S3 URLs, permalink leaks via referrer).

| Option | Verdict |
|---|---|
| Keep non-expiring tokens | The status quo. Eternal capability. Rejected. |
| Add `exp` and **reject** any token without one | Correct end-state, but upstream's *entire issuance history* signed `{p, f}` with no `exp`, and those permalinks live in persisted chat messages — this would **404 every historical link the instant it deploys** (a silent flag-day). |
| Switch to short-lived pre-signed R2 URLs | Larger change; loses the "safe to store in chat history" property this token scheme was built for. Out of scope here. |
| **New tokens carry `exp` and are enforced; legacy no-`exp` tokens accepted transitionally (still HMAC-verified)** | **Chosen.** Bounds all *new* exposure immediately while keeping existing links working. Legacy tokens are still signature-verified, so only the server could have minted them. |

```mermaid
flowchart TD
    T["/download/:token"] --> S{HMAC signature valid?}
    S -- no --> Rej["reject (404)"]
    S -- yes --> E{has exp claim?}
    E -- "yes" --> V{exp well-formed<br/>and in the future?}
    V -- yes --> OK["serve file"]
    V -- "no / expired / malformed" --> Rej
    E -- "no (legacy)" --> OK2["serve file<br/>(transitional back-compat)"]
```

**Disclosed trade-off:** accepting legacy `exp`-less tokens is a deliberate, temporary compromise, called out in-code. The intended end state is strict `exp` enforcement once historical links have aged out or been re-issued on access — a one-line tightening. `downloadTokens.expiry.test.ts` locks in all four behaviours: fresh accepted (exp present), legacy accepted, expired rejected, tampered/malformed-`exp` rejected (including stripping `exp` as a tamper route). This PR reflects the reviewed final policy (the strict-then-relaxed history from #227 is collapsed into the transitional policy plus its tests).

### What's in this PR
- `backend/src/core/downloadTokens.ts` — new signer/verifier with `exp`, timing-safe compare, legacy branch.
- `backend/src/lib/downloadTokens.ts` — delegates to core; doc header reflects the policy.
- Tests: `downloadTokens.expiry.test.ts` (7 tests).

### Reading
[RFC 8725 (JWT Best Current Practices)](https://www.rfc-editor.org/rfc/rfc8725) · [OWASP Access Control Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Access_Control_Cheat_Sheet.html)

### DIFF
--- a/backend/src/core/downloadTokens.ts
+++ b/backend/src/core/downloadTokens.ts
@@ -80,15 +80,25 @@ export function verifyDownloadPayload(
       return null;
     }
     if (!parsed.p || !parsed.f) return null;
-    // Every token must carry an expiry. A token without `e` is legacy (issued
-    // before expiry existed) and would otherwise be valid forever, so reject it
-    // — any such link is long stale (all issuers have set `e` since), and a
-    // fresh, expiring token is re-issued on next access.
-    if (
-      typeof parsed.e !== "number" ||
-      parsed.e < Math.floor(Date.now() / 1000)
-    ) {
-      return null;
+    // Expiry policy:
+    // - All NEW tokens are signed with an `e` (exp) claim; expired tokens are
+    //   rejected, and a malformed `e` is rejected as tampering.
+    // - Tokens WITHOUT an `e` claim are legacy: upstream signed `{p, f}` with
+    //   no expiry until this change, and those permalinks are baked into
+    //   persisted chat messages. Rejecting them would 404 every historical
+    //   download link on deploy (a silent flag-day), so we accept them as a
+    //   transitional back-compat measure. They are still HMAC-verified —
+    //   only the server can have minted them. Once historical links have
+    //   aged out (or been re-issued on access), this branch can be tightened
+    //   to reject missing `e` as well.
+    if ("e" in parsed && parsed.e !== undefined) {
+      if (
+        typeof parsed.e !== "number" ||
+        !Number.isFinite(parsed.e) ||
+        parsed.e < Math.floor(Date.now() / 1000)
+      ) {
+        return null;
+      }
     }
     return { path: parsed.p, filename: parsed.f };
   } catch {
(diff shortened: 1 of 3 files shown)


willchen96:
Rejecting this PR. The download endpoint already requires authentication and rechecks document authorization on every request, so a leaked link alone cannot grant access. Adding expiry does not fix a demonstrated vulnerability; it only causes download links persisted in chat history to stop working after 30 days.

PR #252, opened 2026-07-23
[Security 9/9] Backend ESLint security rules, tiered to stay credible

amal66:
## [Security 9/9] Backend ESLint security rules, tiered so the gate stays credible

### TL;DR
Add [`eslint-plugin-security`](https://github.com/eslint-community/eslint-plugin-security) in a flat config over `backend/src`, wired to a CI job. Rules that only fire on genuinely unsafe APIs (`eval` with an expression, `new Buffer`, `pseudoRandomBytes`, buffer `noassert`) are **errors** that block CI; heuristic detectors (object injection, non-literal `fs`/regex, timing attacks) are **warnings** that inform review. Currently **0 errors / 131 warnings** on `main`.

### Risk to user data
**Severity: preventive.** Static analysis doesn't fix a live breach; it stops whole *classes* of the bugs the other eight PRs fix by hand from being reintroduced. `pseudoRandomBytes` for a token, `eval` on user input, an unauthenticated `new Buffer(n)` — these are the raw material of the SSRF, crypto, and injection issues addressed elsewhere in this split. A gate catches them at PR time instead of in production.

### Flows affected
- CI only (`.github/workflows/security-lint.yml`, `npm run lint`).
- No runtime code change.

### Attack precedent — on the *process*
The failure mode here isn't an exploit, it's **alert fatigue**: a linter that cries wolf gets ignored, then disabled, then catches nothing. The [security detectors are heuristic](https://github.com/eslint-community/eslint-plugin-security#rules) and false-positive on legitimate code (e.g. `detect-object-injection` fires on any `obj[variable]`). A gate that's 90% noise trains reviewers to rubber-stamp.

| Option | Verdict |
|---|---|
| All security rules as **errors** | 131 mostly-false-positive failures → the gate gets disabled within a week. Rejected. |
| All as **warnings** | Never blocks anything, so a real `eval`-on-expression sails through. Rejected. |
| Turn on only the high-precision rules | Better, but loses the heuristic detectors' review value entirely. |
| **Tier by precision: unsafe-API rules = error, heuristics = warning** | **Chosen.** The blocking tier is trustworthy (near-zero false positives), so a red build *means something*; the warning tier still surfaces suspects for human review without gating. |

```mermaid
flowchart TD
    R["eslint-plugin-security rule fires"] --> P{fires only on a<br/>genuinely unsafe API?}
    P -- "yes<br/>(eval-expr, new Buffer,<br/>pseudoRandomBytes, noassert)" --> E["ERROR → fails CI"]
    P -- "no<br/>(heuristic: object injection,<br/>non-literal fs/regex, timing)" --> W["warning → informs review,<br/>never blocks"]
```

The principle generalizes past linters: **a security gate's credibility is a resource you can spend**, and a noisy blocking gate spends it to zero. Precision-tier before you make anything blocking.

### What's in this PR
- `backend/eslint.config.mjs` — flat config, tiered rules (new).
- `backend/package.json` + lockfile — eslint toolchain (lockfile regenerated on current `main`).
- `.github/workflows/security-lint.yml` — CI job running `npm run lint`.

> Kept as its own workflow file so it composes with the existing `ci.yml` rather than conflicting. The config is the amal66 fork's `apps/api` config minus its fork-specific `no-console` rule (this backend logs via `console`).

### Reading
[eslint-plugin-security](https://github.com/eslint-community/eslint-plugin-security) · [OWASP: static analysis (SAST)](https://owasp.org/www-community/Source_Code_Analysis_Tools)

### DIFF
--- a/.github/workflows/security-lint.yml
+++ b/.github/workflows/security-lint.yml
@@ -1,0 +1,28 @@
+name: security-lint
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+jobs:
+  # Runs the backend ESLint config, whose error tier is limited to rules that
+  # only fire on genuinely unsafe APIs (eval with an expression, new Buffer,
+  # pseudoRandomBytes, buffer noassert). Heuristic detectors are warnings and
+  # never fail this job — see backend/eslint.config.mjs.
+  backend-security-lint:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: backend
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: npm
+          cache-dependency-path: backend/package-lock.json
+      - name: Install dependencies
+        run: npm ci
+      - name: ESLint (security rules)
+        run: npm run lint
--- a/backend/bun.lock
+++ b/backend/bun.lock
@@ -1,6 +1,5 @@
 {
   "lockfileVersion": 1,
-  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "mike-backend",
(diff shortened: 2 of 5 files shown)


willchen96 (changes requested):
The branch is 46 commits behind main. Please:

- Rebase onto current main.
- Regenerate backend/package-lock.json and backend/bun.lock.
- Run the linter against the current backend.
- Obtain green CI, including the frontend check and production build.

PR #210, opened 2026-07-11
MIKE-04: deploy isolated AWS staging environment

EduardoCarone:
### AWS ALB URL
```
http://atlas-mike-staging-alb-1821218227.us-east-1.elb.amazonaws.com
```

### Functional Credentials
- **Email:** admin@atlasgov.com
- **Password:** AtlasStaging2026!

### Status
- Infrastructure provisioned: **partial**
- Application operational: **yes**
- Login operational: **yes**
- Sprint status: **in progress**

### What's Working
- Backend health endpoint
- Supabase Auth (GoTrue) — signup, login, health
- Kong API gateway routing
- PostgREST running
- User created and login tested with JWT
- ALB native hostname over HTTP (no external domain)
- Zero references to external domains in code

### Commits
1. feat(infra): isolated AWS staging infrastructure
2. feat(storage): S3 through ECS IAM roles
3. feat(containers): hardened production images
4. feat(supabase): self-hosted auth and API services
5. feat(deploy): OIDC staging deployment
6. test(config): AWS IAM storage mode tests
7. docs(sprint): checkpoint documentation

### Known Limitations
- Frontend uses dev mode (production build requires CI runner)
- HTTP only (ACM cannot issue certs for *.elb.amazonaws.com)
- No Route53 custom domain (using native ALB hostname)

PR #254, opened 2026-07-23
[Testing 21] test depth: mutation testing + SSE load harness (on-demand)

amal66:
Two tools that go a level deeper than the regular test suite. **Neither one gates merges** — both are run-on-demand, so they cost you nothing day to day.

**1. Mutation testing for the security-critical code.** Line coverage says a test *ran* a line; it doesn't say a test would *notice* if that line's behavior changed. Mutation testing checks exactly that: Stryker plants hundreds of small deliberate bugs (flip a comparison, delete an early return, weaken a regex) and re-runs the suite for each one. Any bug the suite doesn't catch is listed as "survived" — a concrete missing assertion. It's scoped to the four places where a hollow test would be dangerous: sharing/access checks (`access.ts`), signed download tokens (`downloadTokens.ts`), secret redaction (`safeError.ts`), and citation extraction (`chat/citations.ts`). Current score: **~74–76% of planted bugs caught**; the run fails only if that drops below 69, so it flags real regressions, not noise. Run it with `cd backend && npm run test:mutation` (~3 min) or from the Actions tab; a monthly scheduled run catches slow drift. The HTML report shows each surviving mutant inline in the code — the surviving `safeError` redaction-regex mutants are ready-made test cases for whoever wants a small follow-up PR.

**2. A load test for chat streaming.** The SSE stream behind `POST /chat` is the hot path and has caused incidents before (streams timing out during long tool calls). `loadtest/sse-stream.js` (k6) ramps up to N simultaneous chat streams and verifies each one starts quickly and actually runs to completion (`[DONE]` arrives) — with deliberately lenient, documented thresholds: a red run means "streams are hanging," not "we missed an SLO we never set." Run it locally per `docs/test-depth.md`, or via the manual **SSE load test** workflow, which boots nothing itself — you point it at an already-running staging stack (the one from #210 is the intended target) and store a test user's token in the `LOADTEST_AUTH_TOKEN` secret. Never point it at production: it creates real chats and burns real tokens.

**Why not merge gates?** Mutation testing multiplies test runtime; the load test needs a live stack with real keys. Both are too slow/stateful to put in front of every PR, and a flaky required check is worse than none. If the project gains contributors and a permanent staging stack, the natural upgrade path is in `docs/test-depth.md`.

Verified: two full local Stryker runs completed green (2m51s / 2m25s); k6 script validated via `k6 inspect` + a dry run in docker; both workflows YAML-parse. Lockfile changes are npm-generated (`npm install`, never hand-edited).

### DIFF
--- a/.github/workflows/loadtest.yml
+++ b/.github/workflows/loadtest.yml
@@ -1,0 +1,81 @@
+# SSE streaming load test: k6 against the /chat streaming endpoint.
+#
+# ON-DEMAND ONLY, and this workflow boots NOTHING itself: you point it at a
+# stack that is already running — use the staging stack from PR #210, or
+# any dev deployment you own. It exists because the SSE chat stream is the
+# product's hot path and has bitten us before (stream timeouts on long tool
+# calls); loadtest/sse-stream.js measures time-to-first-byte and whether
+# streams actually deliver events through to [DONE] under concurrency.
+#
+# NEVER point this at production: every iteration creates a real chat row
+# and burns real LLM tokens on the target. Also raise RATE_LIMIT_CHAT_MAX
+# on the target stack first (default is 30 req / 15 min / IP — the runner
+# is one IP, so the default rate limit fails the test immediately).
+#
+# Setup (once): add a repository secret LOADTEST_AUTH_TOKEN containing a
+# Supabase access token for a throwaway test user on the target stack. The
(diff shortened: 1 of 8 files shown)


willchen96:
Please rebase to main and then split the PRs into two separate ones. Thanks.

willchen96 (approved):

PR #255, opened 2026-07-23
[Testing 18] test(frontend): API-client/SSE tests + coverage ratchet

amal66:
This gives the frontend the same safety net the backend already has: unit tests over the client library plus a coverage floor that fails CI if a change drops tested behavior.

**What's tested**
- `mikeApi.ts` — the API client every page goes through: auth headers, how server errors become readable messages (including the MFA-required case), file downloads, how stored chat events become the messages you see, and the four streaming endpoints.
- The SSE stream reader in `useAssistantChat` — the code that turns the backend's live stream into text on screen. The network is allowed to cut events into arbitrary pieces, so the tests feed it streams chopped mid-JSON, several events in one chunk, error events, garbage lines, and streams that end without a final newline. This is the frontend half of the SSE contract; the backend half is tested on its side.
- Upload validation (which file types are accepted) and model availability (which models light up for your API keys) — both now at 100%.

**The ratchet** — same idea as the backend one: `vitest.config.mts` now enforces coverage floors over `src/app/lib/**` set just below today's measured numbers (54/73/32/52 for statements/branches/functions/lines). They're a floor, not a target: never lower them; raise them in the same PR whenever you add tests. `docs/frontend-testing.md` explains how to run everything and lists what to test next, in priority order — assistant message rendering and tabular review state are the top two.

**CI** — the frontend test step now runs `npm run test:coverage --if-present`, so the floor check gates every PR; on a tree without the script it still no-ops. No other CI changes.

One new devDependency: `@vitest/coverage-v8` (installed via npm; lockfile npm-managed).

Verified locally: 94 tests green with the floors enforced, `npm run lint` 0 errors, `npm run build` green.


willchen96 (approved):
LGTM

PR #256, opened 2026-07-23
[Testing 16] ci: run the RLS/stack integration suite on every PR

amal66:
**What this fixes**

Mike already has integration tests that check the two security properties that matter most for a legal tool: that the database refuses to hand rows to browser clients (the "deny-all" row-level security firewall), and that login tokens resolve to the right user through the backend's auth path. But those tests only run against a real local Supabase stack, and nothing in CI ever started one — so on every PR they quietly skipped themselves. A change that broke RLS could have merged with a green checkmark.

This PR adds a `Stack tests` workflow that boots a disposable Supabase stack on the CI machine itself, loads the database schema, and runs that suite on every PR and every push to main. No accounts, no secrets, nothing leaves the runner — the stack is created and destroyed inside each CI run. Only the services the suite touches are started (db/auth/rest/kong), keeping the job fast.

**How the database is built**

The workflow mirrors `backend/scripts/test-stack.sh`, the existing local harness: an empty stack, then `backend/schema.sql` — the complete, current shape for fresh databases — and nothing else. This is the documented fresh-install path from the README, so the suite proves exactly the property that matters: **a fresh Mike database enforces the expected auth and RLS protections.**

*(Revision note: an earlier version of this PR also replayed the historical migrations on top of `schema.sql` and modified five of them to make that sequence pass. Per review, that was testing an artificial sequence no real deployment executes — the migration edits have been reverted and the bootstrap narrowed to `schema.sql` only. Upgrade-path/drift testing, if wanted, deserves its own separately designed change: a pinned baseline dump from an older release, the migrations dated after it, and a schema diff against a fresh-schema database.)*

**Maintainer action required**

Once this job has run green on a PR or two, please make it mandatory: **Settings → Branches → branch protection rule for `main`** → under "Require status checks to pass before merging" add the `Stack tests` job. Until you do that, the job runs and reports, but a red result won't actually block a merge.

### DIFF
--- a/.github/workflows/stack-tests.yml
+++ b/.github/workflows/stack-tests.yml
@@ -1,0 +1,106 @@
+# CI: Supabase stack integration tests.
+#
+# Runs the gated stack-level suite (backend/src/__tests__/integration/
+# *.supabase.test.ts) against a REAL local Supabase stack booted on the
+# runner — GoTrue auth plus Postgres with RLS. On every PR this proves two
+# things mocks cannot: the deny-all RLS firewall actually holds for anon and
+# authenticated clients, and the auth↔API contract (sign-in, JWT verification,
+# service-role access) still works against the real services. The database
+# bootstrap mirrors backend/scripts/test-stack.sh, which is the source of
+# truth for how the stack is prepared locally; the suite itself self-skips
+# without SUPABASE_TEST_* env vars, which is why the plain CI workflow never
+# exercised it — this workflow is the trigger that finally does.
+#
+# Everything runs locally on the runner: no secrets, no external services.
+# The keys used are the Supabase CLI's well-known local development keys.
+name: Stack tests
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main, upstream-main]
+  workflow_dispatch:
+
+# Booting a Supabase stack is the slowest job in the repo, so cancel runs
+# that a newer push to the same PR has superseded. Pushes to main each get
+# their own group (keyed by sha) and are never cancelled.
+concurrency:
+  group: stack-tests-${{ github.event_name == 'push' && github.sha || github.ref }}
+  cancel-in-progress: ${{ github.event_name != 'push' }}
+
+jobs:
+  stack:
+    name: Supabase stack integration tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 25
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: npm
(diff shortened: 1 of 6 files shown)


willchen96 (changes requested):
# Review: Request changes

The Supabase stack and real RLS integration tests are valuable. They address an important gap: tests that require Supabase were previously able to skip when no local stack was available.

However, the database initialization process in this PR is incorrect, and the resulting changes to historical migrations should not be merged.

## Main issue: migrations are being applied to the wrong starting point

`backend/schema.sql` represents the current database structure for a fresh installation.

The dated migration files represent incremental changes for databases created from older versions of Mike. They predate parts of the current schema and are intended to move an older database forward—not to be replayed after the latest schema has already been loaded.

The workflow currently tests this sequence:

```text
Empty database
→ current schema.sql
→ every historical migration
→ integration tests
```

That is not a real installation path.

A fresh installation should use:

```text
Empty database
→ current schema.sql
→ integration tests
```

Replaying old migrations after creating the current schema can cause them to encounter tables, columns, and functions that have already been changed by later work. Those failures do not prove that the migrations are defective; they show that they are being executed against the wrong starting state.

## Historical migrations should not be changed for this test

This PR modifies five old migration files so that they can run on top of the current schema.

Those changes should be reverted because:

- Existing deployments may already have executed those migrations.
- Fresh deployments should use the current `schema.sql`.
- Rewriting old migrations does not help databases that have already applied them.
- The changes make an artificial test sequence pass rather than testing the documented installation process.
- Historical migrations should remain a record of the database change that was required at that point in time.

## Scope of this PR

We do not need to add upgrade-path testing here.

The purpose of this PR should be narrower:

> Start a real, disposable Supabase stack and prove that a fresh Mike database enforces the expected authentication and RLS protections.

That can be achieved by loading the current schema and immediately running the integration tests.

If upgrade testing is ever required, it should be designed separately using a real database snapshot from a specific older release and only the migrations
```

amal66:
Thanks @willchen96 , You're right I shouldn't change historical migrations. My changes are safe to re-run on a blank state but wouldn't work well on existing database snapshots. The failures were harness artifacts not defects. 

Modified so 
* Historical migrations are restored to their former state 
* Added one migration to the end 
* PR description is updated 

The upgrade path should be a separate PR. I drafted it in #293, it builds 2 databases and checks they end up identical. That PR found 2 bugs 
1. that `workflow_open_source_submissions` didn't exist in schema.sql, causing errors when someone submits a workflow or exports their data, which is fixed. 
2. The service_role grants in schema.sql from commit b2dbb39 wasn't in a migration so older deployments don't have it. Added a migration to bring them up to date.

willchen96 (approved):

PR #260, opened 2026-07-26
[Extensibility 2/4] feat: keyless demo mode — a real first-run answer instead of a 401 (stacked on #259)

amal66:
> **Stacked on #259** — merge that first; until it merges, this diff shows its commit too. This PR's own change is the single `keyless demo mode` commit.

## Summary

A brand-new instance answers its first question with a canned, clearly-labelled, context-aware reply — plus a nudge toward Settings → API Keys — instead of a raw provider 401. First-run users and zero-setup demos work with **no API keys at all**, entirely offline.

## Changes

Backend: `providers/demo.ts` (new, registered via the registry), `DEMO_MODEL = "mike-demo"`, and `resolveDemoFallback()` in `routes/chat.ts` — a request for a keyless provider runs on the demo model instead of failing.

Frontend: "Demo (no key needed)" entry in the model picker; `ChatInput` falls back to demo instead of the blocking key popup; dismissible `ApiKeyBanner` shown while no provider key is configured.

## Why

Today a fresh install hits a hard error on the very first message. This is the single highest-friction moment for new users and demos; after this PR the first message always answers, labelled as demo output, and documents never leave the workspace until a real key is added.

## Testing

On this branch: backend `npm test` **279 passed**; frontend `npm test` **38 passed**, `eslint` 0 errors, `next build` clean.

## Provenance

Mechanical port of the demo mode running in amal66/mike (main), including its keyless-fallback fixes. Full provenance in [amal66#28](https://github.com/amal66/mike/pull/28).

### DIFF
--- a/backend/src/core/apiKeyProviders.ts
+++ b/backend/src/core/apiKeyProviders.ts
@@ -1,0 +1,66 @@
+export type ApiKeyProvider = string;
+export type ApiKeySource = "user" | "env" | null;
+
+type ProviderRecord = {
+    readonly envVars: readonly string[];
+};
+
+// Table-driven: env-var names live here, not in a switch statement.
+// Adding a new provider is one registerApiKeyProvider() call — no edits here.
+const _providerRegistry = new Map<string, ProviderRecord>([
+    ["claude", { envVars: ["ANTHROPIC_API_KEY", "CLAUDE_API_KEY"] }],
+    ["gemini", { envVars: ["GEMINI_API_KEY"] }],
+    ["openai", { envVars: ["OPENAI_API_KEY"] }],
+    ["openrouter", { envVars: ["OPENROUTER_API_KEY"] }],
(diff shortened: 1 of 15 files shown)


willchen96:
The assistant now:

- Shows that the selected model is missing an API key.
- Prevents the message from being submitted.
- Displays an **“API key required”** warning.
- Explains how to configure a key and links to account settings.

This avoids the raw provider `401` without pretending that real AI analysis occurred.

This PR now adds a separate product behavior: a built-in `mike-demo` model that returns a canned placeholder and saves it as an assistant response. I do not think we need that behavior.

It introduces several disadvantages:

- Users may mistake the placeholder for genuine AI analysis.
- Missing configuration is concealed by a successful response.
- API clients could unexpectedly receive a canned answer instead of an error.
- The behavior is inconsistent with features such as tabular reviews, which still require a real provider key.
- It depends on the larger provider-registry work in #259.
- It adds substantial complexity for a first-run problem that the current warning already handles clearly.

PR #261, opened 2026-07-26
[Extensibility 3/4] feat: local LLMs via Ollama, behind ENABLE_OLLAMA (stacked on #259)

### LINKED ISSUE #23: Allow for self-hosted inference engines.
Right now the front-end only allows for Google or Anthropic models, and requires API keys by default. Hope you can include on the roadmap more model provider / inference engine options. I am running my own vLLM inference engine.

amal66:
> **Stacked on #259** — merge that first; until it merges, this diff shows its commit too. This PR's own change is the single Ollama commit.

Closes #23.

## Summary

Opt-in local LLM support via Ollama's OpenAI-compatible endpoint: a self-hosted deployment can run chat inference entirely on its own hardware — documents never leave the building, and with this flag neither does inference. Everything is behind `ENABLE_OLLAMA=true` (default **off**): with the flag unset, no new models appear and nothing changes.

## Changes

- `providers/ollama.ts` (new): registry entry that routes local model IDs (llama3.3, phi4, qwen2.5, + `OLLAMA_MODELS` extras) through the existing OpenAI adapter pointed at `http://localhost:11434/v1`. **No new dependencies, no new streaming code path.**
- `baseUrl.ts` (new) + `openai.ts`: `OPENAI_BASE_URL` resolution with an SSRF guard — http/localhost/private-IP endpoints rejected in production unless explicitly allowed. Defaults to `https://api.openai.com/v1` (unchanged when unset).
- `privateIp.ts` (new): shared private/reserved IP classifier (IPv4 + IPv6 incl. mapped/NAT64/6to4).
- `.env.example`: documents the opt-in config.
- Tests: ollama gate, baseUrl, privateIp (21 tests).

## Why

The most-requested self-hosting capability (#23), done as a registry entry instead of a bespoke client — the first external provider to use the registry, which is exactly what it was built for.

## Testing

On this branch: backend `npm test` — **300 passed, 5 skipped** (includes the 21 new tests). Flag gate verified: only the exact string `"true"` registers the provider.

## Provenance

Mechanical port of code running in amal66/mike (main); `privateIp.ts` and `providers/ollama.ts` byte-identical to the fork. Full provenance in [amal66#34](https://github.com/amal66/mike/pull/34).

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -16,6 +16,18 @@ R2_BUCKET_NAME=mike
 GEMINI_API_KEY=your-gemini-key
 ANTHROPIC_API_KEY=your-anthropic-key
 OPENAI_API_KEY=your-openai-key
+# Optional: OpenAI-compatible gateway endpoint, e.g. OpenRouter/Azure proxy/Ollama-compatible bridge.
+# Must be https in production unless OPENAI_ALLOW_LOCAL_BASE_URL=true.
+OPENAI_BASE_URL=https://api.openai.com/v1
+OPENAI_ALLOW_LOCAL_BASE_URL=false
+# Local models via Ollama (opt-in, off by default). To enable: set ENABLE_OLLAMA=true
+# and point the OpenAI-compatible client at your local Ollama server:
+#   ENABLE_OLLAMA=true
+#   OPENAI_BASE_URL=http://localhost:11434/v1
+#   OPENAI_ALLOW_LOCAL_BASE_URL=true
+#   OPENAI_API_KEY=ollama          # Ollama accepts any non-empty string
+#   OLLAMA_MODELS=my-custom-model  # optional; extra models beyond the defaults
+ENABLE_OLLAMA=false
 RESEND_API_KEY=your-resend-key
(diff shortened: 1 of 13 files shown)


amal66:
Keeping this in draft — an adversarial re-review found three confirmed bugs to fix first: (1) the https-in-production check fires before `OPENAI_ALLOW_LOCAL_BASE_URL` is consulted, so the documented production config throws on every request; (2) the reused OpenAI tool loop depends on `previous_response_id`, which Ollama's stateless endpoint silently drops — tool-using chats lose all context; (3) the model picker is a static list, so enabled Ollama models aren't selectable from the UI. Please skip this one until it's reworked — #259/#260/#262 are the reviewable part of the series.

willchen96:
Closing as superseded by the Ollama implementation already on `main`.

The current implementation includes dynamic model discovery, frontend model selection, tool-calling with conversation context preserved, configurable connection settings, optional authentication, and documentation.

This draft is also conflicting and its known issues are addressed by the implementation now on `main`, so there is nothing remaining here that needs to be merged.

PR #263, opened 2026-07-26
Db pagination for tabular reviews

amay0048:
## Summary

For the Tabular Reviews page, this moves the keyword search and order by functions from the data table into the backend, and adds pagination with lazy load fetching.

## Why / Motivation

First part of the pagination feature request from the discord chat.

## Changes

Implement offset based paging in the database for the tabular view.

## Tradeoffs & risks

Considered multiple UI paging strategies and interface options, against the potential perf impacts and scope for the number of records considering this is an open source project and may be used in multi-tenanted environments.

This is an implementation on a single view, and once the approach is reviewed and approved, should be mirrored in the other views which require paging.

## How verified

Added some tests, but mainly by manual testing in my local instance manually populated with seed data to test pagination.

## Checklist

- [x] Updated docs / env examples if setup, config, or behavior changed.


willchen96:
Added a follow-up reliability pass in `6b0bd1d` based on review feedback:

- Applied search, scope, project, and sorting filters in the database before pagination.
- Added pagination and load-more behavior to project-specific tabular review pages.
- Added deterministic `id` ordering for tied rows.
- Aligned the search expression with the trigram index so PostgreSQL can use it.
- Kept the legacy three-argument RPC callable while making the paginated signature unambiguous to PostgREST.
- Kept `backend/schema.sql` synchronized with the migration.
- Consolidated pagination behavior into a shared, abortable frontend hook.
- Added retryable initial-load and load-more error states.
- Scoped row selection to the active query so hidden/stale rows cannot be batch-deleted.
- Removed redundant client filtering and project table props.
- Added hook tests and real Supabase integration coverage for page boundaries, tied rows, search/scope filters, sorting, and legacy RPC compatibility.

Validation completed:

- Backend unit tests: 270 passed
- Frontend tests: 41 passed
- Local Supabase integration tests: 9 passed
- Backend and frontend production builds passed
- Lint completed with no errors (existing unrelated warnings remain)

willchen96 (approved):

willchen96:
Addressed the remaining review feedback in d6b35c9:

- Fixed the select-all debounce race by including both the immediate and debounced search values in the selection identity, and blocking select-all while they differ.
- Replaced the unbounded bulk-delete Promise.all with bounded concurrency (5 requests at a time). Only confirmed deletions are removed locally; failed rows remain selected and surface a warning.
- Added per-row deletion feedback: the checkbox becomes a spinner and the row is greyed out and disabled while deletion is in progress.
- Increased the initial tabular-review page size from 20 to 30.
- Escaped %, _, and backslashes in both review-search RPCs so searches are literal.
- Fixed project searches with no matches to show “No reviews found” instead of the general creation empty state.
- Added regression coverage for pagination selection, bounded deletion, literal wildcard searches, and the debounce transition.

Verification completed:
- 47 frontend tests passed
- 14 live Supabase integration tests passed
- ESLint passed
- Production build and TypeScript checks passed

PR #271, opened 2026-07-28
feature: add workspace history of actions (History page + audit_events)

jmooves:
## User Experience
Currently no audit history exists within Mike. This feature enhancement adds an audit trail of user actions with a top-level **History** page and adds a sidebar entry below Workflows.

## What it does
- **History page** (`/history`): User · Created · Title (linked to the chat/review) · Status · Type · Application · Model, with search, a type filter, a date range, CSV export and pagination.
- **Recorded actions**: chat turns (model used; `cancelled` status when a stream is aborted) with per-artifact rows mined from the persisted stream events (generated documents, edits, applied workflows), document uploads, tabular review creation and generation runs, and data exports.
- **Visibility**: each user sees their own actions plus activity in projects they own or that are shared with them and reuses the existing `shared_with` model, no new permission concept.

## Design notes
- `audit_events` is **append-only**, added to `schema.sql` (fresh installs) and as a dated migration (existing deployments). Accessed only via the service-role backend.
- The recorder (`lib/audit.ts`) is **fire-and-forget**: it can never throw or block a user-facing path - failures log and are swallowed.
- Artifact rows come from the already-persisted assistant events, so the tool layer needed no changes.

## Testing
- `npx vitest run` green on this branch; `tsc` clean on both apps; `next build` includes `/history`.
- Validated end-to-end on a self-hosted deployment: a chat turn, review creation and a chats export all appeared in `/audit` with correct fields; the type filter and CSV export verified.

🤖 Generated with [Claude Code](https://claude.com/claude-code) and human reviewed. :)

left_pane
new_history_pane


amal66:
Thanks for taking this on — an audit trail is a real gap in Mike, and you got the hard parts right: the fire-and-forget recorder genuinely can't reject (which matters, since a floating rejection would take down the process), and visibility failing closed on a project-lookup error is exactly the right default.

A few things I'd consider blocking, all with small fixes:

- `audit_events` has no revoke or RLS, so hosted PostgREST exposes it to browser roles for read *and* write — every other backend-only table revokes `anon, authenticated`. Needs the same in both `schema.sql` and the migration, plus `enable row level security`.
- Account deletion never purges `audit_events` (`userDataCleanup.ts`) — a right-to-erasure problem, and the account export omits them too.
- `csvCell` in `routes/audit.ts` doesn't neutralize leading `=`/`+`/`-`/`@`, and chat titles are attacker-controlled cross-user — classic CSV formula injection.
- In `schema.sql` the blanket grants run before the `audit_events` DDL, so fresh installs get permission-denied inserts (silently, since the recorder swallows errors).

Also worth picking up: `/audit/export` skips the rate limiter and MFA gate the other exports have; project-document uploads aren't actually audited and `doc_replicated` events are mined from the wrong event shape; `page`/`from`/`to` are unvalidated and can 500; and the History page has no error state and races on rapid filter changes.

I've opened https://github.com/jmooves/mike/pull/1 against your branch with all of this fixed, plus tests — merge it and this PR updates, or cherry-pick what you agree with. It sits directly on your branch: five focused commits, nothing unrelated, with rebasing onto current `main` left to you. Both typechecks are clean and both suites pass on the result (backend 273, frontend 38).

jmooves:
Thank you for the review — and for the fix branch, which is a model of how to hand feedback back. Merged jmooves/mike#1 wholesale (all five commits verified: the revoke/RLS + grant-ordering fix, deletion purge + export inclusion, CSV formula-neutralisation, export limiter/MFA parity, input bounds, the doc_replicated event shape, project-upload auditing, and the History page error/race fixes) and brought the branch up to current main (one trivial import conflict in app.ts). Both suites green on the result (backend 496, frontend passing; both typechecks clean).

amal66:
Glad the fix branch was useful — and thanks for doing the main sync cleanly.

Two checks are still red on the current head, and both are my fault rather than yours, introduced by that same fix round:

- **CodeQL (high)**: the ILIKE escaper I added in the `/audit` hardening commit escapes `%`/`_` but not backslash itself, which is the pattern's own escape character. One-line fix plus tests.
- **Frontend coverage ratchet**: all 179 tests pass — the job fails because `getAuditHistory`/`exportAuditHistory` live in the one floor-gated file (`mikeApi.ts`) with no tests, which drops global statements to 74.9% against the 79% floor. Fixed with tests, not by lowering the floor.

Same drill as before: https://github.com/jmooves/mike/pull/2 sits directly on your branch tip with exactly these two commits, verified green locally (backend 501, frontend 179 with the ratchet passing, both typechecks clean). Merge it and this PR should go fully green — from my side that's the last blocker; CLA is signed, the branch contains current main, and e2e/backend/evals already pass.

willchen96:
Implemented the History page refinements:

- Refactored the page to use the shared PageHeader, TableToolbar, and table primitives.
- Moved search and export into the page header and added server-side status, application, and sorting controls.
- Added a rolling 30-day date filter with separate Start and End date pickers.
- Kept history paginated at 50 events per request with incremental loading.
- Added batched profile resolution so history shows separate Username and Email columns, with email retained when a display name is unavailable.
- Replaced status pills with capitalized, color-coded text and liquid-glass dots.
- Moved History into the account menu and refreshed the clock icon and empty state.
- Updated the audit endpoint, CSV export behavior, API client, and test coverage.
- Merged the latest main changes and resolved the sidebar/API test conflicts while preserving the new pagination behavior.

Validation completed successfully: backend TypeScript build, 15 audit tests, frontend lint and TypeScript checks, and 151 focused frontend tests.

willchen96 (approved):

willchen96:
@jmooves thank you for this excellent contribution. one note - please use the existing primitive components for new pages - i.e. PageHeader, TableToolbar, and table primitives etc.

willchen96:
@jmooves are you already in the discord?

PR #273, opened 2026-07-29
Remove dead jsonLimitForPath abstraction

b1rdmania:
`jsonLimitForPath()` ignores its `path` argument and always returns `"50mb"`, and the per-request wrapper re-invokes `express.json()` on every request for no effect. This replaces both with a single static `express.json({ limit: "50mb" })` mount.

- Behaviour identical, one insertion, seven deletions
- `npm run build` passes; full vitest suite passes (259 passed, 5 skipped)
- If per-path body limits are wanted later (which the function's shape suggests was the original intent), happy to implement that properly against an explicit route list

Found while reading the codebase for our fork work (b1rdmania).

### DIFF
--- a/backend/src/app.ts
+++ b/backend/src/app.ts
@@ -17,6 +17,10 @@ import { caseLawRouter } from "./routes/caseLaw";
 export const app = express();
 const isProduction = process.env.NODE_ENV === "production";
 
+// Ceiling for JSON request bodies. Generous because chat and tabular
+// routes post document text inline; uploads go through multer, not here.
+const JSON_BODY_LIMIT = "50mb";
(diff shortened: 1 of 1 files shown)


amal66 (changes requested):
Having descriptive variable names is better than unnamed strings. Could we please revise this to have a descriptive constant named in all caps at the top of the file, and reusing that? 

Some rationale from a related article, under "Why Use constants": https://realpython.com/python-constants/

willchen96 (approved):

willchen96:
Merging in as change request has been addressed. Made commit to clarify comment.

PR #54, opened 2026-05-10
feat: folder-grouped tabular reviews with per-user page limit

ecarjat:
## Summary

- **Folder-grouped rows**: a new toggle in the tabular review creation flow lets users treat all documents inside the same project subfolder as a single review row (label = folder path, subtitle = N documents). Documents without a subfolder still get their own row.
- **Per-user page limit**: a configurable `tabular_max_pages` setting (10–2000, default 250) in Account → Models & API Keys controls how large a grouped row can be before it is skipped. Rows that exceed the limit now surface a single, full-width error banner instead of repeating the error icon in every column.
- **PDF.js noise fixes**: all `getDocument` calls now pass `standardFontDataUrl` and `verbosity: 0`, eliminating the `standardFontDataUrl` and `TT: undefined function: 32` console warnings.

## Database changes

Two new migrations ship alongside an updated `schema.sql`:

| Migration | Purpose |
|---|---|
| `20260510000000_tabular_review_rows.sql` | Adds `document_grouping` to `tabular_reviews`; introduces `tabular_review_rows` and `tabular_review_row_sources` tables; backfills existing cells; adds RLS policies |
| `20260510000001_user_tabular_max_pages.sql` | Adds `tabular_max_pages integer default 250` to `user_profiles` |

The initial schema (`20240101000000_initial_schema.sql`) is also updated to keep it in sync with the new tables for fresh installs.

## Test plan

- [x] Create a project with at least two subfolders, each containing multiple documents
- [x] Start a tabular review → enable "Treat documents in the same subfolder as one row" → confirm rows are grouped by folder with correct labels and document counts
- [x] Run all columns → confirm each folder row is processed as a unit
- [x] In Account → Models & API Keys, set Max pages to a low value (e.g. 10) and re-run → confirm the row shows the full-width error banner with the page-count message
- [x] Confirm no `standardFontDataUrl` or `TT: undefined function` warnings in backend logs after PDF upload/processing

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #274, opened 2026-07-29
feat: add folder-grouped tabular review rows

ecarjat:
## Summary

Focused replacement for #54, rebased on the current `main` branch.

## Changes

- Adds an optional folder grouping mode when creating a project tabular review.
- Persists review rows and their source documents, while preserving document-level rows for files outside a subfolder.
- Adds the schema migration, fresh-schema support, and backend-owned table access hardening.

## Why

Lets a tabular review process all documents in a project subfolder as one logical row without coupling the change to the separate page-limit or PDF.js cleanup work from #54.

## Testing

- `npm test --prefix frontend`
- `npm test --prefix backend`


amal66:
Thanks for splitting this out of #54. The idempotent migration with a legacy backfill, the create-path rollback, and the revoke + RLS on the two new tables — matching, even exceeding, the repo's backend-owned-table pattern — are genuinely good work.

The core problem is that the feature currently ends at the database. Must-fix:

- Folder rows are write-only. Cells are inserted with `document_id: null`, but `/generate`, `/regenerate-cell`, and the table UI all key on `document_id`, so folder cells stay pending forever and the checkbox ships a broken state.
- Frontend `tsc --noEmit` fails with three errors from widening `TabularCell.document_id` to `string | null`, which breaks `next build`.
- Rebase: `schema.sql` conflicts with main's pagination RPC rework — the resolution must keep main's new functions — and the migration is dated before ones already shipped on main, so operators would skip it; rename it to a later date.
- `PATCH /:reviewId` edits cells by `document_id` only and never maintains the new rows/sources tables, so folder reviews desync on edit.

Smaller: several helpers ignore Supabase errors (a transient failure yields a silently empty review), and adding the two new tables to the stack test's `PUBLIC_TABLES` would enforce the hardening claim in CI.

To make this easy to act on I've opened https://github.com/ecarjat/mike/pull/1 against your branch: row-based generation and rendering wired end to end, the tsc errors and migration date fixed, tests added — merge it or cherry-pick. It sits directly on your branch as three focused commits; the rebase onto main (including the `schema.sql` resolution) is left to you. On the result, both typechecks are clean and both suites pass (backend 268, frontend 41).

amal66 (commented):
Details in my comment below — the schema work is solid, but generation/rendering aren't wired to the new rows yet, the frontend typecheck fails, and the migration needs re-dating. There's a follow-up PR against your branch implementing the fixes.

willchen96:
Pushed follow-up commit 6ae1f98 with the completed folder-grouped tabular review experience.

What changed:
- Added folder-based review rows for project folders and nested library folders, while retaining individual rows for ungrouped documents.
- Moved document grouping to the first creation step and replaced the checkbox with the shared toggle primitive.
- Updated the file picker to render nested folders consistently across Files and Projects; the tabular-review picker no longer shows Templates.
- Updated the first column to Folder / Document with the appropriate folder or file-type icon. Folder rows expand to show their source documents without divider lines.
- Made grouped results and the side panel expandable so users can inspect and open each underlying document.
- Added document IDs to citations so citations from grouped rows open the correct source document.
- Updated exports, table navigation, chat context, API types, migrations, schema, and access handling for logical review rows.
- Removed the previous 120,000-character aggregate truncation so later documents in a grouped folder are not silently excluded from extraction.
- Added focused frontend and backend coverage for folder rows, nested directories, citations, document switching, and the toggle primitive.

Validation completed: backend TypeScript build, 33 tabular route tests, focused frontend tests, frontend type-checking, ESLint, and diff checks.

willchen96 (approved):

PR #279, opened 2026-07-31
feat: content hashes on document versions + signed project export manifest

b1rdmania:
Rebase of #181 onto current main, with the manifest signing you asked for there.

## What it does

Exports cannot currently prove that a document version is the file the workspace held. This stores a SHA-256 of every version's bytes at write time and adds a per-project manifest listing them, alongside the accept/reject trail. To check a file: `shasum -a 256 lease.docx` against the manifest.

Versions written before this shipped carry a `null` hash, so an old file set reads as unverifiable rather than as falsely verified.

## Signing

The manifest carries a SHA-256 digest over its own body, meaning everything except `digest` and `signature`. Serialising sorts object keys, keeps array order, and emits no whitespace, so a recipient can recompute the digest from parsed JSON.

Set `MANIFEST_SIGNING_KEY` and that digest is signed. Optional rather than required, because Mike self-hosts and a deployment without key custody should still be able to export: it gets `signature: null` and the weaker guarantee. A *malformed* key fails the export with a clear error instead of quietly producing an unsigned manifest, which is the failure signing exists to prevent.

The signature is a raw Ed25519 signature over the bytes `mike-project-manifest-v1`, a NUL byte, then the digest bytes. Two consequences worth stating:

- The context prefix means a manifest signature cannot be replayed as a signature over some other object that hashes to the same 32 bytes, so one key can cover other object types later.
- There is no library to adopt. Any Ed25519 implementation checks it in three lines, shown in the README.

The public key is served at `GET /manifest-signing-key`, unauthenticated. Take the key from there rather than from the manifest: whoever edits a manifest can re-sign it with a key of their own, so the embedded copy shows consistency, never provenance.

Key material is a hex seed from `openssl rand -hex 32`, the same pattern as `DOWNLOAD_SIGNING_SECRET`.

## Against current main

Not a git rebase. The branch on #181 predates the `app.ts` factory, the vitest suite, and the `chatTools.ts` to `lib/chat/tools/` split, so I re-applied it by hand:

- Hashes at every byte-write site on current main, including the three that did not exist in June: `generateOfficeDoc`, the library upload path, and the project upload path. `library.ts` renames do not touch bytes, so no hash there.
- Accept/reject resolution and assistant-edit reuse rewrite bytes in place, as they do today. Storage and Postgres cannot be written atomically, so the hash is cleared before the bytes change and set again after. An interrupted resolution leaves the version unhashed and therefore unverifiable, never hashed against content it no longer holds.
- Export route registered with the existing `exportLimiter` and `requireMfaIfEnrolled`, matching the account export routes.
- Migration under `backend/migrations/` with the current dated naming, nullable column, safe to re-run.

The diff adds 892 lines and removes 5. Nothing here rewrites existing behaviour.

## Testing

Full suite 298 passed, 14 skipped. `npm run build` clean.

- `contentSha256` against the FIPS 180-4 vector across Buffer, ArrayBuffer, and offset views, since multer hands us views into a larger buffer.
- Signing: canonical serialisation, determinism, a body edited after signing, and a signature made over a different context string. The wire-format tests rebuild the payload and the key independently and verify with plain `crypto.verify`, so they pin the format rather than agreeing with the implementation.
- Export route: access gate, manifest shape, signed and unsigned paths, and no error leakage on failure.
- `GET /manifest-signing-key`: serves the public key, never the seed.

## Deliberately left out

- **Backfill for existing rows.** It would mean streaming every object out of R2, and I would rather not guess how you want that batched.
- **Trusted timestamping.** The manifest attests to bytes, not to when they existed; `exported_at` is the server clock. RFC 3161 or a transparency log is the honest next step, and both need infrastructure a self-hosted deployment may not have.

## Held back rather than bundled

This is the smallest version I could defend. Three things are already built and not included. Happy to open any of them, or none:

- A `verifyManifest()` helper returning one verdict — `verified`, `unsigned`, `key-mismatch`, `bad-signature`, `tampered` — where checking against the key inside the manifest never returns `verified`.
- A `docs/` page: worked manifest example, verification walkthrough, key rotation, and stated limits.
- Two things that predate this PR, which I noticed and did not touch. `selectAll` can duplicate or drop rows across page boundaries when `created_at` ties, because there is no tiebreaker column in the sort. And two concurrent accept/rejects on one document race on the stored bytes. Both are yours to weigh; I have no code attached to either here.

---

This came out of an audit layer in an IDE I have been building, where I was riffing on audit trails and what makes AI-assisted work insurable: [Legalise](https://github.com/b1rdmania/legalise). [`docs/TRUST.md`](https://github.com/b1rdmania/legalise/blob/main/docs/TRUST.md) is the write-up if it is useful.

### DIFF
--- a/backend/src/lib/chat/tools/toolDispatcher.ts
+++ b/backend/src/lib/chat/tools/toolDispatcher.ts
@@ -1681,7 +1681,10 @@ export async function runToolCalls(
                 version_number: 1,
                 filename: d.filename,
                 file_type: active?.file_type ?? sourceInfo.file_type,
-                size_bytes: active?.size_bytes ?? raw.byteLength,
+                // From `raw`, not `active`, so size and hash always describe
+                // the same bytes. A verifier that stats a file before hashing
+                // it must not see a size that disagrees with content_sha256.
+                size_bytes: raw.byteLength,
                 page_count: active?.page_count ?? null,
                 content_sha256: contentSha256(raw),
               }));
--- a/backend/src/lib/chat/tools/toolDispatcher.ts
+++ b/backend/src/lib/chat/tools/toolDispatcher.ts
@@ -33,7 +33,10 @@ import {
 import { convertedPdfKey } from "../../convert";
 import { contentTypeForDocumentType } from "../../documentTypes";
 import { buildDownloadUrl } from "../../downloadTokens";
-import { loadActiveVersion } from "../../documentVersions";
+import {
+  contentSha256,
+  loadActiveVersion,
+} from "../../documentVersions";
 import { type EditInput } from "../../docxTrackedChanges";
 import {
   citationReminder,
@@ -1680,6 +1683,7 @@ export async function runToolCalls(
                 file_type: active?.file_type ?? sourceInfo.file_type,
                 size_bytes: active?.size_bytes ?? raw.byteLength,
                 page_count: active?.page_count ?? null,
+                content_sha256: contentSha256(raw),
               }));
               const { data: insertedVersions, error: verErr } = await db
                 .from("document_versions")
--- a/backend/src/lib/manifestSigning.ts
+++ b/backend/src/lib/manifestSigning.ts
@@ -1,0 +1,188 @@
+import crypto from "crypto";
+
+/**
+ * Ed25519 signing for project export manifests.
+ *
+ * A manifest carries a SHA-256 per document version, which proves an exported
+ * file has not changed *if you trust the manifest*. Signing closes that gap:
+ * the manifest's own digest is signed with a server-held key, so a recipient
+ * holding the deployment's public key can tell an untouched manifest from an
+ * edited one.
+ *
+ * Optional by design. Mike self-hosts, and a deployment that has not set up a
+ * key should still be able to export — it just gets `signature: null` and the
+ * weaker guarantee. Set MANIFEST_SIGNING_KEY to turn signing on; a malformed
+ * key throws rather than silently downgrading to unsigned, because a manifest
+ * that is quietly unsigned is the failure signing exists to prevent.
+ *
+ * Key material is a raw 32-byte Ed25519 seed, hex-encoded, matching the
+ * `openssl rand -hex 32` pattern already used for DOWNLOAD_SIGNING_SECRET.
+ */
+
+// PKCS#8 DER prefix for an Ed25519 private key (RFC 8410). Node will not build
+// a key object from a bare 32-byte seed, so the seed is wrapped in this fixed
+// envelope first. Ed25519 seeds and public keys are both 32 bytes.
+const PKCS8_ED25519_PREFIX = Buffer.from(
+    "302e020100300506032b657004220420",
+    "hex",
+);
+const ED25519_KEY_BYTES = 32;
+
+/**
+ * Domain separation. The signature covers this context string and a NUL byte
+ * before the digest bytes, so a manifest signature can never be replayed as a
+ * signature over some other object that hashes to the same 32 bytes. Bump the
+ * version suffix if the signed payload's shape changes.
+ */
+const SIGNING_CONTEXT = "mike-project-manifest-v1";
+
+function signedPayload(digestHex: string): Buffer {
+    return Buffer.concat([
+        Buffer.from(`${SIGNING_CONTEXT}\0`, "utf8"),
+        Buffer.from(digestHex, "hex"),
+    ]);
+}
+
+export interface ManifestSignature {
+    algorithm: "ed25519";
+    /** First 16 hex chars of SHA-256 over the raw public key. Rotation aid. */
+    key_id: string;
+    /** Raw Ed25519 public key, hex. */
+    public_key: string;
+    /** Signature over SIGNING_CONTEXT + NUL + the raw digest bytes, hex. */
+    value: string;
+}
+
+export interface ManifestDigest {
+    algorithm: "sha256";
+    value: string;
+}
+
+export interface SigningIdentity {
+    algorithm: "ed25519";
+    key_id: string;
+    public_key: string;
+}
+
+function seedFromEnv(): Buffer | null {
+    const raw = process.env.MANIFEST_SIGNING_KEY?.trim();
+    if (!raw) return null;
+    if (!/^[0-9a-fA-F]{64}$/.test(raw)) {
+        throw new Error(
+            "MANIFEST_SIGNING_KEY must be a 32-byte hex string " +
+                "(generate with `openssl rand -hex 32`). Unset it to export " +
+                "unsigned manifests.",
+        );
+    }
+    return Buffer.from(raw, "hex");
+}
+
+/**
+ * The configured signing key, or null when signing is off. Signing and key
+ * publication both go through here so they cannot disagree about the key id.
+ */
+function signingKey(): {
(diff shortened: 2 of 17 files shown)


backend/src/lib/manifestSigning.ts
```diff
@@ -0,0 +1,188 @@
...
+ */
+export function canonicalize(value: unknown): string {
+    // NaN and Infinity both stringify to "null", which would let two different
+    // bodies share a digest. Nothing in a manifest is a float, so this guards
+    // a case that should not arise — but a digest collision is the one thing
+    // canonicalisation must not permit.
+    if (typeof value === "number" && !Number.isFinite(value)) {
+        throw new TypeError("Manifest bodies must hold finite numbers");
+    }
+    if (value === null || typeof value !== "object") return JSON.stringify(value);
```
amal66, line 142: `canonicalize` guards against non-finite numbers but silently mishandles non-plain objects. A `Date` lands in the object branch and serializes as `{}` (where `JSON.stringify` would give the ISO string), a `Buffer` enumerates its numeric indices, and a `Map` serializes as `{}`. Today's manifest body only holds JSON-safe values from the DB, so this isn't a live bug — but this function is the foundation of the digest, and a silent wrong serialization is the same failure class as the NaN case you already reject loudly.

Suggest rejecting anything whose prototype isn't `Object.prototype`/`null`, e.g.:

```ts
const proto = Object.getPrototypeOf(value);
if (proto !== Object.prototype && proto !== null) {
    throw new TypeError("Manifest bodies must hold plain objects");
}
```

(Cycles are impossible with DB rows for the same reason, so I'd leave those unguarded — the recursion will throw on its own.)

backend/src/lib/manifestSigning.ts
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * The configured signing key, or null when signing is off. Signing and key
+ * publication both go through here so they cannot disagree about the key id.
+ */
+function signingKey(): {
+    privateKey: crypto.KeyObject;
+    identity: SigningIdentity;
+} | null {
+    const seed = seedFromEnv();
```
amal66, line 88: Two related points on `signingKey()`:

1. It re-derives the keypair from the env var on every call — each export and each `/manifest-signing-key` hit does a PKCS#8 wrap, public-key derivation, and SPKI export. Ed25519 derivation is cheap so this is not a perf problem, but a module-level cache keyed on the seed string would make it free and keep the test ergonomics (tests mutate `process.env` between cases, so key the cache on the seed value, not a boolean).

2. More importantly, a malformed `MANIFEST_SIGNING_KEY` isn't discovered until the first export or key fetch fails at runtime. Failing closed there is correct, but a boot-time validation call (even just invoking `manifestPublicKey()` once at startup and logging/exiting) would surface the misconfiguration at deploy time instead of when a user first tries to export.

backend/src/lib/chat/tools/toolDispatcher.ts
```diff
@@ -1680,6 +1683,7 @@ export async function runToolCalls(
                 file_type: active?.file_type ?? sourceInfo.file_type,
                 size_bytes: active?.size_bytes ?? raw.byteLength,
                 page_count: active?.page_count ?? null,
+                content_sha256: contentSha256(raw),
```
amal66, line 1689: Nit, but worth a look in a tamper-evidence context: `content_sha256` is computed over `raw` (which is what `uploadFile` writes, so the hash itself is correct), while `size_bytes` and `file_type` on the same row are inherited from `active` with `raw`-derived fallbacks. If `active`'s metadata ever disagrees with the actual source bytes, the manifest would carry a correct hash next to an incorrect size — and a verifier who checks `stat` before `shasum` sees a false mismatch. Since you have `raw` in hand, `size_bytes: raw.byteLength` unconditionally seems strictly more accurate here.

backend/src/lib/userDataExport.ts
```diff
@@ -160,6 +161,115 @@ export async function buildUserTabularReviewsExport(
...
+            ? Promise.resolve([])
+            : selectAll(
+                  db,
+                  "document_versions",
+                  (query) =>
+                      query
+                          .in("document_id", documentIds)
+                          .order("created_at", { ascending: true })
+                          .order("id", { ascending: true }),
+                  "id, document_id, version_number, source, filename, file_type, size_bytes, content_sha256, deleted_at, created_at",
```
amal66, line 212: Confirming this is intended: the manifest includes soft-deleted versions (`deleted_at` is selected and exported, with no filter on it). For an audit trail I think including them is the right call — a trail that silently drops deleted versions is a weaker attestation — but it does mean anyone with project access sees metadata (filenames, timestamps) of versions a user believed deleted. If intended, a line in the README's Tamper-Evident Export section saying deleted versions appear in the manifest would set expectations; if not, this needs a `.is("deleted_at", null)` filter.

amal66 (commented):
Love the idea of tamper-evident exports! Thanks for the testing and defensive engineering, and good failure asymmetry (absent key → unsigned, malformed key → hard error)

Things I checked that are fine and need no change:

The clear-then-write hash pattern at the two true in-place rewrite sites (edit resolution, assistant-edit version reuse) fails in the safe direction.
The PUT version-replace route looks like a third in-place site but isn't: it uploads to a fresh storage key and switches storage_path + content_sha256 in one atomic row update, so a failure between upload and update leaves the old path with its old, still-correct hash.
The unauthenticated /manifest-signing-key endpoint is a deliberate and sound choice — verifiers are outside the workspace and must not trust the key embedded in the manifest.
Inline comments cover the remaining items: a defensive gap in canonicalize, per-request key derivation with no boot-time validation, a metadata/hash consistency nit in the bulk-copy path, and a question about soft-deleted versions appearing in the export.

The two pre-existing issues flagged in the PR description (selectAll pagination ties, concurrent accept/reject races) should be filed as separate issues regardless of the outcome here — thanks for calling them out rather than bundling fixes.

amal66 (approved):

b1rdmania:
Thanks, that's a properly careful read - especially working out that the PUT version-replace route isn't a third in-place site. All four addressed in 18a0051.

**canonicalize** - taken, and you're right that it's the same failure as the NaN case. Rejecting anything whose prototype isn't `Object.prototype` or `null`, with a test covering Date, Buffer and Map, plus null-prototype objects still passing. Left cycles unguarded for the reason you gave.

**Boot validation** - added to `index.ts`, so a malformed key stops startup with the same message instead of surfacing on someone's first export. A working key logs its key id, which is useful for confirming which key a deployment is on after a rotation. I skipped the derivation cache: Ed25519 derivation is cheap, reading the env var each time means a rotation takes effect without a restart, and a seed-keyed cache is a bit of state I'd rather not have in the signing path.

**Bulk copy metadata** - taken, `size_bytes: raw.byteLength` unconditionally. The false-mismatch-on-stat case is exactly the sort of thing that would waste someone's afternoon.

**Soft-deleted versions** - intended, and your reasoning matches mine: a trail that drops deleted versions is a weaker attestation. Documented in the README, including the consequence that a deleted version's filename and timestamps stay visible to anyone with project access.

On the two pre-existing issues, @willchen96 picked those up in Discord. The paging one is a tiebreaker on the sort. For the concurrent accept/rejects we're landing on compare-and-swap rather than locking, since the bug is a lost update rather than a workflow problem - both as separate PRs, not bundled here.

willchen96 (approved):

PR #293, opened 2026-08-05
[Testing 17] ci: schema-drift check — fresh install vs upgraded deployment must converge

amal66:
> **Stacked on #256** — this branch contains #256's commits, so the diff shows both until it merges. The new work here is the last four commits: `.github/workflows/schema-drift.yml`, `backend/scripts/schema-fingerprint.sql`, `backend/migrations/20260805_01_narrow_service_role_grants.sql`, and a `backend/schema.sql` fold-in. Draft until #256 lands.

**What this adds**

The follow-up promised in #256's review: upgrade-path drift testing, designed the way the review described — a pinned baseline from an older point in history, only the migrations dated after it, and a schema comparison against a fresh install.

Mike maintains the database shape twice by design: `backend/schema.sql` is what a fresh install runs; `backend/migrations/` is what an existing deployment applies to move forward. Nothing structurally forces the two to agree, and when they drift apart, fresh installs and upgraded deployments silently diverge — no ordinary test notices, because test databases are always built fresh.

The new `Schema drift` workflow builds **both real installation paths** in one disposable Supabase stack:

```text
upgraded:  schema.sql @ pinned baseline commit (9a1277b, the commit that
           introduced dated migrations)
           + only the migrations git says were ADDED since that commit
fresh:     today's schema.sql
```

Each database is reduced to a canonical fingerprint — tables, columns, constraints, indexes, RLS policies, full function definitions, triggers, views, enums, and per-privilege ACLs, every section totally ordered — and the two fingerprints are diffed. Any hunk is a structural or security difference between what an upgraded deployment has and what a fresh install has.

Notably, this is **not** the replay-history approach reverted in #256: no old migration ever runs against a schema from its future, and the check cannot be "fixed" by editing shipped migrations (the failure message explicitly says so). A fingerprint is used instead of a raw `pg_dump` diff because upgraded and fresh databases legitimately differ in physical column order and in GRANT execution order; the fingerprint sorts columns by name and explodes ACLs per privilege, so those benign differences are ignored while every real one is kept.

**It caught two instances of real drift — one in each direction**

1. **Fresh installs are missing a live table.** Migration `20260629_01` created `workflow_open_source_submissions` (merged 2026-07-04), but the table was never folded into `schema.sql` — `git log -S` shows it has never appeared there. The table is queried by `backend/src/routes/workflows.ts`, `lib/userDataExport.ts`, and `lib/userDataCleanup.ts`, so on every fresh install today, submitting a workflow to the open-source queue or exporting/cleaning up user data hits `relation does not exist`. Upgraded deployments were fine. **Fix: fold the table into `schema.sql`** (verbatim from the migration, placed with the other workflow tables, plus its deny-all revoke line).

2. **Upgraded deployments have over-broad service_role grants.** Commit b2dbb39 ("narrow service role schema grants", 2026-07-23) tightened `service_role` from `GRANT ALL` to `select, insert, update, delete` — in `schema.sql` only, with no migration. A deployment created before that date and upgraded by the documented path still lets `service_role` `TRUNCATE` any table, create `TRIGGER`s, and reset sequences; a fresh install does not. Fresh installs were fine. **Fix: new dated migration** `20260805_01_narrow_service_role_grants.sql` revoking the excess privileges.

The symmetry is the point: each drift was fixed with the tool appropriate to its direction — snapshot fold-in when the migration was right, new migration when the snapshot was right — and never by touching a shipped migration. With both fixes, the check runs green: an upgraded deployment and a fresh install now converge on the identical schema, down to per-privilege ACLs.

**Maintainer notes**

- The baseline pin lives in the workflow as `BASELINE_REF` with instructions: advance it deliberately to a newer known-good sync point, never to silence a failure.
- If this check goes red on a future PR, the failure message explains the two legitimate fixes: fold the missing change into `schema.sql`, or ship a new dated migration.
- The stack-tests suite from #256 passes against the updated `schema.sql`, so the fold-in is exercised by real RLS tests, not just the fingerprint.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -7,9 +7,9 @@
 # this job "e2e / playwright" as a required status check in branch protection —
 # see docs/e2e-ci.md ("Make it merge-blocking").
 #
-# The suite is green with NO secret: the 4 LLM-dependent specs (chat + critical
-# path) skip themselves when ANTHROPIC_API_KEY is absent (see e2e/llm.ts), so a
-# keyless run passes the other 23 specs. The gate exists because the app has no
+# The suite is green with NO secret: of the 31 specs, the 4 LLM-dependent ones
+# (chat + critical path) skip themselves when ANTHROPIC_API_KEY is absent (see
+# e2e/llm.ts), so a keyless run passes the other 27 specs. The gate exists because the app has no
 # keyless model — with no provider key configured, ChatInput blocks every send —
 # not because of the (best-effort) title-generation call. To run and enforce
 # those 4, set the ANTHROPIC_API_KEY repository secret; exact steps, the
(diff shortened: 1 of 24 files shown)


willchen96:
Thanks this is useful. 

Although to note that workflow submissions table will be removed and added to the landing page app instead which is private so it is not integrated into the OSS app which does not need such functionality.

willchen96 (approved):

PR #298, opened 2026-08-07
add raw File bytes streaming to API

drewmiller:
## Summary

This adds raw File access to the backend API for serving raw bytes to the frontend without pre-rendering.

## Why / Motivation

This is in connection with enabling online file editing.

## Tradeoffs & risks

There shouldn't be any tradeoffs.

## How verified

We're building an integration which leverages the new API endpoints.

## Checklist

- [N/A] Updated docs / env examples if setup, config, or behavior changed.

### DIFF
--- a/backend/src/routes/documents.ts
+++ b/backend/src/routes/documents.ts
@@ -174,6 +174,45 @@ documentsRouter.get("/:documentId/display", requireAuth, async (req, res) => {
   }
 });
 
+// GET /single-documents/:documentId/file
+// Streams active-version source bytes for browser editors. Unlike /display,
+// this never substitutes a generated PDF rendition for an Office document.
+documentsRouter.get("/:documentId/file", requireAuth, async (req, res) => {
+  const userId = res.locals.userId as string;
+  const userEmail = res.locals.userEmail as string | undefined;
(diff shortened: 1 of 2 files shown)


willchen96:
going to close this for now as 1. is does not actually stream the file bytes since it uses the downloadFile function 2. i would favour a toggle on the file fetch endpoint instead of an extra end point.

drewmiller:
1. We used the term "stream" here to mirror the existing comment in the `docx` endpoint, but I agree they do imply a functionality that doesn't exist since it's shipping the whole file.

image

which is implemented with

image


2. I'm not sure I follow. The toggle in this case seems to be in the endpoint route itself, as demonstrated by the existing suite of get routes:

`/:documentId/versions`
`/:documentId/docx`
`/:documentId/file`
`/:documentId/display`
`/:documentId/url`, etc.

It seems to me perhaps an alternative would be to use the existing `docx` and add a `pdf`, but then we'd need one for each supported type. 

If we try to add an additional parameter to one of the existing endpoints, it's likely to be a breaking change that would have to wait for a new version, no?

We've got an integration that needs this endpoint, so let us try to clean up some of that stuff and we'll re-submit a pull request.

PR #299, opened 2026-08-07
feat(word-addin): align chat UI with web app

willchen96:
## Summary

- redesign the Word task pane with a floating liquid-glass header, Assistant/Quick Actions/Workflows navigation, chat history, new chat, and the animated personalized initial view
- align the composer with the web app through model selection, workflow attachment, Desktop Files/Web files document sources, replicated document/workflow modals, shared dropdown/modal primitives, and skeuomorphic SVG assets
- keep document context and tracked-edit suggestions enabled, retain tracked changes for existing-text edits, and remove the insert-below controls and document insertion helper
- harden add-in API configuration, SVG bundling, responsive task-pane layout, documentation, and end-to-end coverage

Screenshot 2026-08-12 at 9 26 18 PM
Screenshot 2026-08-12 at 9 26 32 PM
Screenshot 2026-08-12 at 9 27 22 PM
Screenshot 2026-08-12 at 9 27 13 PM
Screenshot 2026-08-12 at 9 27 02 PM
Screenshot 2026-08-12 at 9 26 43 PM

## Testing

- `npm run typecheck`
- `npm run build:e2e`
- `npm run test:e2e` — 59 passed, 7 skipped (Projects is intentionally not exposed)

## Build note

The deployment production build requires `REACT_APP_API_BASE_URL`, `REACT_APP_SUPABASE_URL`, `REACT_APP_SUPABASE_ANON_KEY`, and `REACT_APP_WEB_APP_URL`; the environment-free verification uses the repository’s `build:e2e` configuration.


word-addin/src/taskpane/components/primitives/LiquidActionRow.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+export function LiquidActionRow({
+  children,
+  className,
+  ...props
+}: React.ComponentProps<"div">): React.ReactElement {
+  return (
+    <div
+      className={cn(
+        "flex shrink-0 items-center gap-2 rounded-full border border-white/70 bg-app-surface px-1 py-1 shadow-[0_8px_24px_rgba(15,23,42,0.06)] backdrop-blur-2xl",
```
amal66, line 12: **[High] LiquidActionRow uses app-surface utilities whose tokens are not defined in the add-in, so header pills have no background and no hover/active feedback**
*UI alignment · independently verified*

LiquidActionRow/LiquidIconButton/LiquidTextButton (the floating header's menu bubble, New chat, Chat history, workflow Use/Back controls) use bg-app-surface, hover:bg-app-surface-hover, and active:bg-app-surface-active. The web app defines --color-app-surface / -hover / -active in frontend/src/app/globals.css @theme inline, but the add-in's copied token sheet (word-addin/src/shared/styles/tokens.css) omits every --app-* token. Under Tailwind v4, utilities referencing unknown theme colors are silently not generated, so the pill row renders transparent (border+shadow only, unlike the web's near-white #fdfdfe surface) and none of the header buttons show any hover or pressed state. Verified empirically: compiling word-addin/src/taskpane/styles.css through @tailwindcss/postcss emits no rule for bg-app-surface or app-surface-hover while bg-primary is emitted.

**Suggested fix:** Add the --app-background/--app-surface/--app-surface-hover/--app-surface-active/--app-floating custom properties and their @theme inline --color-app-* mappings to word-addin/src/shared/styles/tokens.css (copy the block from globals.css), or switch LiquidActionRow to tokens that exist in the add-in sheet.

<details><summary>Evidence</summary>

Compiled the add-in stylesheet with its own @tailwindcss/postcss: `.bg-app-surface` generated: false, `app-surface-hover` generated: false, `.bg-primary` generated: true. grep confirms --app-surface is defined nowhere under word-addin/; frontend/src/app/globals.css defines --app-surface: #fdfdfe etc. and the web's liquid-dropdown/AppSidebar rely on those same classes.

</details>

word-addin/src/taskpane/App.tsx
```diff
@@ -37,66 +68,165 @@ export default function App(): React.ReactElement {
...
-            >
-              {tab.label}
-            </TabPillButton>
-          );
-        })}
-      </nav>
-
-      {/* Active tab content */}
-      <div className="flex flex-1 flex-col overflow-hidden">{renderTab()}</div>
+        {renderSection()}
```
amal66: **[High] Section navigation destroys the live conversation while keeping its chatId, silently appending later sends to the invisible chat**
*Chat runtime · independently verified*

App renders exactly one section via renderSection(), so ChatPanel (and useWordAssistantChat, which owns the messages array in hook state) unmounts whenever the user opens Quick Actions, Workflows, History, or Settings from the header menu. changeSection() intentionally does NOT reset chatId/initialMessages (only startNewChat/openSelectedChat do). On returning to the Assistant section, ChatPanel remounts and the hook re-seeds messages from the stale initialMessages prop — [] for any chat started in this session, or the original history snapshot for a chat opened from history. The user sees a blank initial view (or stale snapshot), yet chatId still points at the old chat: the next send posts history=[newUserMessage] to the same chat_id, so (1) the model loses all prior conversational context (backend builds context only from client-sent messages), and (2) in cloud/local storage the new turn is appended to a chat the user believes is gone — history later shows one chat mixing both 'conversations'. Also, if a stream is in flight when the section switches, the unmount cleanup aborts it with no warning. This is an everyday flow (peek at Quick Actions or History mid-chat, or the designed Workflows → 'Use' → chat path) — users will hit it. The e2e suite (chat.spec.ts section-menu tests) exercises the menu but never asserts conversation survival across a round trip.

**Suggested fix:** Either keep ChatPanel mounted (render all sections, hide inactive ones), lift messages state up to App next to chatId, or on ChatPanel remount with a non-null chatId re-fetch the chat (getCloudWordChat/getLocalWordChat) into initialMessages before allowing a send.

<details><summary>Evidence</summary>

App.tsx:90-137 renderSection() returns a single section component; :153-159 changeSection() only clears workflow state, never chatId/initialMessages/sessionKey; useWordAssistantChat.ts:59 messages live in hook state, :69-78 unmount aborts the stream, :80-95 mount effect re-seeds from initialMessages. No code path reloads messages for a non-null chatId on remount (getCloudWordChat/getLocalWordChat are only called from ChatHistoryList on explicit selection).

</details>

backend/src/routes/wordChat.ts
```diff
@@ -0,0 +1,515 @@
...
+
+  try {
+    write(
+      `data: ${JSON.stringify({
+        type: "chat_id",
+        chatId,
+        assistantMessageId,
+      })}\n\n`,
+    );
+    const { fullText, events, citations } = await runLLMStream({
```
amal66: **[Medium] ask_inputs tool is exposed to /word-chat but the route and add-in cannot handle it**
*Backend · independently verified*

runLLMStream always includes the full TOOLS set (streaming.ts:201 `baseTools = [...TOOLS, ...]`), which contains `ask_inputs` (toolSchemas.ts:125) whose description tells the model to use it 'when required documents have not been attached'. The word route only disables research tools (`includeResearchTools: false`). Unlike chat.ts, wordChat.ts never parses `ask_inputs_response` and the add-in has no ask_inputs UI (grep of word-addin/src returns zero hits; wordPrompt.ts never mentions it). If the model calls ask_inputs in a Word chat — plausible, since the composer supports document sources — the stream pauses (AssistantStreamAskInputsPause), persists an event the client can't render, and the user sees the assistant silently stop with no way to answer the question. The `messageTable` parameter added to appendAskInputsResponseToLastAssistantMessage suggests this was anticipated but never wired up.

**Suggested fix:** Either filter ask_inputs out of the tool set for /word-chat (add an option like includeAskInputs to runLLMStream), or instruct against it in buildWordChatSystemPrompt and add ask_inputs_response support to the route + add-in.

<details><summary>Evidence</summary>

wordChat.ts POST body parsing has no parseOptionalAskInputsResponse (lines 204-229); streaming.ts:199-201 builds tools as `[...TOOLS, ...researchTools, ...WORKFLOW_TOOLS]` with no way to exclude ask_inputs; `grep -rn ask_inputs word-addin/src` → no matches.

</details>

backend/src/routes/wordChat.ts
```diff
@@ -0,0 +1,515 @@
...
+    .from("word_chat_messages")
+    .select("*")
+    .eq("chat_id", req.params.chatId)
+    .order("created_at", { ascending: true });
+  if (error) return void res.status(500).json({ detail: error.message });
+  res.json({ chat, messages: messages ?? [] });
+});
+
+// POST /word-chat — Word-specific streaming endpoint.
+wordChatRouter.post("/", requireAuth, async (req, res) => {
```
amal66, line 233: **[Medium] POST /word-chat duplicates ~150 lines of chat.ts streaming orchestration — drift already visible**
*Backend · independently verified*

The reservation insert, 3-attempt updateAssistantMessage retry helper, SSE header block, abort-close wiring, AssistantStreamError partial-persistence, error-event persistence, and title-truncation logic in wordChat.ts:359-514 are near-verbatim copies of the same code added to chat.ts:526-720 in this same PR (updateReservedAssistantMessage, identical retry loop, identical catch structure). Any future fix to one (e.g., backoff in the retry loop, changed error semantics, [DONE] ordering) must be manually mirrored. Drift has already started: chat.ts hydrates edit statuses on read and handles ask_inputs; wordChat.ts bumps updated_at while chat.ts does not — differences that are intentional today but indistinguishable from bugs tomorrow.

**Suggested fix:** Extract a shared helper (e.g., persistStreamedAssistantTurn({ db, table, chatId, assistantMessageId, write }) plus a shared runStreamedChatTurn wrapper) parameterized by message table, the same way contextBuilders.ts was parameterized with messageTable in this PR.

<details><summary>Evidence</summary>

Compare wordChat.ts:384-400 with chat.ts updateReservedAssistantMessage (identical 3-iteration retry over .update().eq(id).eq(chat_id)), and the catch blocks wordChat.ts:466-510 vs chat.ts's abort/error persistence — same structure, different table name and log prefix.

</details>

word-addin/src/shared/styles/tokens.css
```diff
@@ -1,11 +1,10 @@
 /*
  * Shared design tokens for the Mike design system.
  *
- * A verbatim subset of the web app's tokens (frontend/src/app/globals.css),
- * consumed today only by the Word add-in (word-addin/src/taskpane/styles.css);
- * the web app still keeps its own copy until the shared package lands. Keep
- * values in sync with globals.css. Import AFTER `@import "tailwindcss";` so
- * the `@theme` mappings and base layer apply.
+ * A local subset of the web app's tokens (frontend/src/app/globals.css),
```
amal66, line 4: **[Medium] PR converts the shared design system from a keep-in-sync vendored copy into a declared local fork, and drift is already visible**
*UI alignment · independently verified*

main aliased word-addin/src/vendor/shared as @mike/shared with a header saying 'A verbatim subset of the web app's tokens... Keep values in sync'; this PR renames it to src/shared and rewrites the header to 'A local subset... adapted for the Word add-in', dropping the @mike/shared alias from webpack.config.js and tsconfig.json entirely. Given the project rule that all client UI/UX must mirror the main web app and the intended single-shared-package direction, this formalizes a fork, and divergence has already begun: shared/chat/ChatInput.tsx is a new composer shell with a leftSlot/rightSlot/attachments API the web's frontend/src/app/components/assistant/ChatInput.tsx does not have; shared/ui/button.tsx dropped the web button's cva variant/size API (frontend/src/app/components/ui/button.tsx) down to a single hardcoded style; the model catalog is re-declared as STATIC_MODELS in the add-in ModelToggle instead of sharing the web's exported MODELS list, so the next model launch must be edited in two places or the surfaces silently disagree.

**Suggested fix:** Keep the 'keep in sync with <web path>' provenance comments on every duplicated file (tokens.css lost its sync mandate; button.tsx and ChatInput.tsx have none), and track the shared-package extraction so the fork is explicitly temporary. At minimum restore a sync checklist (tokens, MODELS list, composer classes) in word-addin/README.md.

<details><summary>Evidence</summary>

git diff main...HEAD tokens.css comment change; webpack.config.js diff removes the '@mike/shared' alias block; word-addin ModelToggle.tsx:25-35 duplicates frontend ModelToggle.tsx:25-36 verbatim; word-addin/src/shared/ui/button.tsx (21 lines, no variants) vs frontend ui/button.tsx (cva with 6 variants/6 sizes).

</details>

word-addin/src/taskpane/components/shell/FloatingHeader.tsx
```diff
@@ -0,0 +1,195 @@
+import React, { useState } from "react";
+import type { Message } from "../../types";
+import { Check, ChevronLeft, Ellipsis, Menu, Plus, X } from "lucide-react";
+import {
+  LiquidActionRow,
+  LiquidIconButton,
+  LiquidTextButton,
+} from "../primitives/LiquidActionRow";
+import chatIcon from "../../../../../frontend/public/icons/app-sidebar/chat.svg";
```
amal66: **[Medium] Add-in components import SVG assets by relative path out of frontend/public, coupling the two builds; 4 icons were added to frontend that frontend never uses**
*UI alignment · independently verified*

FloatingHeader.tsx:9-13, InitialView.tsx:2, DirectoryIcons.tsx:3-10, and DocumentSourceMenu.tsx:10-11 import icons via ../../../../../frontend/public/icons/... The word-addin bundle now breaks if it is built without the frontend directory present (the add-in previously vendored everything under src/ precisely so it could build standalone), and any rename/move inside frontend/public silently breaks the add-in. The PR also adds chat-history.svg, desktop.svg, earth.svg, and settings.svg to frontend/public/icons/app-sidebar/ although no frontend source references them (grep over frontend/src returns nothing), so the web app ships dead assets that exist only to serve the add-in's cross-package imports.

**Suggested fix:** Copy the needed SVGs into word-addin (e.g. src/assets/icons/) or move genuinely shared icons into the shared directory both surfaces import from. Reusing the identical artwork is right; reaching across package roots is not.

<details><summary>Evidence</summary>

grep -rn 'frontend/' word-addin/src lists 16 cross-package imports; grep -rn 'chat-history.svg|desktop.svg|earth.svg|settings.svg' frontend/src returns no matches; word-addin/tsconfig.json include is ["src"] so these files are outside the project's own compilation root.

</details>

word-addin/src/taskpane/components/assistant/ChatInput.tsx
```diff
@@ -0,0 +1,318 @@
...
+              </button>
+            </div>
+          )}
+          <ChatInputShell
+            value={input}
+            onValueChange={setInput}
+            onSubmit={submit}
+            isLoading={isResponseLoading}
+            onCancel={onCancel}
+            disabled={isResponseLoading}
```
amal66: **[Medium] Composer is fully disabled while a response streams, and submit is not gated on API-key availability — both diverge from the web composer**
*UI alignment · independently verified*

The add-in passes disabled={isResponseLoading} to the shared shell, which sets disabled on the textarea (shared/chat/ChatInput.tsx:131), so users cannot type or compose their next message while the assistant streams — the web ChatInput never disables its textarea (frontend assistant/ChatInput.tsx:485-506 has no disabled attr), only the send button. Disabling also drops keyboard focus from the textarea mid-conversation. Separately, the web composer blocks submission when the selected model's provider key is missing and opens ApiKeyMissingPopup (frontend ChatInput.tsx:309-311); the add-in's ModelToggle shows the red AlertCircle for unavailable models but submit() (line 153-176) sends anyway, surfacing only a post-hoc requestError banner after the backend rejects the call.

**Suggested fix:** Pass disabled={false} (or drop the prop) and rely on isLoading to gate submission, matching the web; add a pre-submit availability check reusing the key status the add-in's ModelToggle already fetches via getApiKeyStatus().

<details><summary>Evidence</summary>

word-addin assistant/ChatInput.tsx:221 `disabled={isResponseLoading}`; shared/chat/ChatInput.tsx renders `disabled={disabled}` on the textarea; frontend ChatInput.tsx submitMessage(): `if (apiKeys && !isModelAvailable(model, apiKeys)) { setApiKeyModalProvider(...); return; }` — no equivalent in the add-in submit path.

</details>

word-addin/src/taskpane/hooks/useWordDoc.ts
```diff
@@ -1,280 +1,1083 @@
...
+    .filter((change) => String(change.type).toLowerCase() === "deleted")
+    .map((change) => change.text)
+    .join("");
+  const onlyExpectedTypes = changes.every((change) => {
+    const type = String(change.type).toLowerCase();
+    return type === "added" || type === "deleted";
+  });
+  return (
+    onlyExpectedTypes &&
+    addedText === toWordText(edit.replacement) &&
```
amal66: **[Medium] Multi-paragraph replacements likely fail revision verification in real Word (mock encodes '\n' verbatim, hiding it)**
*Office.js · independently verified*

trackedChangesMatchEdit requires the concatenated 'added' revision text to strictly equal toWordText(edit.replacement), which joins paragraphs with '\n' (wordText.ts:58). In real Word, insertText converts newlines into paragraph/line breaks and Range/TrackedChange .text renders paragraph marks as '\r' (line breaks as '\v'), and a multi-paragraph insertion is typically split into per-paragraph Added revisions. So for any replacement that toWordText turns into more than one line, the strict equality fails: at apply time exactRevisions becomes false with reason 'unexpected-revisions' (line 931-934) and entry.changes is left empty (line 1005); the card still shows Accept/Reject, but resolution then goes through resolveThroughAnchor (line 315), which re-runs the same failing comparison and returns 'changed', surfacing ChangedRevisionSetError: 'The revisions in this passage changed after Mike applied the edit' (lines 440-447). Accept/Reject therefore never works for these edits, and after reload restoreTrackedEdit (line 555) degrades them to view-only. The e2e office mock stores the inserted string verbatim (office-mock.ts:595-599, createStoredRevisionGroup(entry, query, newText)), so the entire e2e suite cannot detect this host-representation mismatch. The system prompt keeps ORIGINAL single-paragraph but places no such restriction on REPLACEMENT (backend/src/lib/chat/wordPrompt.ts:22).

**Suggested fix:** Normalize break characters on both sides before comparing (e.g. collapse /[\r\n\v]+/ to a single '\n' in both the Word-reported revision text and toWordText output), or compare per-paragraph segments. Also consider seeding the office mock with '\r' paragraph marks so e2e exercises real host text representation.

<details><summary>Evidence</summary>

useWordDoc.ts:249-253: `addedText === toWordText(edit.replacement) && deletedText === edit.original` where wordText.ts:58 is `toWordParagraphs(value).join("\n")`; office-mock.ts:595 stores `newText` unchanged as the Added revision text.

</details>

amal66 (commented):
## Multi-agent review — 6 analysts + 6 adversarial verifiers

Each dimension (backend API, security/privacy, UI alignment with the web app, Office.js/tracked edits, chat runtime/streaming, tests/tooling) was analyzed by a dedicated agent, and every finding was then re-checked against the code by an independent verifier instructed to refute it. Of 38 raw findings, none were refuted, one is marked uncertain, two were downgraded, and cross-analyst duplicates were merged, leaving **34 findings: 2 high, 10 medium, 16 low, 6 nit**. No critical or security-breaking issues — auth, ownership checks, and RLS on the new word-chat tables held up.

**Overall: solid, shippable feature with two high-severity fixes needed first:**
1. Header pills have no background/hover state — `LiquidActionRow` uses `app-surface` utilities whose tokens don't exist in the add-in bundle.
2. Section navigation destroys the live conversation while keeping its `chatId`, so later sends are appended to a chat the user can no longer see.

Recurring themes: the `vendor/` → `src/shared/` move turns the shared design system into a declared local fork with drift already visible; `wordChat.ts` duplicates ~150 lines of `chat.ts` streaming orchestration (drift already started); debug `console.log`s ship in the production pane (including full user message content and a literal "This is not supposed to run"); and the document-identity UUID embedded in the .docx travels with every copy of the file.

---

### Findings not anchorable to a diff line (4)

<details><summary><b>[Medium]</b> Backend CORS allowlist has no entry for a deployed add-in origin — <code>backend/src/app.ts:118</code></summary>

The API's CORS allowlist contains exactly one origin (FRONTEND_URL, default http://localhost:3000). This PR introduces a production build path for the add-in served from WORD_ADDIN_PUBLIC_URL — a different HTTPS origin — whose bundle calls REACT_APP_API_BASE_URL directly (the webpack same-origin proxy exists only in dev). A deployed task pane's fetches to /word-chat will be blocked by the browser because the response carries no Access-Control-Allow-Origin for the add-in's origin. Neither the PR's backend changes nor word-addin/README's 'Production build' section addresses this, which pressures operators toward the unsafe workaround of wildcarding CORS instead of adding a second explicit allowed origin (e.g. a WORD_ADDIN_URL env var).

**Suggested fix:** Add an optional WORD_ADDIN_URL (or comma-separated ALLOWED_ORIGINS) env var to the allowedOrigins set, and document in word-addin/README's Production section that the API must either share the add-in's origin behind a reverse proxy or explicitly allowlist WORD_ADDIN_PUBLIC_URL.

</details>

<details><summary><b>[Medium]</b> E2E hermeticity depends on enumerating eager endpoints; no catch-all guard — <code>word-addin/e2e/support/fixtures.ts:169</code></summary>

The fixture registers default routes for exactly the four requests known to fire on mount (user/api-keys, models/ollama, user/profile, GET word-chat?*), each with a comment explaining it exists because an unmocked request escapes to a developer's live backend where "a 401 would clear the seeded session mid-test". That failure mode was clearly hit repeatedly, yet there is still no last-resort route: the next component that fetches on mount will silently leak to localhost:3001/54321 again and produce environment-dependent flakes rather than a loud failure.

**Suggested fix:** Register one catch-all route first (so it matches last in Playwright's newest-first order) that aborts or fulfills-with-500 any request to the API/Supabase origins, making an unmocked call fail the test deterministically.

</details>

<details><summary><b>[Low]</b> Office mock cannot catch missing load()/sync() or Word search semantics, and the PR removes the only real-Word run — <code>word-addin/e2e/support/office-mock.ts:315</code></summary>

In the shim, context.sync() is `() => Promise.resolve()` and every load() is a no-op, so all proxy properties are always readable. Real Word throws PropertyNotLoaded/GeneralException when a property is read without load()+sync or when a range handle goes stale outside the seeded opt-in knobs — a whole class of Word.run misuse passes the 4,750-line suite and fails only in the host. body.search is plain case-folded substring matching (lines 533-549), so Word's ^-control-code parsing and its ~255-char search failure never occur in tests (the app's guard at useWordDoc.ts:830 is itself only validated against the mock). At the same time the PR deletes e2e-live/live-demo.spec.ts and playwright.live.config.ts, leaving no automated execution against real Office at all — only manual-session.mjs.

**Suggested fix:** Cheap fidelity upgrade: have mock ranges/collections throw on property reads unless load() was called and a sync() has since resolved — this catches the most common real-Word breakage class. Document in the README that redline/tracked-edit changes require a manual pass via e2e-live/manual-session.mjs.

</details>

<details><summary><b>[Low]</b> Fixed-delay timing patterns in streaming and scroll tests are flake/weak-assertion risks — <code>word-addin/e2e/chat.spec.ts:833</code></summary>

The stop-control test holds the /word-chat response for a fixed 1500ms (`holdMs: 1500`) and must complete ~8 assertions including two expect.poll calls inside that window; on a loaded CI worker the stream can resolve mid-assertion and the Stop button disappears (masked locally by retries: 2 on CI only). The scroll-spacer test (lines 372-391) uses two fixed 750ms waits with a <2px pixel tolerance on animated scroll positions. Separately, several negative assertions are gated on 50-100ms settles (chat.spec.ts:191, 1277; history.spec.ts:228, 342), which only prove the regression didn't appear within that window.

**Suggested fix:** Replace holdMs with an explicitly released gate promise inside the route handler — auth.spec.ts already demonstrates this pattern with refreshGate in the sign-out-during-refresh test — and poll for positional stability instead of sleeping twice.

</details>


word-addin/src/taskpane/hooks/usePaginatedChats.ts
```diff
@@ -0,0 +1,116 @@
...
+    let timedOut = false;
+    let timeoutId: number | null = null;
+    requestPendingRef.current = true;
+    if (limit === pageSize) setLoading(true);
+    else setLoadingMore(true);
+    setError(null);
+
+    const request =
+      storageMode === "cloud"
+        ? listCloudWordChats(documentId, limit + 1, controller.signal)
```
amal66: **[Medium] Chat history pagination silently dead-ends at 100 chats: client's limit+1 probe collides with the backend's 100 cap**
*Chat runtime · independently verified*

usePaginatedChats fetches limit+1 rows and sets hasMore = next.length > limit. The backend clamps limit to 100 (wordChat.ts:127-128: Math.min(Math.max(requestedLimit,1),100)). Once the client's limit reaches 100 (10 'load more' clicks in the dropdown at pageSize 10), it requests 101, the server silently returns at most 100, next.length (100) is never > limit, so hasMore flips to false and every chat beyond the 100 most recent becomes permanently unreachable from the add-in with no error or indication. The same collision also mis-reports hasMore at exactly 100 chats. The local-storage path is unaffected (listLocalWordChats has no cap).

**Suggested fix:** Use real offset/cursor pagination (e.g. created_at cursor or range header) instead of refetching an ever-growing limit, or at minimum raise/remove the server cap awareness client-side (stop offering Load more when limit+1 exceeds the server max, and surface that more exist).

<details><summary>Evidence</summary>

usePaginatedChats.ts:53-56 `listCloudWordChats(documentId, limit + 1, controller.signal)` + :68-69 `setHasMore(next.length > limit)`; backend/src/routes/wordChat.ts:126-129 clamps the requested limit to 100; loadMore (usePaginatedChats.ts:96-101) grows limit unboundedly with no awareness of the cap.

</details>

backend/src/__tests__/integration/wordChat.routes.test.ts
```diff
@@ -0,0 +1,220 @@
...
+  dbState.document = {
+    data: { id: "word-document-row-1" },
+    error: null,
+  };
+  dbState.chatList = { data: [], error: null };
+  dbState.chatDetail = { data: null, error: null };
+  dbState.messages = { data: [], error: null };
+}
+
+describe("Word chat history routes", () => {
```
amal66, line 123: **[Medium] POST /word-chat authorization and validation paths have no tests**
*Tests & tooling · independently verified*

The new wordChat.routes.test.ts covers only GET list/detail error-vs-404 discrimination. Neither it nor the word-chat tests added to chat.routes.test.ts exercise the POST negative paths: 400 for a missing/invalid document_id, 400 for an invalid storage value, and — most importantly — 404 when resuming a chat_id that fails the ownership/document scoping lookup (wordChat.ts lines 253-268). Because the route uses the service-role Supabase client, the .eq("user_id")/.eq("word_document_id") filters in getAccessibleWordChat ARE the authorization; the permissive mock in chat.routes.test.ts always returns a chat row, so dropping one of those filters on the POST resume path would pass the entire suite. Only the GET detail test pins the filter chain.

**Suggested fix:** Add POST tests: invalid document_id -> 400, storage:"weird" -> 400, and a resume case where dbState.chatDetail is null -> 404 with no word_chat_messages insert and no runLLMStream call. Reuse the wordChat.routes.test.ts recordedQueries harness to pin the POST-path filter chain the way the GET detail test does.

<details><summary>Evidence</summary>

chat.routes.test.ts word-chat tests (lines 265, 341, 637, 652, 686) all send either no chat_id or a chat_id the always-succeeding mock accepts; grep for "document_id must", "storage must", or a 404 assertion against POST /word-chat returns nothing. wordChat.routes.test.ts asserts filters only for GET /word-chat/:chatId (lines 200-206).

</details>

backend/src/routes/wordChat.ts
```diff
@@ -0,0 +1,515 @@
...
+    .eq("user_id", userId)
+    .order("updated_at", { ascending: false });
+  if (limit) query = query.limit(limit);
+  const { data, error } = await query;
+  if (error) return void res.status(500).json({ detail: error.message });
+  res.json((data ?? []).map((chat) => ({ ...chat, project_id: null })));
+});
+
+// GET /word-chat/:chatId?document_id=<embedded document UUID>
+wordChatRouter.get("/:chatId", requireAuth, async (req, res) => {
```
amal66, line 175: **[Low (analyst said medium; verifier downgraded)] Non-UUID chat ids return 500 with raw Postgres error instead of 404; DB error text leaks to clients**
*Backend · independently verified — independently flagged by the Security analyst too*

GET /word-chat/:chatId passes req.params.chatId straight into `.eq("id", chatId)` on a uuid column, and POST accepts any non-empty string as chat_id (parseOptionalChatId, requestValidation.ts:47-54, does not enforce UUID even though the route defines UUID_PATTERN for document_id). Against a real database, `/word-chat/chat-1?...` produces PostgREST error `invalid input syntax for type uuid: "chat-1"`, which the route returns verbatim: lines 137, 150, 168, 181, 191 all do `res.status(500).json({ detail: error.message })`. So a malformed id yields a 500 with internal DB error text rather than a 404. Notably the new test wordChat.routes.test.ts uses chatId "chat-1" and expects 404 — it only passes because the DB is mocked; the real behavior differs. Raw error.message passthrough is also inconsistent with safeErrorMessage used later in the same file for stream errors.

**Suggested fix:** Validate chatId (path param and body chat_id) against UUID_PATTERN and return 404/400 early; replace `detail: error.message` with a generic message plus safeErrorLog server-side.

<details><summary>Evidence</summary>

wordChat.ts:36-37 defines UUID_PATTERN but applies it only to document_id; getAccessibleWordChat (line 104-111) returns `{ ok: false, detail: error.message }` which line 181 sends to the client; wordChat.routes.test.ts:212-219 asserts 404 for `/word-chat/chat-1` against a mock that cannot reproduce the uuid cast error.

</details>

backend/src/routes/wordChat.ts
```diff
@@ -0,0 +1,515 @@
...
+    docAvailability,
+    buildWordChatSystemPrompt(),
+    docIndex,
+    false,
+    nonce,
+  );
+  const workflowStore = await buildWorkflowStore(userId, userEmail, db);
+  const assistantMessageId = randomUUID();
+
+  if (persistChat) {
```
amal66, line 410: **[Low] Reservation pattern can strand permanent null-content assistant rows in history**
*Backend · independently verified*

Both routes now insert an assistant row with content=null before streaming. If the process crashes/restarts mid-stream, if all three update attempts fail, or if a client aborts before any events accumulate (partial.events is empty so the update writes null again), the null row persists forever. GET /word-chat/:chatId (line 186-192) and GET /chat/:chatId (chat.ts:161-168) return these rows unfiltered; the add-in normalizes null content to an empty string (word-addin/src/taskpane/api/mikeApi.ts:172-188), so users see an empty assistant bubble in reloaded history. Under the previous insert-at-end design these rows could not exist.

**Suggested fix:** Filter assistant rows with null content out of GET history responses (or delete the reserved row in the abort path when no events were produced).

<details><summary>Evidence</summary>

wordChat.ts:362-368 inserts `{ role: "assistant", content: null }`; the abort path (467-483) writes `partial.events.length ? partial.events : null`; no read path filters `content is null` assistant rows.

</details>

backend/src/routes/wordChat.ts
```diff
@@ -0,0 +1,515 @@
...
+  );
+  if (!documentLookup.ok) {
+    return void res.status(500).json({ detail: documentLookup.detail });
+  }
+  const wordDocumentRowId = documentLookup.value;
+  if (!wordDocumentRowId) return void res.json([]);
+
+  let query = db
+    .from("word_chats")
+    .select("id, user_id, title, created_at")
```
amal66: **[Low] Chat list orders by updated_at but omits it from the response, and default is unbounded**
*Backend · independently verified*

GET /word-chat selects only `id, user_id, title, created_at` while sorting on updated_at, so the client cannot display or locally re-sort by recency, and a missing/invalid `limit` (`limit = null`) returns every chat for the document with no cap even though the route clamps explicit limits to 100. Also, when the assistant save fails (line 437-448) the route returns before the updated_at bump at 449-464, so the most recently active chat can sort stale.

**Suggested fix:** Select updated_at, and apply a default limit (e.g., 50) when the client sends none.

<details><summary>Evidence</summary>

Line 144 `select("id, user_id, title, created_at")` vs line 147 `.order("updated_at", ...)`; lines 126-129 leave limit null when the query param is absent; line 148 `if (limit)` skips `.limit()` entirely in that case.

</details>

word-addin/src/taskpane/lib/wordDocumentIdentity.ts
```diff
@@ -0,0 +1,68 @@
...
+function saveSettings(settings: Office.Settings): Promise<void> {
+  return new Promise((resolve, reject) => {
+    settings.saveAsync((result) => {
+      if (result.status === Office.AsyncResultStatus.Succeeded) resolve();
+      else reject(new Error(result.error?.message || "Could not save the document ID."));
+    });
+  });
+}
+
+async function getOrCreateWordDocumentId(): Promise<string> {
```
amal66, line 48: **[Low] Document identity UUID is embedded in the .docx and travels with every copy of the file**
*Security · independently verified — independently flagged by the Office.js analyst too*

getOrCreateWordDocumentId persists 'mike.word.documentId.v1' into Office document settings, which are stored inside the .docx package and travel with the file. Two consequences: (1) any external recipient of the document (in a legal product, that includes opposing counsel) can open the package and see a stable Mike identifier — disclosing that an AI assistant was used on the document and giving a correlatable ID across every derivative and draft sent out; (2) for the same user, a 'Save As' copy carries the original's ID, so the copy silently surfaces the original document's full chat history — chats about matter A's draft appearing under what the user now treats as a separate document. Cross-USER leakage is correctly prevented (the server scopes word_documents by (user_id, client_document_id), migration line 11), so this is a metadata-disclosure and same-user-bleed issue, not an access-control one.

**Suggested fix:** Document the trade-off in the README at minimum. Consider pairing the embedded ID with a locally stored (OfficeRuntime.storage) mapping or a document-content fingerprint check so a copied file can prompt 'treat as new document?', and note that the setting can be removed before sending documents externally.

<details><summary>Evidence</summary>

wordDocumentIdentity.ts:26-31 writes the UUID into Office.context.document.settings and saveAsync persists it into the file; backend/migrations/20260809_01_word_addin_chats.sql:11 `unique (user_id, client_document_id)` confirms per-user scoping.

</details>

word-addin/src/taskpane/lib/localWordChats.ts
```diff
@@ -0,0 +1,191 @@
...
+    const chatMessages = (await requestResult(
+      messages.index("chat_id").getAll(IDBKeyRange.only(args.chatId))
+    )) as LocalMessageRow[];
+    const nextSequence =
+      chatMessages.reduce(
+        (highest, message, index) =>
+          Math.max(highest, message.sequence ?? index),
+        -1
+      ) + 1;
+    chats.put({
```
amal66, line 116: **[Low] Local-mode chats persist unencrypted in IndexedDB after sign-out, partitioned only by a client-derived owner id**
*Security · independently verified*

Local-storage-mode chats (user messages plus assistant events, i.e. potentially privileged legal content) are written to IndexedDB with no expiry and no deletion path — the module exposes save/list/get but no delete, and signOut (word-addin/src/taskpane/auth/session.ts:358-362) clears only the token keys in OfficeRuntime.storage. The owner_id partition comes from an unverified client-side JWT decode (App.tsx:21-39), which is fine as UX segregation but is not a security boundary: anyone with access to the same OS profile's webview storage (or any script running in the pane) can read every account's local chats. For a mode users will pick specifically for confidentiality, 'sign out' leaving all chat content readable on disk indefinitely is a gap users won't expect.

**Suggested fix:** Add a 'delete local chats' affordance (and consider offering it during sign-out), and state in the storage-mode UI/README that local chats remain on this device for anyone using this Windows/macOS profile until deleted.

<details><summary>Evidence</summary>

localWordChats.ts defines only saveLocalWordMessage/listLocalWordChats/getLocalWordChat (no delete/clear); session.ts clearSession() removes only ACCESS_KEY/REFRESH_KEY; App.tsx:21-39 derives owner id by base64-decoding the JWT payload client-side.

</details>

word-addin/src/taskpane/lib/wordChatSettings.ts
```diff
@@ -0,0 +1,40 @@
+import { useCallback, useEffect, useState } from "react";
+
+export type WordChatStorageMode = "cloud" | "local";
+
+const STORAGE_KEY = "mike_word_chat_storage_mode";
```
amal66, line 5: **[Low] Chat storage-mode preference is device-global, not per account**
*Security · independently verified*

The cloud/local preference is stored under a single OfficeRuntime.storage key ('mike_word_chat_storage_mode') with no user scoping. On a machine where two Mike accounts are used, account B inherits account A's choice: if A chose 'local' for confidentiality reasons, B's chats silently skip cloud persistence (data-retention surprise); conversely if B switches back to 'cloud', A's privacy-motivated preference is silently flipped and A's next chats are persisted server-side without A re-consenting.

**Suggested fix:** Key the preference by the authenticated user's id (the same wordChatOwnerId already computed in App.tsx), e.g. `mike_word_chat_storage_mode:${ownerId}`.

<details><summary>Evidence</summary>

wordChatSettings.ts:5 `const STORAGE_KEY = "mike_word_chat_storage_mode";` read/written without any user identifier, while local chat data itself IS user-scoped via owner_id in localWordChats.ts.

</details>

word-addin/src/taskpane/hooks/useSelectedModel.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const STORAGE_KEY = "mike.selectedModel";
+// Substituted at bundle time; a `typeof process` guard is false in the browser
+// and would silently pin the fallback instead of the configured model.
+const DEFAULT_MODEL =
+  process.env.REACT_APP_DEFAULT_MODEL || "claude-sonnet-4-6";
+
+function readStoredModel(): string {
+  try {
+    return window.localStorage.getItem(STORAGE_KEY) || DEFAULT_MODEL;
```
amal66: **[Low] useSelectedModel skips the web hook's stored-value validation and the two surfaces default to different models**
*UI alignment · independently verified*

The web's useSelectedModel validates the persisted id against ALLOWED_MODEL_IDS (plus ollama/ prefix) and falls back to the default when stale; the add-in returns whatever string is in localStorage, so a model retired from the catalog keeps being sent to the backend and the toggle renders the fallback label 'Model' with no way to know why. Additionally the add-in's default is REACT_APP_DEFAULT_MODEL = 'claude-sonnet-4-6' (webpack.config.js EnvironmentPlugin) while the web's DEFAULT_MODEL_ID is 'gemini-3-flash-preview' — same user, different default per surface.

**Suggested fix:** Validate the stored id against the add-in's model catalog (STATIC_MODELS + ollama/ prefix) on read, and either align the default with the web or document the intentional divergence.

<details><summary>Evidence</summary>

frontend/src/app/hooks/useSelectedModel.ts: isAllowed() check on read and write; word-addin hook readStoredModel() returns localStorage value unvalidated; webpack.config.js sets REACT_APP_DEFAULT_MODEL: "claude-sonnet-4-6".

</details>

word-addin/src/taskpane/components/primitives/Modal.tsx
```diff
@@ -0,0 +1,120 @@
...
+  return createPortal(
+    <div
+      className="fixed inset-0 z-[200] flex items-center justify-center bg-white/10 px-3 py-4 backdrop-blur-[2px]"
+      onMouseDown={(event) => {
+        if (event.target === event.currentTarget) onClose();
+      }}
+    >
+      <section
+        role="dialog"
+        aria-modal="true"
```
amal66, line 105: **[Low] Modal declares aria-modal="true" but implements no focus trap or initial focus, and quick-action ids fork the web's under the same storage key**
*UI alignment · independently verified*

Two smaller a11y/consistency gaps. (1) Modal.tsx sets role=dialog aria-modal=true (an improvement over the web Modal, which has neither) but never moves focus into the dialog nor traps Tab, so keyboard focus stays on the trigger behind the overlay and Tab walks the obscured page — aria-modal asserts inertness that isn't implemented (the web Modal shares the no-trap flaw, so this is parity-with-a-misleading-attribute rather than a regression). Similarly ChatHistoryDropdown.tsx:54 places a search textbox and plain <button> rows inside Radix DropdownMenu content, so the popup announces as a menu whose rows aren't menu items and arrow-key navigation does nothing. (2) lib/quickActions.ts:68 reuses the web's exact storage key 'mike.quickActions.visible' but with kebab-case ids ('compare-documents') where the web uses camelCase ('compareDocuments') — harmless today because origins differ, but a booby trap for any future shared-origin deployment or server-synced preferences.

**Suggested fix:** Add initial focus + a simple Tab loop to Modal (or drop aria-modal until trapped); consider a Popover instead of DropdownMenu for the history list; align quick-action ids with the web's casing or change the storage key.

<details><summary>Evidence</summary>

Modal.tsx has no focus management code (only an Escape keydown listener); ChatHistoryDropdown renders <input> + ChatHistoryListView buttons inside DropdownContent (a Radix menu); quickActions.ts QuickActionId union vs frontend quickActionsPreferences.ts QuickActionId union use different casing under an identical STORAGE_KEY.

</details>

word-addin/src/taskpane/hooks/useWordDoc.ts
```diff
@@ -1,280 +1,1083 @@
...
+            }
 
-        return report;
-      } finally {
-        doc.changeTrackingMode = originalMode;
-        await context.sync();
-      }
-    });
+            try {
+              const matches = doc.body.search(original, { matchCase: true });
```
amal66: **[Low] Caret control codes in ORIGINAL are not guarded before body.search**
*Office.js · independently verified*

applyTrackedEdits guards empty strings, embedded newlines, and the 255-char limit, but passes the search string through otherwise. Word's non-wildcard search interprets caret sequences ('^?' any character, '^#' any digit, '^t' tab, etc.) and rejects some outright ('^p' throws InvalidArgument). If the model copies an ORIGINAL containing a literal caret sequence (e.g. '^2' in a formula-bearing document, or '^?'), the search can match different text than the literal original, and because the deleted text then differs from edit.original the edit lands as 'unexpected-revisions'/mis-targeted rather than being cleanly skipped; '^p'-style strings reject the whole batch as a bare exception. The single-exact-match requirement limits the blast radius but does not eliminate a wrong-range tracked replacement.

**Suggested fix:** Treat originals containing '^' as unsearchable (like multi-paragraph originals), or escape carets ('^' -> '^^') before searching; alternatively verify the matched range's text equals the original before replacing.

<details><summary>Evidence</summary>

useWordDoc.ts:826-836 only checks `!original || original.includes("\n") || original.includes("\r") || original.length > MAX_SEARCH_CHARS`; line 839 `doc.body.search(original, { matchCase: true })` with no wildcard/caret handling.

</details>

word-addin/src/taskpane/hooks/useWordDoc.ts
```diff
@@ -1,280 +1,1083 @@
...
+        ? {
+            error: `The edit was resolved, but anchor cleanup failed: ${cleanupErrors.join(" ")}`,
           }
+        : {}),
+    };
+  } catch (error) {
+    // Office batches can fail after executing earlier queued commands. Do not
+    // offer a blind retry against potentially stale sibling proxies; hand any
+    // remaining revisions back to Word's Review tab.
+    pendingTrackedEdits.delete(handle);
```
amal66, line 433: **[Low] Any resolve failure permanently evicts the in-memory handle, downgrading transient host errors to 'released' for the rest of the session**
*Office.js · independently verified*

In resolveTrackedEditNow's catch, the handle is unconditionally deleted from pendingTrackedEdits and recorded as terminal 'released' — even when the failure was transient (document temporarily locked by co-authoring save, modal dialog, Word Online hiccup). The revision and its hidden bookmark are still intact, but the user loses Accept/Reject on the card until the task pane reloads (which restores from the bookmark). The comment explains the rationale (partially-executed batches make blind retry unsafe), but a bookmark-based re-restore is available in-process and is exactly what a reload would do.

**Suggested fix:** When entry.stableEditId/bookmarkName exist, attempt an in-place restoreTrackedEdit from the bookmark after a failed resolution (mirroring the reload path) before declaring the handle released, or surface a Retry that goes through the restore path.

<details><summary>Evidence</summary>

useWordDoc.ts:418-431: `pendingTrackedEdits.delete(handle); rememberTerminalState(handle, "released");` runs for every error class before classifying it; useWordTrackedEdits.ts:387-392 then deletes the key's handle and shows a terminal error state with no retry affordance.

</details>

word-addin/src/taskpane/components/assistant/ChatView.tsx
```diff
@@ -0,0 +1,377 @@
...
+            setMessagesVisible(true);
+            return;
+        }
+
+        const timer = window.setTimeout(() => {
+            const container = messagesContainerRef.current;
+            const element = latestUserMessageRef.current;
+            if (container && element) {
+                const requestedTop = element.offsetTop - 24;
+                console.log("This is not supposed to run");
```
amal66: **[Low] Debug console.log leftovers in the committed PR, including full user message content and a 'This is not supposed to run' marker**
*Chat runtime · independently verified — independently flagged by the UI alignment, Security analysts too*

Committed HEAD ships unconditional debug logging: ChatView.tsx:130-133 and :141-145 log the latest user message's full content to the webview console on every turn, ChatView.tsx:259 logs the literal string "This is not supposed to run" inside the reveal-scroll effect, and useWordAssistantChat.ts:114-117 logs every send with generation/sendToken. The codebase already has a gated devLog pattern (api/client.ts:23-26) these bypass. Logging user chat content unconditionally in a production legal-document add-in is a data-hygiene problem, and the line-259 marker is an obvious leftover from debugging the scroll behavior (a working-tree edit is already removing some of these, confirming they were not meant to ship).

**Suggested fix:** Remove the logs (or route them through the isDev-gated devLog helper) and update the e2e assertion that depends on the console output.

<details><summary>Evidence</summary>

git show HEAD:word-addin/src/taskpane/components/assistant/ChatView.tsx lines 130, 141, 259 contain the console.log calls (line 130/141 include `content: latestUserMessageDetails.content`); useWordAssistantChat.ts:114 `console.log("[WordAssistantChat] Starting send", ...)`. An e2e test even asserts on the logged content (chat.spec.ts:358 `scrollLogDetails?.content`).

</details>

word-addin/src/taskpane/hooks/useWordAssistantChat.ts
```diff
@@ -0,0 +1,466 @@
...
+            });
+          } else if (wordChatStorage === "cloud") {
+            notifyWordChatHistoryChanged();
+          }
+        } catch (error) {
+          const sessionIsCurrent =
+            mountedRef.current &&
+            generation === sessionGenerationRef.current &&
+            sendToken === sendSequenceRef.current;
+          if (controller.signal.aborted) {
```
amal66, line 383: **[Low] Cancelled cloud streams never refresh chat history, so a server-persisted partial turn (and a newly created chat) stays invisible**
*Chat runtime · independently verified*

The backend persists the user message and the partial assistant response when a stream is aborted (wordChat.ts: AssistantStreamError abort branch calls updateAssistantMessage with the partial events), and on a first turn it has already created the chat row. The client's abort branch (useWordAssistantChat.ts:344-373) saves locally for local mode (which fires notifyWordChatHistoryChanged internally) but never calls notifyWordChatHistoryChanged for cloud mode — unlike the success path (:336-338) and the error path (:401-406). After cancelling a first-turn cloud send, the chat exists server-side with content, but the history dropdown/page won't list it until some unrelated history event fires. Additionally the server abort path never sets the chat title (title update only runs on the success path), so the chat surfaces later as untitled.

**Suggested fix:** In the abort branch, call notifyWordChatHistoryChanged() when wordChatStorage === "cloud"; on the backend, set the title (and updated_at) in the abort path too when the chat is new.

<details><summary>Evidence</summary>

useWordAssistantChat.ts:344-373: the `controller.signal.aborted` branch returns without any cloud-mode notifyWordChatHistoryChanged; backend/src/routes/wordChat.ts abort branch persists partial events via updateAssistantMessage but skips the `word_chats` title/updated_at update.

</details>

word-addin/src/taskpane/lib/localWordChats.ts
```diff
@@ -0,0 +1,191 @@
...
+      transaction
+        .objectStore(MESSAGE_STORE)
+        .index("chat_id")
+        .getAll(IDBKeyRange.only(chatId))
+    )) as LocalMessageRow[];
+    await transactionDone(transaction);
+    return {
+      chat,
+      messages: rows
+        .sort((left, right) => {
```
amal66, line 189: **[Low] Local chat message ordering uses wall-clock created_at as the primary key, with the purpose-built sequence only as a tiebreaker**
*Chat runtime · independently verified*

getLocalWordChat sorts by created_at string compare first and only consults the per-chat monotonic `sequence` when timestamps are exactly equal. sequence was added precisely to encode append order, but any backwards clock adjustment (NTP correction, manual change) between two saves reorders the conversation on reload — e.g. an assistant reply sorted before its own user message. Since every save recomputes sequence from the store under the same transaction, sequence alone is a strictly safer primary key; created_at should be the tiebreaker (or dropped from ordering).

**Suggested fix:** Sort by `sequence` first (with the existing role-based fallback for legacy rows lacking it), using created_at/id only as tiebreakers.

<details><summary>Evidence</summary>

localWordChats.ts:176-181: `const timestampOrder = left.created_at.localeCompare(right.created_at); if (timestampOrder !== 0) return timestampOrder;` before the sequence comparison, while :97-102 computes a correct monotonic sequence per chat.

</details>

word-addin/src/taskpane/components/documents/AddDocumentsModal.tsx
```diff
@@ -0,0 +1,588 @@
...
+  open: boolean;
+  onClose: () => void;
+  onSelect: (documents: Document[]) => void;
+  initialSelectedDocuments?: Document[];
+}
+
+const TABS: { value: DirectoryTab; label: string }[] = [
+  { value: "files", label: "Files" },
+  { value: "templates", label: "Templates" },
+  { value: "projects", label: "Projects" },
```
amal66, line 43: **[Low] Projects and Templates tabs of the document-source modal are untested; deleted projects.spec.ts not fully replaced**
*Tests & tooling · independently verified*

projects.spec.ts (7 tests: project list, empty state, load error, per-project documents, upload flows) was deleted. Its successor UI — the AddDocumentsModal with files/templates/projects tabs, including project expansion and attaching project documents — is only covered on the Files/upload path (chat.spec.ts lines 939-1060). No spec touches the Projects or Templates tabs (grep for "Templates" in e2e/ has zero hits; "Projects" only appears asserting a menu item's absence at chat.spec.ts:124), so project listing/expansion/attachment errors would ship unnoticed.

**Suggested fix:** Add one spec per tab: Projects tab lists mocked projects, expands one to show its documents, attaches one and asserts the outgoing message's files payload; Templates tab equivalent; plus a load-error case mirroring the deleted spec's coverage.

<details><summary>Evidence</summary>

AddDocumentsModal.tsx lines 31-43 define the three DirectoryTab values and lines 174+ implement toggleProject; ripgrep across word-addin/e2e/*.spec.ts finds no interaction with either tab.

</details>

backend/tsconfig.json
```diff
@@ -1,9 +1,8 @@
 {
   "compilerOptions": {
     "target": "ES2022",
-    "module": "CommonJS",
-    "moduleResolution": "node",
-    "ignoreDeprecations": "5.0",
+    "module": "Node16",
```
amal66, line 4: **[Nit] Backend-wide module system switch (CommonJS → Node16) bundled into a feature PR**
*Backend · independently verified*

The PR flips the whole backend from module CommonJS to Node16 resolution, forcing collateral edits (`import("../systemWorkflows.js")` extension in contextBuilders.ts:663, dynamic `await import("@google/genai")` in gemini.ts). tsc --noEmit passes (verified), but this changes emit semantics for every file in the backend and is unrelated to Word chat; if the add-in feature is ever reverted, the build-system change goes with it.

**Suggested fix:** Land the module-resolution migration as its own commit/PR with its own rationale, or at minimum call it out prominently in the PR description.

<details><summary>Evidence</summary>

backend/tsconfig.json diff: `"module": "Node16", "moduleResolution": "Node16"` replacing CommonJS/node; gemini.ts converted client() to async solely to use dynamic import.

</details>

word-addin/src/shared/chat/ChatInput.tsx
```diff
@@ -0,0 +1,150 @@
...
+    isLoading = false,
+    onCancel,
+    placeholder = "Ask Mike…",
+    disabled = false,
+    leftSlot,
+    rightSlot,
+    attachments,
+    className,
+}: ChatInputProps) {
+    const textareaRef = React.useRef<HTMLTextAreaElement>(null);
```
amal66, line 46: **[Nit] Minor copy and affordance divergences from the web composer/initial view**
*UI alignment · independently verified*

Placeholder is 'Ask Mike…' vs the web's 'How can I help?'; the greeting reads 'Hello, {name}' vs the web's 'Hi, {username}' and falls back to 'there' instead of the email prefix; the add-in ModelToggle omits the Check glyph the web renders next to the selected available model (relying on the gray selected-row background instead); and the web's workflow slash-commands (typing '/', with combobox aria wiring) have no counterpart in the pane composer.

**Suggested fix:** Align the two copy strings, add the selected-item Check to the add-in DropdownItem, and note slash commands as a known scope cut (or port them, since WorkflowModal already loads the workflow list).

<details><summary>Evidence</summary>

shared/chat/ChatInput.tsx:46 placeholder default; InitialView.tsx:89 'Hello, {name}' vs frontend InitialView.tsx:179 'Hi, {username}'; frontend ModelToggle.tsx:131-133 renders Check for the selected item, add-in ModelToggle.tsx renders none; frontend ChatInput.tsx:26-34 slash-command imports have no add-in equivalent.

</details>

word-addin/src/taskpane/hooks/useWordDoc.ts
```diff
@@ -1,280 +1,1083 @@
...
+    return results;
+  });
+}
+
+/**
+ * Hook exposing document read/write helpers that wrap the Word JS API.
+ * All functions return Promises and must be called in a component context
+ * where Office.js has already initialised (i.e. inside Office.onReady).
+ */
+export function useWordDoc() {
```
amal66, line 1006: **[Nit] useWordDoc returns new function identities every render, defeating downstream useCallback memoization**
*Office.js · independently verified*

readDocumentText and applyTrackedEdits are re-created on each render (no useCallback/useMemo). useWordTrackedEdits.applyStreamedEdit lists applyTrackedEdits as a dependency (useWordTrackedEdits.ts:244), so applyStreamedEdit and processLiveRedlines get new identities every render as well. No correctness impact today because stream callbacks capture them at send time, but it silently disables the memoization the consumers wrote and will bite if these ever land in an effect dependency list.

**Suggested fix:** Hoist readDocumentText/applyTrackedEdits to module scope (they use no hook state) or wrap them in useCallback with empty deps.

<details><summary>Evidence</summary>

useWordDoc.ts:758-1082 — plain const arrow functions returned from the hook body; both are stateless wrappers over module-level functions.

</details>

word-addin/src/taskpane/lib/networkError.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+  })();
+  return (
+    `Couldn’t reach ${request.method} ${request.url} — ${detail}. ` +
+    `Check that the server at ${origin} is running and reachable from Word.`
+  );
+}
+
+/** True when a thrown value is a transport failure rather than an HTTP status. */
+export function isNetworkFailure(error: unknown): boolean {
```
amal66: **[Nit] isNetworkFailure is dead code and would misclassify the client's own wrapped transport errors**
*Chat runtime · independently verified*

isNetworkFailure tests `error instanceof TypeError`, but api/client.ts sendRequest (lines 73-88) re-wraps every transport failure into a plain `Error` (with the TypeError as `cause`), so any future caller passing an error thrown by the API client would get `false` for genuine network failures. No code currently calls it (grep finds zero callers outside the module), making it a misleading exported API.

**Suggested fix:** Delete the unused export, or make it cause-chain aware (walk `error.cause` looking for TypeError) if it is meant to be used.

<details><summary>Evidence</summary>

networkError.ts:49-51 `return error instanceof TypeError;`; client.ts:79-87 `throw new Error(describeNetworkFailure(...), { cause: error })` replaces the TypeError before callers see it.

</details>

word-addin/e2e/chat.spec.ts
```diff
@@ -225,82 +895,920 @@ test("'Suggest tracked edits' forces document context and sends the format contr
...
 
-  await page.getByRole("button", { name: "Apply 2 tracked edits" }).click();
+  await expect(modal.getByText("agreement.pdf")).toBeVisible();
+  await expect(modal.getByText("Date", { exact: true })).toBeVisible();
+  await expect(modal.getByText("Size", { exact: true })).toBeVisible();
+  await expect(modal.getByText("12 B", { exact: true })).toBeVisible();
+  await expect(modal.locator('img[src*="/icons/pdf."]')).toBeVisible();
+  const uploadedRow = modal.getByRole("button", { name: /agreement\.pdf/ });
+  await expect(uploadedRow).toHaveCSS(
+    "background-color",
```
amal66: **[Nit] Assertions pin exact oklch computed colors, mirroring Tailwind palette constants**
*Tests & tooling · independently verified*

17 toHaveCSS assertions in chat.spec.ts check literal resolved colors such as `oklch(0.928 0.006 264.531)` for selected/unselected document rows. These encode Tailwind's current palette math rather than behavior: any design-token or Tailwind version bump fails the suite with no functional change, and the assertion says nothing about what the color means.

**Suggested fix:** Assert selection state via aria-selected/aria-pressed or a data-state attribute, and keep at most one CSS assertion where the visual itself is the contract.

<details><summary>Evidence</summary>

chat.spec.ts lines 975-986: three consecutive toHaveCSS("background-color", "oklch(...)") calls toggling a row's selection.

</details>

word-addin/package.json
```diff
@@ -2,38 +2,48 @@
...
   "private": true,
+  "engines": {
+    "node": ">=22"
+  },
   "scripts": {
-    "dev": "webpack serve --mode development",
-    "typecheck": "tsc --noEmit",
-    "build": "webpack --mode production && node scripts/build-manifest.js",
+    "predev": "node scripts/clear-sideload.js",
+    "dev": "office-addin-debugging start manifest.xml --dev-server \"bun run dev:server\" --dev-server-port 3000",
```
amal66: **[Nit] dev script hardcodes bun; loadEnvFile needs newer Node than errors will explain**
*Tests & tooling · independently verified*

`"dev": "office-addin-debugging start ... --dev-server \"bun run dev:server\""` requires bun on PATH while every sibling script (including the otherwise-identical `start`) uses npm — a bun-less contributor's `npm run dev` fails inside office-addin-debugging with an opaque spawn error. Relatedly, webpack.config.js:17 calls process.loadEnvFile (Node >=20.12); engines declares >=22 but nothing enforces it, so older Nodes die with `process.loadEnvFile is not a function` instead of a version message.

**Suggested fix:** Make dev use `npm run dev:server` (or detect bun), and guard loadEnvFile with a typeof check that prints the required Node version.

<details><summary>Evidence</summary>

package.json scripts block: dev uses `bun run dev:server`, start uses `npm run dev:server`; webpack.config.js line 17 `process.loadEnvFile(localEnvPath)`.

</details>

willchen96:
Follow-up to the four unanchored findings in the review body (implemented in `a76ca56` unless noted):

1. **Production CORS:** backend allowlisting now combines `FRONTEND_URL`, optional `WORD_ADDIN_URL`, and comma-separated `ALLOWED_ORIGINS`. The variables are documented in `backend/.env.example` and the add-in production README, with integration coverage for the configured origins.
2. **Hermetic E2E network behavior:** the shared fixture now installs a catch-all guard for API/Supabase traffic so an unmocked request fails deterministically instead of escaping to a developer service. The local static server and the intentional Office CDN fallback remain allowed.
3. **Office mock fidelity:** the mock now exposes inserted paragraph marks as `\r`, which exercises the real Word representation used by multi-paragraph revision verification. The README also requires a manual desktop Word pass for multi-paragraph tracked replacements and Accept/Reject before release. Full property `load()`/`sync()` enforcement remains a future mock-hardening item rather than being claimed as complete here.
4. **Fixed-delay tests:** streaming tests now use explicitly released route gates, and layout/scroll assertions advance controlled stream frames and poll for positional stability. The affected fixed `holdMs` and timing sleeps were removed.

Verification for the review-fix commit: add-in typecheck passed, backend build passed, 35 focused backend route/CORS tests passed, the full backend suite passed (555 tests), and the full add-in browser suite passed (112 tests).

amal66:
**Flagging one piece of tech debt we're accepting in this PR: the shared UI is now hand-forked from the web app.**

With `src/vendor/` deleted, the add-in carries its own copies of 10+ web-app components — `shared/chat/mike-icon.tsx` is byte-identical (320/320 lines) to `frontend/src/app/components/chat/mike-icon.tsx`, and `input`, `toggle-switch`, `tab-pill-button`, `PillButton`, `ChatInput`, `Markdown`, `UserMessage`, and `messageStyles` all carry "duplicated from the web app" comments. There is no shared package or webpack alias on the branch, so the two copies will drift apart silently.

We're not blocking this PR on it. The plan is a follow-up that extracts the import-clean leaf components (mike-icon, toggle-switch, tab-pill-button, input, PillButton at minimum) into a local shared package both apps consume, drawing on the shared-UI package extraction indexed in #205. Until then, please avoid growing the forked set further — genuinely pane-specific adaptations (ChatView, EditCard) stay local, everything else should wait for the package.

PR #304, opened 2026-08-10
Extend `/docx` downloads to `/file` for arbitrary file types

drewmiller:
## Summary

This extends the existing `/docx` endpoint to allow for download of arbitrary files, not just DOCX. It consolidates the old `/docx` endpoint with a new `/file` endpoint and redirects the former there.

It's just a suggestion to allow other folks to extend it as we have.

## Why / Motivation

Folks implementing in-browser editors and document viewers for the miscellaneous supported formats will want access to the raw file bytes as opposed to the results supplied by `/display`.

## Changes

As above.

## Tradeoffs & risks

The very minor trade-off is that folks who want to avoid the imperceptible redirect from `/docx` to `/file` will need to migrate. Then the `/docx` endpoint could be deprecated in favor of the more general `/file`.

## How verified

We ran the test suite, which was unaffected by the changes.

## Checklist

- [x] Updated docs / env examples if setup, config, or behavior changed.

### DIFF
--- a/backend/src/routes/documents.ts
+++ b/backend/src/routes/documents.ts
@@ -174,6 +174,45 @@ documentsRouter.get("/:documentId/display", requireAuth, async (req, res) => {
   }
 });
 
+// GET /single-documents/:documentId/file
+// Streams active-version source bytes for browser editors. Unlike /display,
+// this never substitutes a generated PDF rendition for an Office document.
+documentsRouter.get("/:documentId/file", requireAuth, async (req, res) => {
+  const userId = res.locals.userId as string;
+  const userEmail = res.locals.userEmail as string | undefined;
+  const { documentId } = req.params;
+  const versionIdParam =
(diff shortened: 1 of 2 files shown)


willchen96:
Thanks @drewmiller, would you mind updating the PR to resolve the merge conflicts? I had some maintainer fixes that I wanted to add if you could enable maintainer edits.

drewmiller:
> I had some maintainer fixes that I wanted to add if you could enable maintainer edits.

@willchen96 Unfortunately since this PR is from an org account, apparently I can't "allow maintainer edits" in the normal way. I would be happy to (1) grant you write access to tritium-legal/mike or (2) turn the edits manually. Just let me know.

willchen96 (approved):

willchen96:
@drewmiller thanks! I'll place my changes in a separate PR.

PR #326, opened 2026-08-12
feat: formalize design system on existing shadcn/Tailwind foundation

### LINKED ISSUE #322: Formalize a design system on top of the existing shadcn/Tailwind foundation
## Problem

The frontend already has the bones of a design system — shadcn/ui (`new-york` style, `frontend/components.json`), Tailwind v4 with CSS-variable tokens (`frontend/src/app/globals.css`: color, sidebar, and chart tokens), and Lucide icons — but it isn't formalized. `frontend/src/app/components/ui/` has only 14 shared primitives (button, input, dropdown-menu, toggle-switch, search-bar, pill-button, etc.) against 144 total component files, so most UI is bespoke per-feature rather than composed from a documented shared set. There's no catalog (no Storybook, no style guide doc) and no written guidance on when to reuse an existing primitive vs. add a new one-off component. Token discipline is actually decent already (only 5 files repo-wide use raw hex/rgb outside `globals.css`), which is a good foundation to formalize rather than a mess to clean up.

## Proposal

Formalize what already exists rather than rebuild it: document the current tokens and primitives, expand the shared primitive set where duplication is highest, and add a lightweight catalog so contributors can find and reuse components instead of re-implementing them.

**Scope note (updated):** the component catalog (Storybook/Ladle) originally listed as item 2 below is split out to a follow-up issue, #323, so this first phase stays docs + consolidation + accessibility only.

### Scope

1. **Document current foundations** — a single `docs/design-system.md` covering:
   - Color tokens and their semantic intent (from `globals.css` — background/surface/sidebar/chart scales already exist, just undocumented)
   - Typography (Inter + EB Garamond, current usage split)
   - Spacing/radius conventions currently in use (audit `frontend/src/app/components` for the de facto scale)
   - When to use `components/ui/*` primitives vs. shadcn's own registry vs. a new one-off
2. ~~Component catalog~~ — deferred to #323.
3. **Audit for consolidation candidates** — identify feature components that are really the same primitive repeated with minor variations (a quick grep-based pass, not a full redesign) and promote the best version into `components/ui/`.
4. **Accessibility baseline** — since there's no current audit, a pass on the 14 existing primitives (focus states, ARIA on `dropdown-menu`/`toggle-switch`, color contrast on the token palette) is worth bundling in, since primitives are exactly where an a11y fix pays off everywhere at once.
5. **Contribution guideline** — one paragraph in `CLAUDE.md`/`CONTRIBUTING.md`: new UI work should check `components/ui/` and the shadcn registry before writing a new component from scratch.

### Explicit non-goals

- Not a visual redesign — this is about documenting and consolidating the existing look, not changing it.
- Not adopting a new component library — shadcn/ui + Tailwind stays; this formalizes usage of what's already chosen.
- Not cataloging all 144 components up front — start with `components/ui/*`, expand later if it proves useful.
- Not setting up Storybook/Ladle in this phase — see #323.

### Why this shape (industry practice)

Design-system best practice for 2026 treats documentation as a first-class deliverable, not an afterthought: usage guidelines (when to use / when not to), live code examples, and the token/typography foundations, all in one browsable place — this is what GitHub Primer, Shopify Polaris, and IBM Carbon get right and what smaller OSS projects usually skip until it's expensive to retrofit. Since Mike already has semantic, purpose-named Tailwind tokens in place, the highest-leverage move is documenting and reusing that investment rather than introducing new tooling or a new visual language.

## Acceptance criteria

- [x] `docs/design-system.md` documents current color/typography/spacing tokens and their intended usage
- [x] ~~Storybook (or equivalent) set up~~ — moved to #323
- [x] At least 2-3 duplicated feature-component patterns consolidated into shared primitives
- [x] Accessibility pass on existing primitives (focus, ARIA, contrast) with any fixes applied
- [x] Contribution note added on reusing primitives before adding new components

## Follow-up

- #323 — component catalog (Storybook or Ladle) for `components/ui/*`

## Sources

- [10 Best Design Systems in 2026](https://webpixels.io/blog/best-design-systems)
- [15 Best Design System Examples in 2026: Real-World Systems Worth Studying](https://www.uxpin.com/studio/blog/best-design-system-examples/)
- [12 Design System Examples Worth Studying in 2026](https://www.themasterly.com/blog/design-system-examples)

bmersereau:
## Summary

Formalizes the design system that already exists rather than changing it: documents the current tokens and primitives, promotes three duplicated patterns into the shared layer, and fixes the accessibility gaps in the primitives. No visual redesign — the only pixel-level changes are the accessibility fixes listed below.

Closes #322

## Scope note: catalog tool deferred

The component catalog (acceptance-criteria item 2, Storybook or Ladle) was **intentionally deferred to a follow-up issue** and is not in this PR. As a result this PR adds **no new dependencies** — `frontend/package.json` and `package-lock.json` are untouched.

## 1. `docs/design-system.md`

Documents the existing foundations, measured off `globals.css` and all ~120 component files rather than assumed:

- **Colour** — the three vocabularies actually in use, and when each applies: the Tailwind grey ramp (~930 usages, the real working palette), the `app-*` surface tokens, and the shadcn semantic tokens (~45 usages, effectively primitives-only). Plus the accent blue, and the five files that legitimately use raw hex.
- **Dark mode is documented as unshipped** — a full `.dark` token set exists but nothing ever adds the class, and `dark:` variants appear only in the three vendored shadcn primitives. Worth writing down so nobody assumes it works.
- **Typography** — Inter/EB Garamond split, when to reach for `font-serif`, and the de facto type scale (`text-xs` at ~200 uses is the workhorse; this is a dense, table-heavy app).
- **Spacing/radius** — the de facto scale by measured frequency, plus the "bigger surface, bigger radius" convention.
- **The glass/"liquid" idiom** — and the rule that layered `shadow-[...]` recipes are never retyped, since that is this codebase's main source of copy-paste drift.
- **Which layer to reach for** — a 5-step ladder from existing primitive → shared constant → shadcn registry → feature-local one-off → new primitive, with an explicit promotion bar (same markup in 3+ places).

## 2. Consolidated primitives

Chosen from a grep-and-read pass over the feature components, not invented. Each was verified as genuinely the same markup before promoting it:

| New primitive | Replaces | Call sites migrated |
| --- | --- | --- |
| `ui/glass-icon-button.tsx` | A character-for-character identical 300-char glass close button | 4 — `Modal`, `DocumentSidePanel`, `TRSidePanel`, `TREditColumnMenu` |
| `ui/empty-state.tsx` | The icon + serif title + description + action body repeated inside `TableEmptyState` | 8 across 5 files — `ProjectAssistantTable`, `ProjectReviewsTable`, `ProjectsOverview` (×2), `WorkflowList` (×3), `WorkflowDetailPage` |
| `ui/check-square.tsx` | The selection square, including its mixed and muted states | 5 — `AddProjectDocsModal`, `QuickActionsModal`, `FileDirectory` (×3) |

Deliberately **not** consolidated, to keep this honest:

- Inline `Loader2` spinners — 43 sites with real size/colour variation at each; a wrapper would trade one line for one line.
- Tag pills — already shared via `tabular/pillUtils.ts`.
- The hand-rolled "Actions" dropdown (5 copies of an outside-click effect + panel). Real duplication and the biggest win, but migrating it to Radix changes focus/Escape/portal behaviour, so it belongs in its own PR rather than riding along with a docs change.
- `ProjectReviewsTable`'s load-error state and the short "no results found" variants — superficially similar but a different heading size and a deliberately smaller state.

Two design decisions worth flagging: `EmptyState` renders no wrapper, so it drops straight into the existing `TableEmptyState` container; and `CheckSquare` is deliberately role-free and ARIA-free, because call sites disagree about whether the checkbox semantics belong on the square or on the row button wrapping it, and adding a role would double up on the three that already declare one.

## 3. Accessibility fixes

All applied, not just reported, and each covered by a test:

- **`cite-button`** — was missing `type="button"` (so it submitted any surrounding form) and had **no accessible name at all** when rendered icon-only. Now labelled only when there is no visible text, so the visible "Cite" stays the accessible name (WCAG 2.5.3).
- **`search-bar`** — the input had no accessible name (a placeholder is not a name, WCAG 4.1.2), and its only focus indicator was a `white/70` → `white/90` shift, which is imperceptible (WCAG 2.4.7).
- **`tab-pill-button`** — the inactive state used `text-gray-400` on a translucent white pill, ~2.5:1, failing WCAG 1.4.3. Now `text-gray-500` at ~4.8:1.
- **`toggle-switch`** — the off state was a white thumb on a `gray-100` track, ~1.07:1, effectively invisible (WCAG 1.4.11). Fixed with `ring-1 ring-gray-300` rather than a border, so the absolutely-positioned thumb doesn't shift by 1px.
- **`dropdown-menu`** — menu items set `outline-hidden` and relied on `focus:bg-accent`, but the `liquid-dropdown` skin overrides that with a near-white hover colour (~1.02:1), leaving keyboard users no focus indicator. Added a background-independent ring. Also, a selected radio item was marked by `bg-gray-100` on white (~1.03:1) and by colour alone — now also carries `font-medium` as a second, non-colour channel (WCAG 1.4.1).
- **`pill-button`** — had no focus-visible ring.
- Verified as already passing: the oklch token palette (`--muted-foreground` is ~4.6:1 on the light background) and `button`/`input`, which already ship upstream shadcn focus and `aria-invalid` handling.

## 4. `CONTRIBUTING.md`

One additive `## Frontend UI` section on checking `components/ui/` and the shadcn registry before hand-rolling a component. Placed in its own section and does not restructure the file, so it should not conflict with #324's separate addition.

## Test plan

- [x] 39 new tests across 9 files in `frontend/src/app/components/ui/` (new primitives + the ARIA/focus/contrast assertions), matching the existing house style: queried by role and accessible name, asserting only the classes that carry meaning.
- [x] Written first and confirmed failing for the right reasons before implementing.
- [x] Full frontend suite: 315 passed. The 4 failures in `src/app/lib/mikeApi.test.ts` (`blob.text is not a function`) are pre-existing and reproduce on an unmodified tree — a local Node 24 `Response.blob()` difference. `CI` and `Code Quality` are green on `main` at this branch's base commit (`1af9231`), and CI runs Node 22, so no issue was filed.
- [x] `npm run test:coverage` (the CI gate) — the `src/app/lib/**` ratchet floor is untouched by this change.
- [x] `npm run lint` with the cache cleared — 0 errors, and no warnings in any file this PR touches.
- [x] `npm run build` — passes with the same placeholder env vars CI uses.
- [x] `tsc --noEmit` clean.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,6 +55,22 @@ Frontend:
 npm run build --prefix frontend
 ```
 
+## Frontend UI
+
+Before writing a new component from scratch, check
+`frontend/src/app/components/ui/` for a primitive that already covers it — most
+new UI should be composed from that shared layer rather than hand-rolled. If
+there's no primitive but the pattern is a standard one (dialog, tooltip,
+popover, select, tabs), take it from the [shadcn/ui](https://ui.shadcn.com)
+registry so we inherit Radix's keyboard and ARIA behaviour instead of
+reimplementing it. Hand-write a one-off only when it's used on a single screen
+and has no interaction semantics worth sharing, and keep it in the feature
+folder. When you find yourself copying the same markup into a third place —
+especially a `shadow-[...]` glass recipe — promote it to `components/ui/` with a
+test and migrate every call site in that PR. Tokens, the de facto spacing and
+type scales, and the accessibility rules for primitives are documented in
+[`docs/design-system.md`](docs/design-system.md).
+
 ## Testing
 
 ```bash
--- a/docs/design-system.md
+++ b/docs/design-system.md
@@ -1,0 +1,337 @@
+# Design System
+
+How Mike's frontend actually looks and how to build with it. This documents the
+existing system — it is not a proposal to change it. Everything below was read
+off `frontend/src/app/globals.css` and measured across
+`frontend/src/app/components/` (~120 component files), so where the de facto
+convention differs from the theoretical one, the de facto one is what's written
+down.
+
+Stack: Next.js App Router, Tailwind v4, shadcn/ui (`new-york`, see
+`frontend/components.json`), Lucide icons.
+
+---
+
+## 1. Colour
+
+There are three colour vocabularies in the codebase. Knowing which one you're
+in is most of the job.
+
+### 1.1 The Tailwind grey ramp — the working palette
+
+~930 usages of `text-gray-*` / `bg-gray-*` / `border-gray-*` across the feature
+components. This, not the semantic token set, is what the app is actually built
+out of.
+
+| Usage | Class |
+| --- | --- |
+| Primary text | `text-gray-900` (headings), `text-gray-700` (body) |
+| Secondary text | `text-gray-600` |
+| Muted / meta text | `text-gray-500` |
+| Placeholder, disabled-ish | `text-gray-400` |
+| Hairlines and borders | `border-gray-100`, `border-gray-300` |
+| Filled control (checkbox, pill) | `bg-gray-900` |
+| Inverted text on filled | `text-white` |
+
+Contrast floors worth remembering, all against a near-white background:
+`gray-400` is ~2.5:1 and **fails** WCAG AA for text; `gray-500` is ~4.8:1 and
+passes. Do not use `text-gray-400` for text a user has to read — it is fine for
+decorative icons and for genuinely disabled controls, which are exempt.
+
+### 1.2 The `app-*` surface tokens — layered chrome
+
+Defined in `globals.css` as plain hex, exposed as Tailwind colours via
+`@theme inline`:
+
+| Token | Value | Meaning |
+| --- | --- | --- |
+| `--app-background` | `#f9fafb` | The page behind everything |
+| `--app-surface` | `#fdfdfe` | A raised panel or table surface |
+| `--app-surface-hover` | `#f9fafb` | Row/item hover |
+| `--app-surface-active` | `#eff0f3` | Row/item selected |
+| `--app-floating` | `#fefefe` | Floating elements above a surface |
+
+Use these through the shared constants in
+`frontend/src/app/components/ui/liquid-surface.ts` rather than hand-writing the
+classes — `APP_SURFACE_HOVER_CLASS`, `APP_SURFACE_ACTIVE_CLASS`,
+`APP_SURFACE_PRESSED_CLASS`, `APP_SURFACE_GROUP_HOVER_CLASS`. That file also
+holds the two composite surface recipes, `LIQUID_TABLE_SURFACE_CLASS` and
+`LIQUID_PANEL_SURFACE_CLASS`.
+
+### 1.3 The shadcn semantic tokens — primitives only
+
+`--background`, `--foreground`, `--card`, `--popover`, `--primary`,
+`--secondary`, `--muted`, `--accent`, `--destructive`, `--border`, `--input`,
+`--ring`, plus `--sidebar-*` and `--chart-1..5`. All oklch, all in `:root` with
+a `.dark` counterpart.
+
+In practice these are used almost exclusively by the vendored shadcn
+primitives (`button.tsx`, `input.tsx`, `dropdown-menu.tsx`) — about 45 usages
+total against the grey ramp's 930. Keep using them inside anything derived from
+shadcn, so that the primitive stays swappable with upstream. Don't try to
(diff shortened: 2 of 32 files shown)


bmersereau:
Closing to redo against current main — several of the components this touched (Modal.tsx, pill-button.tsx, WorkflowList/WorkflowDetailPage) were substantially refactored upstream (new ModalUI and table primitives) since this branch was written. Redoing the consolidation/a11y work against the new primitives in a fresh PR rather than force-merging stale conflicts.

PR #331, opened 2026-08-13
feat(mcp): DingDuff connector support — OAuth redirect discovery and chat case retrieval

duncanmcqueen:
## Summary

Adds support for DingDuff as a user MCP connector, so the assistant can use DingDuff tools for case retrieval, case reading, and case-specific document lookup directly from chat.

Two changes make this work:

**1. OAuth discovery behind redirects** (`backend/src/lib/mcp/client.ts`, `oauth.ts`)

DingDuff's server 302s its well-known OAuth metadata to a path-scoped issuer, and the MCP SDK's own discovery does not follow redirects, so the OAuth flow could not complete.

- `guardedFetch` now follows redirects hop by hop instead of refusing them outright: every 3xx target is re-validated against the existing SSRF guard (HTTPS-only, private-IP/metadata-host blocks, DNS-pinned dispatcher), credentials are stripped on cross-origin hops (per the fetch spec), and redirected POSTs are converted to GET with the body dropped. Undici still never auto-follows (`redirect: "manual"` on every hop), so no redirect can smuggle egress past the guard.
- New `discoverOAuthServerState` plus the SDK's `OAuthClientProvider.discoveryState` hook hand the SDK pre-discovered metadata, so connectors whose well-known documents live behind a redirect complete the flow.

**2. Chat integration** (`backend/src/lib/chat/streaming.ts`)

When DingDuff MCP tools are present on a turn, the built-in CourtListener tools are withheld and the system prompt directs case retrieval at the DingDuff tools, so the two case-law sources are never mixed within a single turn.

## Tests

- New cases in `client.ssrf.test.ts`: redirect following with per-hop re-validation, refusal to follow a redirect to a blocked address, credential stripping on cross-origin redirects, and POST→GET conversion.
- Full backend suite passes (**711 passed, 24 skipped**); `tsc --noEmit` clean.

## Rebased onto current main

Re-verified that all three pieces are still needed — none has landed upstream in the meantime: `guardedFetch` still sets `redirect: "manual"` with no hop-following, `oauth.ts` has no `discoveryState`/`discoverOAuthServerState`, and `streaming.ts` has no DingDuff handling.

One conflict needed resolving beyond the textual merge. `main` has since introduced `includeAskInputs`/`conversationTools`, which declares its own `baseTools`; this branch declared a second one. The DingDuff contribution here is really the `researchTools` gate, and main's line already consumes that, so the duplicate declaration is dropped and the gate feeds main's `conversationTools` list.

## Notes

- No schema changes; no new dependencies.
- The legal-monitors DingDuff connector source is intentionally **not** part of this PR (that feature set isn't upstream).

### DIFF
--- a/backend/src/lib/chat/streaming.ts
+++ b/backend/src/lib/chat/streaming.ts
@@ -45,6 +45,21 @@ import {
 } from "./tools/documentOps";
 import { verifyCitations } from "./verifyCitations";
 
+function isDingDuffMcpTool(tool: OpenAIToolSchema): boolean {
+  const haystack = [
+    tool.function.name,
+    tool.function.description,
+    JSON.stringify(tool.function.parameters),
+  ]
+    .join(" ")
+    .toLowerCase();
+  return (
+    haystack.includes("dingduff") ||
+    haystack.includes("ding-duff") ||
+    haystack.includes("ding duff")
+  );
+}
+
 export type AssistantEvent =
   | { type: "reasoning"; text: string }
   | AskInputsEvent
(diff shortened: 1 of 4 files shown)


duncanmcqueen:
Being a lawyer a bit here (and did accept it), but the CLA says this:

9. Governing Law
This Agreement shall be governed by and construed in accordance with the laws of [YOUR JURISDICTION, e.g., "the State of California, United States"], without regard to its conflict-of-law provisions.

willchen96:
Thanks @duncanmcqueen would be good to have a new PR given that the repo has moved on quite extensively since this PR was submitted. The user can also just turn off courtlistener integration so there is no need for the manual override when dingduff is active.

PR #335, opened 2026-08-13
feat: per-user dark mode

duncanmcqueen:
## Summary

Adds an opt-in dark mode. Users enable it under **Settings > Appearance > Dark Mode**; the preference persists on the user profile and applies app-wide on subsequent loads.

This branch has been rebased onto current upstream `main` and the toggle is integrated with the new `/settings` page and `SettingsSection` / `SettingsToggle` components.

## How it works

- **Storage**: new `dark_mode boolean not null default false` column on `user_profiles`, following the existing per-column preference pattern. `schema.sql` and `20260813_02_user_dark_mode.sql` keep fresh installs and upgraded deployments aligned.
- **API**: `PATCH /user/profile` accepts boolean `darkMode`; serialized profiles expose it. Older databases without the column fall back to light mode until the migration is applied.
- **Frontend**: `UserProfileContext` toggles the `.dark` class and `color-scheme` on the document root. Toggle changes apply immediately while persistence is in flight; failures restore the previous theme.
- **Styling**: the `.dark` palette covers semantic app tokens plus legacy hardcoded Tailwind neutral surfaces (`bg-white/*`, gray text/borders/divides), avoiding page-by-page forks.

## Tests

- Backend route tests cover persistence/serialization and non-boolean rejection.
- `theme.test.ts` covers applying light and dark root state.
- `UserProfileContext.test.tsx` covers dark -> light -> dark switching and rollback after a failed save.

Validation completed (rebased onto current `main`):

- backend full suite: **709 passed, 24 skipped**; `tsc --noEmit` clean
- frontend `tsc --noEmit` clean; `eslint` **0 errors**
- frontend suite: 470 passed; the dark-mode tests specifically — `theme.test.ts` (2) and `UserProfileContext.test.tsx` (2) — all pass, as do the two backend `PATCH /user/profile appearance` cases

## Rebased onto current main

Re-verified the feature has not landed upstream in the meantime: no `dark_mode` column in `schema.sql`, no dark-mode migration, no `darkMode` in `routes/user.ts`, no `lib/theme.ts`. (`globals.css` does contain a `.dark` block, but that is the stock shadcn token palette — the variant is declared and nothing ever toggles it.)

Two things needed reconciling rather than merging:

**Per-column migration fallbacks.** `main` has since added `quick_actions_visible`, another optional profile column using the same deploy-before-migrate pattern, and both branches edited the same fallback. The fallbacks now cascade per column, so a database that already has `quick_actions_visible` but not yet `dark_mode` keeps the former instead of dropping to the shared base select.

**Settings restructure.** `main` reorganised `/settings` and removed the sign-out control this branch's import list still referenced, so only the icons the page actually renders are imported.

The local full frontend suite has four pre-existing Node 26/jsdom `Blob.text()` failures in untouched `mikeApi.test.ts`; upstream CI uses its normal fresh Node 22 environment.

## Deliberate behavior

- Explicit toggle only; no `prefers-color-scheme` detection.
- A returning dark-mode user may briefly see the light palette before the authenticated profile loads.


willchen96:
@duncanmcqueen @dwmcqueen Hi Duncan it seems like you have two Github accounts - you need to sign the CLA using both thanks!

duncanmcqueen:
Ahh!  I lost access to the other account.  Let me see if Github can fix that.

willchen96 (approved):

willchen96:
@duncanmcqueen thanks - another great contribution, let me know if you want to join the discord

duncanmcqueen:
I'd love to!  Submitted an application to join.

PR #336, opened 2026-08-14
fix(pdf): preserve visual layout when extracting PDF text

duncanmcqueen:
## Summary

Preserve visual layout when extracting PDF text. Positioned fragments such as `constitut` + `e` join correctly, ordinary word gaps remain spaces, and interleaved table columns reconstruct into their visual rows.

## What changed

- Reconstruct lines by clustering Y coordinates independently of PDF content-stream order, then sorting each row by X.
- Use measured X gaps to distinguish kerning, ordinary word spacing, and large column gaps.
- Preserve indentation and paragraph breaks, with caps on indentation and column spacing.
- Apply extraction in the shared `backend/src/lib/pdfText.ts` helper used by document reads and text precomputation.
- Add regression coverage for tiny kerning gaps, 3–3.34pt word gaps at 12pt, interleaved columns with and without `hasEOL`, baseline tolerance, indentation, paragraph breaks, page ordering, and unreadable input.

## Why it changed

Blanket space-joining loses line structure and inserts spaces inside words. The original layout implementation also merged ordinary words and could not reunite same-baseline fragments emitted in column order. This revision addresses both review findings and rebases onto upstream `main` at `4df3edd5`.

## Testing performed

- `npm test --prefix backend -- src/lib/pdfText.test.ts src/__tests__/architecture.test.ts src/lib/dbq/__tests__/handlers2.test.ts src/modules/documents/__tests__/documents.textJobs.test.ts` — 39 tests passed.
- `npm run build --prefix backend` — passed.
- `git diff --check origin/main...HEAD` — passed.

Checks used the rebased branch's locked dependencies. Page markers and the empty-string fallback for unreadable buffers are preserved.

### DIFF
--- a/backend/src/lib/chat/tools/documentOps.ts
+++ b/backend/src/lib/chat/tools/documentOps.ts
@@ -54,6 +54,123 @@ export function citationReminder(
   ].join("\n");
 }
 
+type PdfTextItem = {
+  str?: string;
+  transform?: number[];
+  width?: number;
+  height?: number;
+  hasEOL?: boolean;
+};
+
+/**
+ * Rebuild a page's text from positioned pdfjs items, preserving the visual
+ * layout: lines are reconstructed from y-coordinates/hasEOL markers, words
+ * are joined (or split) based on measured x-gaps rather than a blanket
+ * space, paragraph breaks become blank lines, and indentation is kept
+ * relative to the page's left margin.
+ */
+function layoutPageText(items: PdfTextItem[]): string {
+  type Item = {
+    str: string;
+    x: number;
+    y: number;
+    w: number;
+    h: number;
+    eol: boolean;
+  };
+  const clean: Item[] = [];
+  for (const it of items) {
+    const str = it.str ?? "";
+    if (!str) continue;
+    const t = Array.isArray(it.transform) ? it.transform : [];
+    clean.push({
+      str,
+      x: typeof t[4] === "number" ? t[4] : 0,
+      y: typeof t[5] === "number" ? t[5] : 0,
+      w: typeof it.width === "number" ? it.width : 0,
+      h:
+        Math.abs(typeof t[3] === "number" ? t[3] : 0) ||
+        (typeof it.height === "number" ? it.height : 0) ||
+        10,
+      eol: it.hasEOL === true,
+    });
+  }
+  if (!clean.length) return "";
+
+  // Group into lines: pdfjs marks line ends with hasEOL; also break when the
+  // y coordinate jumps (some producers don't set hasEOL reliably).
+  const lines: Item[][] = [];
+  let cur: Item[] = [];
+  let curY: number | null = null;
+  const flush = () => {
+    if (cur.length) {
+      lines.push(cur);
+      cur = [];
+      curY = null;
+    }
+  };
+  for (const it of clean) {
+    if (cur.length && curY !== null) {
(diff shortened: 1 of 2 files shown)


backend/src/lib/chat/tools/documentOps.ts
```diff
@@ -54,6 +54,123 @@ export function citationReminder(
...
+    for (const it of line) {
+      if (prevEnd !== null) {
+        const gap = it.x - prevEnd;
+        // Only insert a space for a real word gap — small kerning gaps
+        // (e.g. "constitut" + "e") are joined without one.
+        if (gap > lineH * 2.5) {
+          // Preserve obvious columns (signature blocks and simple tables)
+          // without exaggerating ordinary word spacing in justified text.
+          s += " ".repeat(Math.min(16, Math.max(4, Math.round(gap / charW))));
+        } else if (gap > lineH * 0.28) {
```
willchen96: **[P1] Preserve ordinary inter-word gaps.** This threshold can classify a normal space as kerning. At 12pt, the cutoff is 3.36pt, while common spaces are about 3–3.34pt; if PDF.js emits separate positioned items without embedded whitespace, “Agreement” + “Terms” becomes “AgreementTerms”. The test currently embeds leading spaces in “ means” and “ any entity”, so it does not exercise this branch. Please add cases for a zero/tiny kerning gap, a normal word gap, and a large column gap, then calibrate the threshold accordingly.

backend/src/lib/chat/tools/documentOps.ts
```diff
@@ -54,6 +54,123 @@ export function citationReminder(
...
+    if (cur.length) {
+      lines.push(cur);
+      cur = [];
+      curY = null;
+    }
+  };
+  for (const it of clean) {
+    if (cur.length && curY !== null) {
+      const lineH = Math.max(...cur.map((c) => c.h));
+      if (Math.abs(it.y - curY) > Math.max(2, lineH * 0.5)) flush();
```
willchen96: **[P2] Group lines independently of source order.** This loop only joins items that are adjacent in PDF.js output, but a PDF content stream is not guaranteed to be in visual reading order. For example, a table emitted as left-top, left-bottom, right-top, right-bottom is flushed into four groups; sorting those groups afterward cannot reunite fragments that share a baseline, so the rows/columns remain split. Please cluster all items by Y tolerance before sorting each line by X, and add a regression test with interleaved positions.

willchen96 (commented):
Thanks for the contribution. Before another review pass, please rebase this branch onto the latest main; GitHub currently reports merge conflicts. Please also sign or recheck the CLA. The CLA status may be associated with an older revision of the branch, so please trigger a recheck after rebasing even if you have already signed it.

duncanmcqueen:
I updated this PR in commit `281da381` to address both comments.

- The code now keeps normal spaces between words. It joins parts of one word when the gap is very small. It keeps large gaps between columns.
- The code now groups text by its vertical position on the page. It then puts each row in order from left to right. This keeps table rows together when the PDF supplies text one column at a time.
- I added tests for small gaps, normal spaces, large column gaps, and text from separate columns.

I also rebased the branch onto the latest `main` and corrected the commit author from `dwmcqueen` to `duncanmcqueen`.

All 39 selected tests passed. The backend build passed. The CLA check passed.

willchen96 (approved):

willchen96:
Thanks @duncanmcqueen

PR #337, opened 2026-08-14
fix(docx): avoid numbering generated letter paragraphs

duncanmcqueen:
## Summary

- make generated DOCX section numbering opt-in through an optional `numberSections` tool argument
- leave letters and other prose documents unnumbered by default
- stop assigning inferred legal numbers to ordinary body paragraphs
- preserve explicit bullets and manually typed numbering
- add DOCX XML regression tests for default and opt-in numbering

## Problem

The DOCX generator always configured legal numbering and inferred a numbering level for every paragraph beneath a section heading. This caused ordinary demand-letter prose to render approximately as:

`1. DEMAND FOR PAYMENT`  
`1.1 We represent the claimant.`  
`(a) Payment is required within ten days.`

A prompt-only fix would be insufficient because the renderer itself applied numbering even when the model did not request it.

## Behavior after this change

Documents are unnumbered by default. `numberSections: true` numbers section headings when explicitly requested or required by a workflow/template, while ordinary prose remains unnumbered. Explicit bullets use Word bullet formatting. Manually typed prefixes such as `1. This reference is intentional.` remain literal text when automatic numbering is off.

The dispatcher and renderer both use `=== true`, so missing, null, string, or otherwise malformed values cannot enable numbering accidentally.

## Tests

New regression tests inspect `word/document.xml` (and `word/numbering.xml` for bullets) inside generated DOCX bytes and cover:

- demand-letter headings and body prose unnumbered by default
- opt-in legal numbering on headings only
- preservation of manually typed numbering when numbering is off
- explicit bullets rendered as bullets rather than legal clauses

Validation completed (rebased onto current `main`):

- `tsc --noEmit` in `backend` — clean
- `npm test` in `backend` — **711 passed, 24 skipped**
- The four new `documentGeneration.test.ts` cases all pass

## Rebased onto current main

Re-verified the bug has not been addressed upstream in the meantime: `documentOps.ts` on `main` still calls `legalNumbering(inferredLevel)` for inferred paragraphs and still configures the numbering set unconditionally, and `numberSections` appears nowhere in `backend/src`.

The rebase was clean. Worth noting explicitly because it touches neighbouring code: this does **not** conflict with the docx text-value work merged in #328 — that PR's `preserves numeric-looking text through an unrelated edit` regression test still passes alongside the new numbering tests.

No database, frontend, provider, PDF reconstruction, or product-module changes are included.

### DIFF
--- a/backend/src/lib/__tests__/documentGeneration.test.ts
+++ b/backend/src/lib/__tests__/documentGeneration.test.ts
@@ -1,0 +1,161 @@
+import JSZip from "jszip";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const { uploadFileMock } = vi.hoisted(() => ({
+  uploadFileMock: vi.fn(),
+}));
+
+vi.mock("../storage", () => ({
+  downloadFile: vi.fn(),
+  generatedDocKey: (userId: string, docId: string, filename: string) =>
+    `generated/${userId}/${docId}/${filename}`,
+  uploadFile: (...args: unknown[]) => uploadFileMock(...args),
+}));
+
+vi.mock("../downloadTokens", () => ({
+  buildDownloadUrl: () => "/download/test-token",
+}));
+
+import { generateDocx } from "../chat/tools/documentOps";
+
+function fakeDb() {
+  return {
+    from(table: string) {
+      const result = { data: null, error: null };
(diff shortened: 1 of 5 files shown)


willchen96 (approved):

willchen96:
Thanks @duncanmcqueen

PR #338, opened 2026-08-14
fix(migrations) make resumable and safe against schema drift

smtp:
## Summary

  Makes the production SQL migration process resumable and safe against schema drift in the existing Supabase database.

  ## Why / Motivation

  Production already contained parts of the document-version schema, while several historical migrations assumed older documents columns still existed. Re-running the migration sequence therefore
  failed on references such as d.filename, d.storage_path, and changed RPC return types.

  ## Changes

  - Added a migration ledger (public.mike_schema_migrations) so each successful migration is recorded and skipped on subsequent deploys.
  - Updated the migration runner to execute only unapplied migration files and reload the PostgREST schema after completion.
  - Guarded document-version backfills that depend on legacy documents metadata columns (filename, storage paths, file type, size, page count).
  - Made affected migrations safe when columns or tables have already been introduced.
  - Updated overview RPC migrations to drop incompatible existing function signatures before recreating them when the returned row shape changes.
  - Improved migration execution so a failed deploy resumes at the failed migration rather than rerunning the entire history.

  ## Tradeoffs & risks

  - The migration ledger treats a migration as complete only after its SQL file succeeds. A failed migration must be fixed and rerun.
  - Dropping and recreating overview RPC functions creates a very short availability gap for those RPC endpoints during deployment.
  - Guarded legacy backfills intentionally skip data-copy work when the old source columns no longer exist; this prevents deployment failure but does not recreate removed legacy data.
  - This does not replace a full schema-diff or formal migration framework; it hardens the current ordered SQL migration workflow.

  ## How verified

  - Ran migrations against the production Supabase connection through the deployment workflow.
  - Confirmed previously applied migrations are recorded and skipped on rerun.
  - Confirmed migration failures now identify the exact SQL file and statement, allowing fixes to resume from that point.
  - Queried pg_stat_activity through psql to confirm the database connection works and no blocking migration lock was present.

  ## Checklist

  - [x] Reviewed git diff and removed unrelated changes.
  - [ ] Updated docs / env examples if setup, config, or behavior changed.
  - [x] No secrets, API keys, real documents, or .env files committed.


willchen96:
Hi @smtp this is a useful idea - could you sign the CLA?

willchen96:
Thanks Carl @smtp for the contribution—the goal of making database migrations resumable is valuable, and using a migration ledger is a sound approach.

I’m closing this PR because the current implementation is not safe to merge:
- The migration ledger and runner changes described in the PR are not included in the diff.
- Several existing migrations are deleted, removing the upgrade path for deployments that have not applied them.
- Previously shipped migrations are modified instead of adding a new corrective migration.
- Some guarded backfills may skip available data before legacy columns are dropped.
- The schema-drift check is currently failing.

We would welcome a new PR that preserves the migration history, introduces the ledger and runner explicitly, documents how existing databases are baselined, and adds new dated migrations for any repairs. Please also sign the CLA if you intend to contribute. 

Thanks again for identifying this problem and proposing a direction.

PR #339, opened 2026-08-15
feat(llm): add model registry and OpenAI-compatible provider foundation

duncanmcqueen:
## Summary

Lets a deployment declare a self-hosted or third-party OpenAI-compatible model
through `MIKE_MODEL_CONFIG_JSON`, without a code change, and makes those models
usable even when the endpoint behind them cannot be trusted to emit structured
tool calls.

Declared models are served by the existing AI SDK provider layer via
`@ai-sdk/openai-compatible`, so they inherit its transport, retries and
streaming rather than introducing a second HTTP client.

## What changed

**Registry**
- `lib/llm/registry.ts` — parses `MIKE_MODEL_CONFIG_JSON` once per process,
  validating entries and dropping malformed ones. Invalid JSON fails loudly at
  startup.
- `lib/llm/models.ts` — `providerForModel` and `resolveModel` consult the
  registry, so a declared id resolves and persists like any catalog model.
- `lib/llm/providers.ts` — an `openai-compatible` branch builds the adapter
  from the declared `baseUrl`, key and `apiModel`.

Only `openai-compatible` endpoints are declarable. The hosted providers are
already covered by the static catalog and the `openrouter/`, `vercel/` and
`opencode-go/` prefixes, so a configured entry for one of them would be a
second, subtly different way to say the same thing.

**Tool-call tolerance**

Self-hosted Qwen, DeepSeek and GLM builds routinely describe tool calls in
prose and wrap reasoning in `<think>` tags. Rather than a bespoke client, this
is a language-model middleware (`wrapLanguageModel`):

- `lib/llm/toolCallParsing.ts` — pure text handling: `<think>` routing, tool
  markup suppression across chunk boundaries, and recovery of tool calls from
  JSON in `<tool_call>` markers, XML-ish `<function=name>` blocks, DeepSeek
  DSML invocations and single-key maps, with `jsonrepair` for malformed
  arguments.
- `lib/llm/localModelMiddleware.ts` — applies that to the AI SDK stream:
  reasoning parts instead of visible text, real tool-call parts instead of
  prose, and a corrected finish reason.

A tolerant model answering a tool-declaring turn is served through the
endpoint's non-streaming path, because these models interleave markup with
prose in a way a partial stream cannot be reassembled from. Tolerance defaults
on for `location: "local"` and is overridable per model.

**Other**
- `routes/tabular.ts` — availability messages understand declared models: a
  local endpoint needs no key, a cloud one is checked against its own
  declaration.
- `docs/configured-models.md` — configuration reference and the tolerance
  rules.

## Why

Mike could only talk to the providers it was compiled to know about. Operators
running a model on their own hardware had no way in, and the endpoints they
actually run are exactly the ones that need the tolerance layer.

## Testing

- `backend`: `tsc --noEmit` clean; `vitest run` — **764 passed, 25 skipped**,
  including 39 new tests across `toolCallParsing`, `localModelMiddleware` and
  `registry`. The middleware tests drive it with fake `doStream`/`doGenerate`
  results, so they exercise the parsing and stream rewriting without mocking
  the network.

13 files changed, 1818 insertions.

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -28,6 +28,7 @@
         "fast-xml-parser": "^5.7.1",
         "helmet": "^8.1.0",
         "html-to-text": "9.0.5",
+        "jsonrepair": "^3.15.0",
         "jszip": "^3.10.1",
         "libreoffice-convert": "^1.6.0",
         "mammoth": "^1.9.0",
@@ -6640,6 +6641,15 @@
       },
       "engines": {
         "node": ">=6"
+      }
+    },
+    "node_modules/jsonrepair": {
+      "version": "3.15.0",
+      "resolved": "https://registry.npmjs.org/jsonrepair/-/jsonrepair-3.15.0.tgz",
+      "integrity": "sha512-wy8OTjwsJwQRnQJkKnMJJ9vcytRdBPAgIF/Hy6+s1dAj42BHMKiyL8JzEieIl3JY7idt8eyHwBWTO8mh/+mtwA==",
+      "license": "ISC",
+      "bin": {
+        "jsonrepair": "bin/cli.js"
       }
     },
     "node_modules/jszip": {
--- a/backend/package.json
+++ b/backend/package.json
@@ -32,6 +32,7 @@
     "fast-xml-parser": "^5.7.1",
     "helmet": "^8.1.0",
     "html-to-text": "9.0.5",
+    "jsonrepair": "^3.15.0",
     "jszip": "^3.10.1",
     "libreoffice-convert": "^1.6.0",
     "mammoth": "^1.9.0",
--- a/backend/src/lib/__tests__/localModelMiddleware.test.ts
+++ b/backend/src/lib/__tests__/localModelMiddleware.test.ts
@@ -1,0 +1,244 @@
+import { describe, expect, it, vi } from "vitest";
+
+import { localModelToleranceMiddleware } from "../llm/localModelMiddleware";
+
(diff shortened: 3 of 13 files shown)


willchen96:
Follow-up changes from review:

- Added an optional `maxTokensField` setting for configured OpenAI-compatible models.
- The default remains `max_tokens`, preserving compatibility with Ollama, LM Studio, vLLM, and existing deployments.
- Models that require the newer OpenAI request shape can set `"maxTokensField": "max_completion_tokens"`; the provider then rewrites the outgoing token-limit field before sending the request.
- Invalid values are rejected while loading `MIKE_MODEL_CONFIG_JSON`.
- Added documentation and focused tests covering both the default request body and the opt-in `max_completion_tokens` request body.
- Configured model shows up in frontend model toggle UI.

Screenshot 2026-09-18 at 7 05 07 PM


Also added some UI changes and fixes related to the ChatInput in this PR for convenience. 
- ChatInput hidden when chat is loading
- Model toggle scroll bar now sits flush to the side.

Example:

```json
{
  "id": "custom-openai",
  "provider": "openai-compatible",
  "location": "cloud",
  "apiModel": "gpt-5.4-mini",
  "baseUrl": "https://api.openai.com/v1",
  "apiKeyEnv": "OPENAI_API_KEY",
  "maxTokensField": "max_completion_tokens"
}
```

Verification completed:

- 22 targeted backend tests passed
- Backend TypeScript checks passed
- Test TypeScript checks passed
- `git diff --check` passed

willchen96 (approved):

PR #340, opened 2026-08-15
feat: add OpenCode Go as a third model router

duncanmcqueen:
## Summary

Adds [OpenCode Go](https://opencode.ai) — opencode.ai's low-cost subscription gateway for open coding models (GLM-5, Kimi K3, Qwen3.8 Max, MiniMax-M3, DeepSeek V4, …) — as a **third model router**, alongside OpenRouter and the Vercel AI Gateway.

> **Rebased.** This PR originally stacked on #339, which introduced a separate model-registry / OpenAI-compatible-provider foundation. `main` has since landed its own router layer (`llm/openrouter.ts` serving both OpenRouter and Vercel, `lib/routerModels.ts`, `user_router_models`, `RouterSettingsSection`). Rather than land a second, competing provider architecture, the branch has been rewritten onto current `main` in that idiom. **It no longer depends on #339** and is a single self-contained commit. It is also rebased onto the #357 design-system refactor, so the new group is declared once in the shared `ModelToggleUI` primitive that both the web app and the Word add-in now render from.

## What's added

**Backend**
- `lib/llm/openrouter.ts` — `opencode-go` joins the `RouterProvider` union: base URL `https://opencode.ai/zen/go/v1` (overridable via `OPENCODE_GO_BASE_URL`, like the other two), its own key lookup and label. OpenRouter's `HTTP-Referer`/`X-Title` attribution headers and its `reasoning: {…}` payload stay OpenRouter-only; OpenCode Go uses `reasoning_effort` like Vercel.
- `lib/llm/models.ts` — `providerForModel` / `resolveModel` accept `opencode-go/<model>`; `openCodeGoModelId` strips the prefix.
- `lib/userApiKeys.ts` + migration `20260820_01` — new `opencode-go` API-key provider (`OPENCODE_API_KEY` env or encrypted per-user storage); `schema.sql` updated to match so the schema-drift check stays green.
- `routes/models.ts` — `GET /models/opencode-go` returns the user's live catalog (key never leaves the server), deduplicated and sorted by label.
- `routes/user.ts` — the profile PATCH accepts `openCodeGoModels`, and the saved selection gates requests through the existing `user_router_models` choke point, so a hand-crafted request still can't spend the operator's env key on an arbitrary gateway model.

**Frontend + Word add-in**
- `lib/mikeApi.ts` — `getOpenCodeGoModels()`, `opencode-go` API-key provider, `openCodeGoModels` on the profile.
- Settings → API Keys gains an OpenCode Go field; Settings → Routers gains an OpenCode Go model picker.
- `ModelToggle` (web) and the add-in's mirrored catalog gain an "OpenCode Go" group, gated on the user's key.

## Design note

A router's **slug, model-id prefix and API-key provider name are one string**. That invariant already held for `openrouter`/`vercel`, so this PR makes the code depend on it explicitly: `routerForModelId`, the selection lookups, the profile payload fields and the title-model fallback are now driven by `ROUTER_SLUGS` rather than hard-coded pairs. A fourth router should be mostly a list entry plus a label.

The one place the routers genuinely differ is **catalog-id shape**: OpenRouter and Vercel publish `vendor/model` pairs, OpenCode Go publishes bare names (`glm-5`). Requiring a slash would have rejected its entire catalog, so id validation is per-router on both sides (`ROUTER_MODEL_ID_RE` in `routes/user.ts`, `ROUTER_MODEL_ID` in `RouterSettingsSection.tsx`) and the shared `resolveModel` regex has a matching branch.

## Validation

Against `main` @ `5b441de`:

- **backend**: `tsc --noEmit` clean; `vitest run` — **699 passed**, 24 skipped (baseline on `main`: 683 passed; +16 new tests covering the adapter, model ids, the catalog route, the per-router id shapes and the selection gate).
- **frontend**: `tsc --noEmit` clean; `eslint` **0 errors**; `vitest run` — **466 passed** (baseline: 461).
- **word-addin**: `typecheck:app` and `typecheck:e2e` clean; the cross-package catalog-parity and composer-availability drift guards extended to the new router and passing.

Three frontend files (`mikeApi`, `useSelectedModel`, `ChatInput.modelSelection`) fail identically on unmodified `main` in my sandbox — `Blob.text is not a function` and `window.localStorage` undefined, a local jsdom/env issue rather than a code change. Baseline is 12 failures, this branch is 14; the two extra are new `useSelectedModel` tests blocked by that same missing `localStorage`, and CI should run them.


willchen96:
@duncanmcqueen thanks this is a good idea. apologies but will you be able to (1) resign the CLA due to the jurisdiction clause update (2) rebase the PR branch on the latest changes? On (2), I have just implemented both the backend routing for model routing services Vercel AI Gateway and OpenRouter and frontend model selection UI flow. As such you can get your agent to reimplement the Opencode GO connection in the same format.

duncanmcqueen:
Happily!  Will do today!

willchen96:
@duncanmcqueen thanks super useful. are you able to remove your defunct acct from this PR so the CLA check passes?

duncanmcqueen:
> @duncanmcqueen thanks super useful. are you able to remove your defunct acct from this PR so the CLA check passes?

Still waiting on reply from Github as the email address tied to that account is no longer active.

duncanmcqueen:
Done.  Also rebased again (you move fast!)

willchen96:
In `da8ed0e` I routed certain models which had Anthropic compatible endpoints to claude.ts. 

Applied some other routing related changes unrelated to Opencode Go in`0f006e2c`:

- Renamed the settings navigation item from **BYOK** to **Bring Your Own Keys**.
- Moved the frontend page from `/settings/api-keys` to `/settings/byok` and updated the missing-key popup, Word add-in banner, and end-to-end coverage to use the new URL. The backend `/user/api-keys` API remains unchanged.
- Updated user-facing router guidance to say **Settings → Bring Your Own Keys → Routers**.
- Removed the repeated per-provider model captions and added one shared explanation beneath the **Routers** heading.

OpenCode Go catalog notes from review:

- Mike fetches the catalog from OpenCode Go’s `/v1/models` endpoint.
- The live response supplies model IDs but no separate display names or pricing fields, so Mike derives readable names from the IDs and cannot display costs from this endpoint.
- A configured OpenCode Go key alone does not add the provider to the model toggle; at least one OpenCode Go model must also be saved under Routers.

Validation completed: frontend ESLint/TypeScript and router settings tests, backend build and targeted tests, and Word add-in typechecking.

willchen96 (approved):

willchen96:
@duncanmcqueen thank you for the contribution!

PR #341, opened 2026-08-15
feat: user-defined LLM committees

duncanmcqueen:
## Summary

Adds user-defined model committees: a committee answers one prompt with several
models at once and has a chair model synthesize their replies into the single
answer the user sees. Committees are built under **Settings > Models**, stored
on the profile (`user_profiles.model_committees`), and selected from a
**Committee** group in the model picker like any other model.

Stacked on #339 — the first commit here is that PR's foundation. Retarget or
rebase once it merges.

## What changed

**Backend**
- `lib/llm/committee.ts` — members run **concurrently** against the ordinary
  provider layer, then the chair synthesizes. Guards against self-chairing,
  self-membership and cycles.
- `lib/userCommittees.ts` — normalization for reads (malformed rows are
  dropped, never thrown, so a bad row cannot lock a user out of Settings) and
  strict validation for writes (max 8 committees, 2–8 members).
- `lib/llm/models.ts` — `missingCommitteeApiKeyModels` / `modelHasApiKey` /
  `resolveUsableModel`: availability resolves recursively through a
  committee's members and chair.
- `lib/routerModels.ts` — `resolveRequestedModel` recognises committee ids and
  gates them on availability, taking committees the caller already loaded so
  no request reads the profile twice.
- `lib/userSettings.ts`, `lib/chat/streaming.ts`, `routes/chat.ts`,
  `routes/tabular.ts` — committees are loaded once with the rest of the
  profile and threaded to dispatch.
- `routes/user.ts` — `modelCommittees` on the profile (select / serialize /
  PATCH) plus `GET /user/models`. Profile reads get their own retry tier, so a
  database without the column keeps every other live column.
- `migrations/20260823_02_user_model_committees.sql` + `schema.sql`.

**Frontend**
- `components/settings/CommitteeSettingsSection.tsx` — create, edit and delete
  committees. Only models the user can actually run are offered as members.
- `ModelToggle` — a Committee group, shown only when every member and the
  chair is usable.
- `mikeApi` / `UserProfileContext` — committee types, profile state, save.

## Design notes

**Availability is all-or-nothing.** A committee is hidden from the picker and
refused at request time unless every member and its chair is usable. Half a
committee is not a usable answer.

**A deleted committee is an explicit error,** not a silent fallback to another
model — quietly answering with a different model misrepresents which model
wrote the response.

**No tool-calling loop.** When the caller supplies tools (the main chat path
always does) they are dropped and the members are told so, rather than being
left to claim document or case-law work they could not do.

**Cost.** Members are separate model calls, so a committee costs roughly the
sum of its members plus the chair. Running them concurrently keeps the latency
close to the slowest member instead of the sum.

## Why

Different models disagree in useful ways on legal questions. A committee makes
that disagreement visible to a synthesizing model rather than making the user
run the same prompt several times by hand.

## Testing

- `backend`: `tsc --noEmit` clean; `vitest run` — **789 passed, 25 skipped**
  (+25 over #339's foundation, covering committee dispatch, concurrency,
  cycle/self-reference guards, and committee validation).
- `frontend`: `tsc --noEmit` clean; `npm run lint` **0 errors**; `vitest run` —
  **564 passed**, 14 failed. All 14 failures are pre-existing on `main`
  (`mikeApi.test.ts` Blob handling, `useSelectedModel`,
  `ChatInput.modelSelection`) and were confirmed against the baseline before
  these changes.
- Migration follows the dated convention with `schema.sql` updated to match.

28 files changed in this commit (1681 insertions); 37 files and 3498 insertions
including #339's foundation.

### DIFF
--- a/backend/migrations/20260823_02_user_model_committees.sql
+++ b/backend/migrations/20260823_02_user_model_committees.sql
@@ -1,0 +1,5 @@
+-- Migration date: 2026-08-23
+-- Per-user model committees, managed from Settings > Models.
+
+alter table public.user_profiles
+  add column if not exists model_committees jsonb not null default '[]'::jsonb;
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -28,6 +28,7 @@
         "fast-xml-parser": "^5.7.1",
         "helmet": "^8.1.0",
         "html-to-text": "9.0.5",
+        "jsonrepair": "^3.15.0",
         "jszip": "^3.10.1",
         "libreoffice-convert": "^1.6.0",
         "mammoth": "^1.9.0",
@@ -6640,6 +6641,15 @@
       },
       "engines": {
         "node": ">=6"
+      }
+    },
+    "node_modules/jsonrepair": {
+      "version": "3.15.0",
+      "resolved": "https://registry.npmjs.org/jsonrepair/-/jsonrepair-3.15.0.tgz",
+      "integrity": "sha512-wy8OTjwsJwQRnQJkKnMJJ9vcytRdBPAgIF/Hy6+s1dAj42BHMKiyL8JzEieIl3JY7idt8eyHwBWTO8mh/+mtwA==",
+      "license": "ISC",
+      "bin": {
+        "jsonrepair": "bin/cli.js"
       }
     },
     "node_modules/jszip": {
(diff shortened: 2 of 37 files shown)


willchen96:
@duncanmcqueen Thanks for this. I think I'll close this PR without merging for now. After thinking about it this feature is probably not super intuitive for the average lawyer at the moment and I would like to avoid adding too many features. In the future I do want to implement some kind of system where you can have multiple side chats (that each can make research tool calls) and deliver responses that can be viewed in parallel.

PR #347, opened 2026-08-16
[Word] Edit cards: reliable resolution on web + Mac, markdown context, and multi-occurrence editing (replace one or all)

amal66:
## What this PR does

This PR evolved in review rounds; each round below carries its own base-case replication, PR replication, and trade-offs.

- **Rounds 1–2** — fix the broken card ↔ tracked-change link on Word for the web, plus a Direct/Review apply-mode toggle.
- **Round 3** — Word for **Mac desktop** support (split tracked authoring, `Word.Revision` resolution) and the composer control rework.
- **Round 4** — the document reaches the model as **structure-annotated markdown** (headings, list labels, pipe tables) without breaking the verbatim-quote contract that edit targeting and citations depend on.
- **Round 5** — **multi-occurrence editing**: unique-`<original>` contract, positional revision identity, list-aware whole-paragraph deletion, and `<occurrence>all</occurrence>` replace-all.

### Rounds 1–2 (original scope)

Two things:

1. **Fixes the broken card ↔ tracked-change link on Word for the web.** Accept / Reject / Accept-all on an edit card now actually resolves the revision in the document. Before this PR those buttons silently did nothing on Word online: the card flipped to *"The revisions in this passage changed after Mike applied the edit"* while every redline stayed pending in the document.
2. **Adds a Direct / Review apply-mode toggle** at the top of the chat header (explicitly requested): **Review** (default) keeps the approval flow — edits land as tracked changes with linked cards — while **Direct** applies each edit and immediately accepts it, so the document just shows the final text and the card becomes informational ("Applied to the document.").

## The bug, replicated live before fixing

Reproduced against **real Word on the web** (sideloaded dev add-in, real backend, real model):

1. Document with two typos in **one paragraph**: `The Suplier shall provide…` / `…within thirty (30) days of reciept…`
2. Ask Mike: *"Fix the spelling mistakes in this contract."* → two tracked changes stream into the document, two pending cards appear. ✅
3. Click **Accept** on card 1 → ❌ nothing happens in Word. Ground truth read via Office.js showed the document still holding all four pending revisions (`Deleted "Suplier"`, `Added "Supplier"`, `Deleted "reciept"`, `Added "receipt"`), and the card showed the "revisions in this passage changed" error. Same for **Reject** on card 2.

## Root cause: Word for the web reports revisions per-range unreliably

Debug probes added during replication showed three distinct host behaviors (each `anchor scan` line below is one retained range's `getTrackedChanges()` at resolve time):

```
[tracked-edit/apply]   revision re-read {"rereadCount":1,"recovered":false}   ← inside the mutating batch
[tracked-edit/resolve] anchor scan {"pendingCount":1,"addedMatches":1,"deletedMatches":0}  ← inserted range
[tracked-edit/resolve] anchor scan {"pendingCount":1,"addedMatches":1,"deletedMatches":0}  ← search match range
[tracked-edit/resolve] anchor scan {"pendingCount":0,"addedMatches":0,"deletedMatches":0}  ← retained paragraph proxy
```

- Inside the `Word.run` batch that queues the replacement, the generated revisions may not be visible at all.
- Afterwards, **every** range the add-in retains — the search match, the inserted-text range, even the containing paragraph — reports only the **Added** revision. The **Deleted** strikethrough run sits just outside all of them.
- A retained (tracked) paragraph proxy can later report **zero** revisions.
- Only `context.document.body.getTrackedChanges()` reliably lists all pending revisions.

The old resolution logic required the full Added+Deleted pair to be visible through **one** range proxy (strict whole-set equality in `trackedChangesMatchEdit`), so on Word online it always refused and the card lost its link.

## How the fix works

Resolution now identifies the edit by **revision content** rather than proxy identity, in three layers (most precise first), in `useWordDoc.ts`:

1. **Exact retained children** — unchanged fast path; still what Word desktop and the mocked host use.
2. **Per-anchor, per-side matching** — `pickEditRevisionSubset()` selects, from whatever revision set a range reports, exactly one `Added` revision whose text equals `toWordText(replacement)` and one `Deleted` revision whose text equals the original (line-break–normalized). A read-only scan pass locates which anchor can supply each side before anything mutates, so a missing side aborts with the document intact.
3. **Atomic document-level fallback** — `resolveEditThroughBody()` finds the edit's unambiguous Added/Deleted pair in the body collection and resolves both sides in a single batch.

Safety invariant at every layer: **exactly-one-per-side**. If the document holds more than one pending revision with the edit's text on either side, nothing is touched and the card reports an honest error — Mike can never resolve a revision that is not provably its own.

The same subset matching (plus the body fallback) applies to `restoreTrackedEdit`/`restoreTrackedEditsNow`, so cards re-link after a task-pane reload; the batched restore adds a consumed-proxy set so two edits can never bind the same revision. Apply-time now also re-reads the edited paragraph once before giving up on exact review controls, and the persistent bookmark spans the whole edited paragraph when per-revision ranges are unavailable (a bookmark covering only the inserted half restores view-only).

### The toggle

- `useWordEditApplyMode()` persists `mike_word_edit_apply_mode` in `OfficeRuntime.storage` (default `approval`).
- Direct mode reuses the identical apply pipeline (search validation, ambiguity guards, TrackAll authoring) and then immediately accepts the revision — so every safety check and failure path is shared between modes. If the auto-accept cannot complete, the card degrades to the unmanaged state instead of pretending the edit was finalized.
- The mode is bound per edit at scheduling time, so flipping the toggle mid-stream never splits one edit's lifecycle.

## Fix verified live (same real-Word scenario)

- **Review:** Accept card 1 → document text becomes final (`The Supplier …`), pending revisions drop 4 → 2, card shows "Accepted." Reject card 2 → original `reciept` restored, revisions → 0, card shows "Rejected."
- **Direct:** *"Change 'thirty (30) days' to 'forty-five (45) days'…"* → document immediately reads `forty-five (45) days` with **no** pending revision; card shows "Applied to the document."
- Pane console shows `[tracked-edit/resolve] body fallback resolved` — the new path doing the work.

Before

Uploading word_before.mov…

After 
https://drive.google.com/file/d/1FuAdkVKh62yXqPpOGRvdTL6YLiGj7ksl/view?usp=sharing
### Hermetic e2e suite (no Word, no backend — ~2 min)

```bash
cd word-addin
npm install
npx playwright test
```

Builds the production bundle and runs the full suite against the in-page Office.js mock (the mock now mirrors the real host's `body.getTrackedChanges()` and `range.paragraphs`). New coverage lives in `e2e/edit-apply-mode.spec.ts`; the linkage contract is in `e2e/chat.spec.ts` ("keeps an edit reviewable through its passage when Word withholds revision proxies") and `e2e/tracked-edit-persistence.spec.ts` ("keeps review controls when the bookmark holds an unrelated sibling revision" / "stays View-only when the bookmark's revisions no longer identify the edit").

### Live against real Word on the web

1. Start your local stack as usual (Supabase compose + backend on `:3001`), then the add-in dev server:
   ```bash
   cd word-addin
   export SUPABASE_PROXY_TARGET=http://127.0.0.1:<gateway-port> API_PROXY_TARGET=http://localhost:3001
   set -a && source .env.development && set +a
   npm run dev:server   # HTTPS dev server on :3200
   ```
2. Open a document in Word on the web → **Add-ins → More Add-ins → My Add-ins → Upload My Add-in** → pick `word-addin/manifest.xml` (`e2e-live/manual-session.mjs` automates the sideload if you have a persistent MS profile).
3. Sign in to the pane with your local dev user. You should see the **Direct / Review** pills at the top of the chat.
4. Type a paragraph containing two known typos **in the same paragraph** (this is the shape that broke), e.g.:
   > SERVICES AGREEMENT. 1. Services. The Suplier shall provide the consulting services described in Schedule A. 2. Payment. All invoices are payable within thirty (30) days of reciept by the Client.
5. **Review mode:** ask *"Fix the spelling mistakes in this contract."* → both fixes appear as redlines + cards. Click **Accept** on one card (text becomes final in the document) and **Reject** on the other (original text restored). Word's Review tab should show no leftover pending revisions afterwards.
6. **Direct mode:** flip the toggle to **Direct**, ask for another concrete change → the document text updates immediately with no pending revision; the card reads "Applied to the document."
7. Optional ground truth while testing: the pane's devtools console (verbose level) logs `[tracked-edit/apply] …` and `[tracked-edit/resolve] …` lines showing which resolution layer acted.
8. Reload the task pane and reopen the chat from history: the cards for still-pending edits should come back actionable (bookmark restore), not "Historical change."

## Commits

- `fix(word-addin): resolve tracked changes by revision content …` — the linkage fix + mock/spec updates (full mechanism write-up in the commit message).
- `feat(word-addin): Direct/Review toggle chooses whether edits need approval` — the toggle, threaded App → header/ChatPanel → tracked-edits hook, with its own spec file.

A third local commit fixing the flaky workflow-metadata spec was dropped during rebase — main already carries an equivalent (stricter) fix.

---

## Round 3: Word for Mac desktop support + composer control rework

Two more commits:

### `fix(word-addin)`: cards now resolve on Word for **Mac desktop**

**Base-case replication (before this commit):** on Word for Mac 16.112, run the same two-typos scenario from above in the desktop app. Both cards fail with *"Word couldn't update this change. Review it directly in Word before retrying."* and the redlines show **no strikethrough** on the old text.

**Root cause (live-probed in the real host, OOXML-verified):** Mac decomposes a tracked `insertText(Replace)` into an **untracked deletion + tracked insertion** — the .docx contains only `w:ins`, never `w:del` — so the Added+Deleted ownership proof can never hold. This is a superset of the Microsoft-acknowledged, still-open [office-js#5188](https://github.com/OfficeDev/office-js/issues/5188). Additionally, retained cross-batch range proxies die with `GeneralException` at `Document._GetObjectByReferenceId`.

**Fix:** replacements are authored as split tracked operations (`insertText(..., After)` + `delete()`), which Mac records as a real revision pair; resolution and card re-linking gain a capability-gated desktop path over the `Word.Revision` surface (WordApiDesktop 1.4), with a two-step ownership proof for Mac's empty-text Delete revisions (sole-Delete rule, then adjacency to the edit's uniquely matched Insert). Web keeps its existing body-level path untouched.

**PR replication:** repeat the desktop scenario on this branch — redlines now show struck-through original + underlined replacement; Accept/Reject/Accept-all from cards finalize/restore the text and drain the Review tab to zero; Direct mode and format edits also verified live on 16.112.

**Trade-offs / design decisions:**
- Split authoring renders a replacement as adjacent Deleted+Added runs rather than one combined replace mark — visually near-identical, and the shape other production redlining add-ins emit (both public Mac bugs were filed against the `Replace` form).
- The empty-text Delete matching accepts a revision it cannot read the text of **only** when uniqueness (sole pending Delete) or adjacency to the verified Insert makes ownership provable; any ambiguity still aborts untouched.

### `refactor(word-addin)`: composer controls

- The Review on/off switch is now a **dropdown** (same primitive as the model picker): plain icon+label trigger ("👁 Review" / "🖊 Edit"), menu titled "How should edits be applied?" with a described option per mode and a check on the active one. Stored preference key/values unchanged.
- The standalone workflow button moved **into the "+" menu** (Desktop Files / Web files / Workflows).
- The model picker is **icon-only** (Settings2); the model name lives in the tooltip and menu.

---

## Round 4: structure-annotated markdown document context

`document_context` is no longer flat `body.text`: `lib/documentMarkdown.ts` renders headings (`#` marks), list items (their real Word `listString` labels + nesting), and tables (pipe tables) **additively**, leaving each passage's own text byte-identical — because edit blocks and citations locate quotes with Word's search API, and any character the renderer invents poisons that search. A stripped-marker retry catches models that quote the annotations anyway, and any host without the structure APIs falls back to flat text. Full mechanism write-ups in commits `418cd4f` / `4fbf3ab`; coverage in `e2e/document-markdown.spec.ts`.

---

### Base-case replication (before these commits)

1. **Replace-all is impossible.** Document containing `Acme Corp` three times → ask *"replace all Acme Corp with Acme Ltd"* → the card reads *"Skipped — source text appears more than once."* while the answer's prose claims the rename happened. The protocol cannot express "all", and the stream parser even drops repeated `<original>` blocks.
2. **Identical corrections break each other (live bug).** Fix `teh Buyer` → `the Buyer` in ¶1, then `hte Buyer` → `the Buyer` in ¶2 (Review mode, both pending). Accept on **either** card errors with *"The revisions in this passage changed after Mike applied the edit."* — two pending Added revisions share the text `the Buyer`, and the text-only body fallback refuses both.
3. **List deletions leave numbered husks.** *"Remove point 2"* on a numbered list empties the paragraph but keeps its paragraph mark, so Word keeps numbering a blank item.
4. **Wrong-occurrence edits are invisible.** Nothing on a card says *where* the change landed, so a unique-but-wrong match sails through review.

### What changed (three commits; each message carries the full mechanism write-up)

- `feat(word-addin): edit-protocol groundwork` — `<occurrence>all</occurrence>` in both stream parsers (no numeric form — models miscount, the index space shifts under tracked deletions, and Word drops overlapping matches, office-js#4992); honest skip statuses (`ambiguous` names the match count, new `unsearchable`/`conflicted` copy); an `In: «paragraph snippet»` line on cards.
- `feat(word-addin): multi-occurrence tracked edits` — resolution narrows candidate revisions by **location** (`compareLocationWith`, read-only; WordApi 1.3-gated) before the exactly-one-per-side rule; replace-all runs the unchanged single-match pipeline once per occurrence, **last match first** (office-js#2800 corrupts later positions on the web), one handle/bookmark per pass (`${key}#${pass}`), aggregated under one card; whole-item deletions escalate to deleting the paragraph mark so Word renumbers.
- `feat(backend): Word edit-protocol contract` — `<original>` must be document-unique ("shortest passage **that is still unique**"), occurrence-targeted requests quote context *from that occurrence* and name the location in `<reason>`, replace-all uses the tag, citations get the same uniqueness rule.

### PR replication

Hermetic (~2.5 min, no Word/backend):

```bash
cd word-addin && npx playwright test
```

292 passed on chromium + webkit. The interesting specs: `e2e/replace-all.spec.ts` (review accept/reject, direct mode, task-pane reload), `chat.spec.ts` › *"edits sharing replacement text resolve independently by location"* (verified to **fail** with the positional narrowing disabled — the regression guard for base case 2), and `e2e/list-item-deletion.spec.ts` (renumbering escalation, over-length honest skip).

Live against real Word on the web (sideloaded dev add-in, real backend, scripted model stub; video + screenshots recorded, ground truth read via Office.js after every step):

1. **Honest skip:** *"Please rename the company"* → a single ambiguous edit for `Acme Corp` (3 occurrences) → card reads *"Skipped — this text appears 3 times in the document. Tell Mike which one to change."*, document untouched (0 revisions).
2. **Replace-all:** *"Replace all Acme Corp with Acme Ltd"* → **one** card, document holds exactly 6 pending revisions (3× `Deleted "Acme Corp"` + 3× `Added "Acme Ltd"`). One click on **Accept** → 0 pending revisions, body text contains `Acme Ltd` ×3 and `Acme Corp` ×0.
3. **Identical corrections (the round-5 live bug):** `teh Buyer` and `hte Buyer` both corrected to `the Buyer`, both pending simultaneously → Accept each card in turn → both resolve cleanly (`the Buyer` ×2 final, 0 revisions). The pane console shows the anchors reporting only the Added halves (Word-web's usual under-reporting) and `[tracked-edit/resolve] body fallback resolved` — i.e. the new location-narrowed document-level path is what did the work. Before these commits, the first Accept errored with *"The revisions in this passage changed…"*.

(Demo video attached in the PR conversation.)

### Trade-offs / design decisions

- **No `<occurrence>N</occurrence>` index.** Three independent reasons, each silently landing a redline on the wrong sentence: models miscount; tracked deletions keep deleted text searchable so the index space shifts mid-stream; Word drops overlapping matches. "all" needs no counting; single targeted edits disambiguate by unique context instead.
- **Replace-all = N single-occurrence passes, not exactly-N guards.** Generalizing the safety guards to "exactly N per side" is unsound — N identical revisions from *other* cards can satisfy a count, letting a replace-all resolve someone else's revisions. Per-occurrence decomposition costs N `Word.run` round-trips and buys every existing exactly-one guard unchanged.
- **One aggregate card per replace-all** (Accept/Reject act on all occurrences together). Per-occurrence control remains available in Word's own Review tab.
- **Unique-context originals widen redlines** — the whole expanded passage renders struck-through even when one word changed. Accepted for now; `<before>/<after>` context anchors are the follow-up if it grates.
- **`matchWholeWord` only for ASCII-word-bounded originals** — the option is broken on Word-web for accented text and @/#-prefixed terms (office-js#985, #3360, #5223).
- **Deploy order matters:** the add-in parser must reach clients **before** the backend prompt starts emitting `<occurrence>` — an older cached client that receives the unknown tag shows a permanently "Incomplete" card. The commit order mirrors this.
- **Pre-merge live checklist** (the mock cannot answer these): does `body.search` match inside pending tracked deletions; is `matches.items` document-ordered; per-occurrence `getTrackedChanges` scoping on the web; Mac office-js#5188 adjacency pairing with N identical pairs; N-bookmark save/reload scale; whether Word coalesces N adjacent insert+delete pairs in one paragraph; real renumbering behavior for the paragraph-deletion escalation on both hosts.


amal66:
## Markdown document context (418cd4f + 4fbf3ab) — before/after

The pane now sends the active document to Mike as **structure-annotated markdown** instead of flat `body.text`, so the model can see the clause hierarchy, lists, and tables it is asked to reason about.

### What the model receives, same document

**Before** (flat `body.text`):
```
Definitions
Terms have the meanings below.
Affiliate means a controlled entity.
Term Meaning Fee See Exhibit B
```

**After** (structure-annotated markdown — passage text byte-identical):
```markdown
# Definitions
Terms have the meanings below.
a. Affiliate means a controlled entity.

| Term | Meaning |
| --- | --- |
| Fee | See Exhibit B |
```

The design constraint: edits (`<original>`) and citations are located with Word's search API, so the model must quote document text **verbatim**. The renderer is therefore additive-only (no inline `**bold**` — ever), the prompt declares the markers as annotations, and if a model quotes one anyway, the apply/citation paths retry the search with the markers stripped (at most one layer, so `# 1. Introduction` with a real typed "1." survives).

### E2E proof — fails before, passes after

New spec `word-addin/e2e/document-markdown.spec.ts` (3 tests × chromium/webkit), run against both sides:

| | pre-change code (5ca3b00 + specs) | this branch (4fbf3ab) |
|---|---|---|
| context carries `#`/list/table structure | ❌ fail | ✅ pass |
| edit quoting `# Definitions` still applies | ❌ fail (card: "Skipped — source text was not found.") | ✅ pass (1 tracked change on `Definitions`) |
| citation quoting a list label still selects | ❌ fail (search never matches) | ✅ pass |
| **summary** | **6 failed** | **6 passed** — full suite **266 passed** |

Full logs: [red-run.txt](https://github.com/Open-Legal-Products/mike/blob/olp-pr/347-evidence-markdown-context/red-run.txt) · [green-run.txt](https://github.com/Open-Legal-Products/mike/blob/olp-pr/347-evidence-markdown-context/green-run.txt)

### Recordings — same model output, both builds

The model proposes `<original># Definitions</original> → Defined Terms` (quoting the renderer's heading marker):

**Before — the edit dies** (`# Definitions` exists nowhere in the document; verbatim search misses):

https://github.com/user-attachments/assets/a4b3d9c9-9dc8-4cbf-9a55-494b77c83398

**After — the stripped-marker retry locates "Definitions" and lands a pending tracked change** (Accept/Reject/View):

https://github.com/user-attachments/assets/ff86ad1a-0616-4c5c-9115-1e9782c89489

WebM originals + final-frame stills on the evidence branch [`olp-pr/347-evidence-markdown-context`](https://github.com/Open-Legal-Products/mike/tree/olp-pr/347-evidence-markdown-context).

### Replicate

```bash
# after: this branch
cd word-addin && npx playwright test document-markdown        # 6 passed

# before: previous branch tip + only the new test files
git worktree add /tmp/before 5ca3b00
git checkout olp-pr/word-addin-edit-approval -- \
  # copy e2e/document-markdown.spec.ts + e2e/support/office-mock.ts into it
cd /tmp/before/word-addin && npm ci && npx playwright test document-markdown  # 6 failed
```

### Tradeoffs (explicit)

- **No inline bold/italic in the context** — emphasis markers sit *inside* passages and would corrupt verbatim quoting; heading/list/table structure was the value, so this is deliberate scope.
- **Multi-line table cells flatten to one line** in the pipe table; such passages are unsearchable for edits anyway (Word's search rejects paragraph breaks).
- **Host fallback**: if any structure API misbehaves, the pane silently sends flat `body.text` (previous behavior) rather than failing the send. The e2e Office mock without `documentBlocks` exercises exactly this path, so the fallback keeps its coverage.
- **Structured path not yet exercised against a real Word host** — the renderer is proven pure + against the extended mock; first live-Word run is the remaining check.

willchen96 (approved):

PR #351, opened 2026-08-18
[Mac app] Self-contained local mode — the whole stack runs inside Mike.app (stacked on #302)

amal66:
**Explainer:** https://claude.ai/code/artifact/57a3c957-0078-40bc-bb02-9c3c32b4386d

## Summary

Stacked on #302 (desktop shell; hosted default). This PR adds the third and final mode from `desktop/docs/self-contained-plan.md`: **Run locally on this Mac**. The app supervises the entire Mike stack — Postgres 17, GoTrue, PostgREST, gateway, backend, frontend — as loopback-only child processes. No Docker, no server, no account anywhere; documents and data never leave the machine.

The shell-over-rewrite principle survives intact: the web app and backend run **byte-identical** to the docker-compose stack. `desktop/src/local/supervisor.js` is `docker-compose.yml` reimplemented as a process tree; where a step looks odd, the compose file is the reference.

> **Rebased onto `main` `aba3cfca` on 2026-09-14**, and re-stacked on #302's new tip `9d36c826`. Tip **`eb30a98d`**.
>
> `main` moved a long way under this branch, so the rebase was not mechanical. Four adaptations are called out under *What changed on the rebase* — in particular, **the fs storage driver had to grow five new branches** to cover storage functions `main` added (raw-file downloads #456/#304, direct upload sessions), without which local mode would have had neither a working download nor a working upload.
>
> The hands-on testing writeup in [this comment](https://github.com/open-legal-products/mike/pull/351#issuecomment-5338378169) is still worth reading, but its commit SHAs died in two rebases. Current equivalents: `cca0596` → **`6426dd6e`**, `62b8314` → **`9599b5bb`**, `5500e84` → **`066e7df9`**.

---

### Backend (works for any single-node deploy, not just the desktop)

- **`STORAGE_DRIVER=fs`** — a filesystem driver inside `lib/storage.ts` behind the same public API as S3/R2. Documents become plain files under `STORAGE_FS_ROOT`; S3 semantics (string-prefix listing, delete-missing-ok) are preserved and unit-tested.
- **Blob-token signed URLs (read)** — fs mode has no S3 presigner, so `getSignedUrl` returns an **expiring HMAC capability URL** on a new unauthenticated route `GET /download/signed/:token`, with exactly presigned-URL trust semantics: minted by an authenticated route after its own access check, scoped to one object, expiring. Domain-separated from the permanent `/download/:token` HMAC by a literal `"blob:"` prefix inside the MAC input, so the two token kinds can't be replayed as each other (tested).
- **Blob-token signed PUT (write) — new on this rebase.** `main` made direct upload sessions the *only* upload path (the legacy endpoint now 410s), and those presign a browser `PUT`. `getSignedUploadUrl` now has an fs branch that mints the write-side twin: `{path, contentType, size, exp}` under a third HMAC domain (`"blob-put:"`), redeemed by `PUT /download/signed/:token`. The body is streamed straight to disk, and a body whose size or content type differs from the signature is rejected and the partial file removed — the same property S3 gets from binding `Content-Length` into the signature.

### Frontend

`output: "standalone"` behind `NEXT_OUTPUT_STANDALONE=1` — opt-in, Docker image and dev workflow untouched. Plus a **Continue as guest** button that only ever renders inside the local desktop product.

### Desktop

- `src/local/supervisor.js` — ordered boot with health-check chain (postgres `:42810` → gotrue `:42811` → schema → postgrest `:42812` → gateway `:42813` → backend `:42814` → frontend `:42815`), all bound `127.0.0.1`; per-service log files under `userData/local/logs/`; clean shutdown on quit (postgres fast shutdown, reverse boot order, 10 s grace then SIGKILL); stale-pidfile recovery that only clears a pid nothing owns; fail-fast if any child dies during boot.
- First-run bootstrap: `initdb` into `userData/local/pgdata`, the Supabase role ladder (incl. `service_role` **with BYPASSRLS** — schema.sql enables RLS with no policies, so the backend's role must bypass, exactly as the supabase/postgres image sets it), GoTrue's own auth migrations, then `schema.sql` + a **migration ledger** (`public.mike_schema_migrations`): fresh installs baseline all shipped migrations; upgrades apply only unseen ones.
- Per-install secrets minted on first run (JWT secret, anon/service keys, DB password, download + API-key-encryption secrets, and the guest credentials) — **no secret ships in the build**, and `secrets.json` is written `0600`. The frontend bundle carries only the well-known Supabase demo anon key as a placeholder; the local gateway proxy (`src/local/gateway.js`, a Node port of `supabase/gateway.conf`) swaps it for the per-install anon JWT in flight, by exact match, on `apikey` and `Authorization` only. Anon-for-anon only — the service-role key never passes through the gateway, so nothing routed through it can escalate.
- **First-launch chooser** — a pristine launch of a stack-carrying build asks the one question every new user has: **Use Mike Cloud** / **Run everything on this Mac** (two cards) with a **Connect to your own server…** link underneath. Either choice persists and retires the chooser; any explicit signal (`--server-url=`, `--local`, a saved `mode` or `serverUrl`) skips it; builds without the stack go straight to the hosted default.
- **Guest mode** — the local login page offers **Continue as guest**: one click in, no typing, for a database living in your own `~/Library`.
- Connect screen also offers "Run locally on this Mac" (only when the build actually carries the stack); a boot progress page streams supervisor status; `--server-url=` still beats local mode so e2e can always retarget.
- Scripts: `local:fetch` (pinned binaries: zonky Postgres 17.10 via npm, PostgREST v14.12 official macOS build, GoTrue v2.189.0 **built from source** — the upstream release asset named darwin-arm64 actually contains Linux ELF binaries), `local:build` (backend tsc + Next standalone with local URLs baked), `local:stage` + `dist:local` (package Mike.app with the whole stack in resources).
- `e2e/local.e2e.mjs` — wipes its userData and drives the true first-run every time: cold initdb → six services → signup → project → library upload → row-menu download → **both guest rounds**.

---

## Why

Running Mike without handing your documents to someone else's server currently means installing Docker Desktop, opening a terminal, writing a `.env`, and supervising six containers. For a solo lawyer with a laptop, that is not a product — it is a prerequisite list.

Everything the compose stack does is already reproducible as ordinary processes. What was genuinely missing was a place to put documents: `main`'s storage layer is S3/R2-only, so even a hand-rolled local deploy had nowhere to write a file and no way to hand a browser a download link. That gap is what most of the backend diff closes.

---

## Reproduce the base case on main

The point is to see that **there is no no-cloud path that isn't Docker**, and that a local deploy would have had nowhere to put documents even if there were.

1. Check out `main` (or #302's tip) and run the shell: `cd desktop && npm install && npm start`. On #302's tip it loads the hosted service; point it somewhere unreachable (`npm start -- --server-url=http://localhost:9`) and you get the connect screen. The only thing the app can do is *aim at a server*. `grep -r "local" desktop/src/main.js` on #302 finds nothing of the sort.
2. `docker compose up` is therefore the only way to run Mike without an account somewhere. Note what that costs a non-technical user: Docker Desktop installed, a terminal, a `.env`, and ~6 containers.
3. Now the piece that would be needed anyway. On `main`, `grep -n 'STORAGE_DRIVER' backend/src/lib/storage.ts` → **no match**: storage is S3/R2-only. Start the backend with no `R2_*` config and upload a document — storage is simply disabled. There is no filesystem driver, `getSignedUrl` has no non-S3 path, and since #456/#304 neither does `createFileReadStream`, which every raw-file download and zip export now goes through.

---

## Verify on this branch

**Prerequisites for `local:fetch`** (it builds GoTrue from source and hydrates Postgres dylib symlinks): `go` (`brew install go`), **`python3`**, `git`, `curl`, `npm`. Each of its three fetch steps is skipped if the artifact is already present.

```bash
git checkout <branch>
cd desktop && npm install
npm run local:fetch     # pinned pg 17.10 (npm) + postgrest v14.12 + gotrue v2.189.0 from source
npm run local:build     # backend tsc + Next standalone with local URLs baked (few minutes)
MIKE_E2E_DEV=1 npm run e2e:local   # automated first-run proof, dev mode
```

Interactively:

```bash
npm run start:local     # === electron . --local ; boots the stack, lands on the local login
```

Then, by hand:

1. Watch the **boot progress page** stream supervisor status. First run does a cold `initdb` and takes about a minute; later runs are much faster.
2. Land on the local login at **`http://localhost:42815/login`**. Click **Continue as guest** → straight into the product, no typing. (Or sign up normally — it autoconfirms offline.)
3. Create a project, upload a document in Library, download it back from the row menu.
4. Look at the disk: `~/Library/Application Support/Mike/local/` now holds `pgdata/`, `storage/` (your document as a **plain file**), `logs/` (one per service), and `secrets.json` (mode 0600).
5. Prove the download is the new capability URL: in DevTools → Network, the download URL contains **`/download/signed/`**, not an `X-Amz-Signature` presigned URL.
6. **New on this rebase — prove the upload is too.** Start an upload and watch the request: the browser `PUT`s to a `/download/signed/` URL on the backend rather than to an object store. Then confirm the bytes landed under `userData/local/storage`.
7. Prove it's offline: DevTools → Network with "record" on through boot → login → signup → authenticated app. Every request targets `localhost:42813/42814/42815`. (The only outbound calls local mode ever makes are to whatever LLM provider you configure — or `localhost:11434` with Ollama — and CourtListener if you use it.)

**Packaged build** (what a downloader actually gets):

```bash
npm run local:stage && npm run dist:local
open dist/mac-arm64/Mike.app
```

Wipe `~/Library/Application Support/Mike` first to see the true first launch: the **chooser** appears with *Use Mike Cloud* / *Run everything on this Mac* / *Connect to your own server…*. Pick local. **Then click something with a real mouse** — see the drag-region note below for why that specific instruction matters.

---

## What changed on the rebase

`main` moved a long way under this branch. Four adaptations, all resolved "main's design wins".

### 1. The fs driver had to grow five new branches — the largest change

`main` added `createFileReadStream`, `uploadFileFromPath`, `getSignedUploadUrl`, `headFile` and `copyFile` to `lib/storage.ts`, and the raw-file download (#456/#304), the zip export and upload processing all route through them. The fs driver covered none of them, so on current `main` local mode would have had **no working download and no working upload at all**.

Each now has an fs branch matching the S3 branch's semantics:

| Function | fs branch | Why it matters |
|---|---|---|
| `createFileReadStream` | `createReadStream` from the resolved path | Raw-file download + zip export both consume it |
| `uploadFileFromPath` | `fs.copyFile` (streams internally) | Generated PDFs, without buffering them whole |
| `headFile` | `stat`, null etag/contentType | Upload sessions HEAD every staged object to confirm the declared bytes arrived |
| `copyFile` | `fs.copyFile` into a created parent | Version copies |
| `getSignedUploadUrl` | mints a `"blob-put:"` capability | Direct upload sessions are now the only upload path |

Plus `writeBlobFromStream` / `discardBlob` for the PUT route. **This is new code written during the rebase.** It has unit and route tests (see *Testing performed*), but it has not been exercised inside a booted `Mike.app` — see *Human-only checks*.

### 2. Guest sign-in moved off supabase-js

`main` replaced the login page's direct `supabase.auth.*` calls with the backend-mediated `login()` / `signup()` helpers in `app/lib/authApi.ts`, so the session now lives in an httpOnly cookie that `AuthContext` reads back through `refreshSession()`. The guest button no longer compiled — `supabase` is not imported on that page any more.

It now calls the same helpers: `login()` first, `signup()` as the first-ever-click fallback, then `refreshSession()` before navigating. Without that refresh the cookie exists but the context does not know it yet, and `OnboardingGate` bounces the guest straight back to `/login` — a failure invisible to a type check, so it is spelled out at the call site. The signup fallback also now checks `requiresEmailConfirmation`, because a guest install has no inbox to confirm from.

### 3. The guest button sits after SSO, not instead of it

`main`'s login page gained a `SsoAuthButton` in exactly the slot this branch used (SSO #435/#458, plus Google auth). Both are kept, in the order Google → SSO → Continue as guest. The button still renders only when the shell's IPC answers with credentials, so a browser and Mike Cloud see the page unchanged — and there is now a test that pins exactly that.

### 4. The blob PUT got its own rate-limit lane

CodeQL's `js/missing-rate-limiting` flagged the new route. On the real app that was a false positive (`app.use(generalLimiter)` is registered well above it), but chasing it surfaced something true: an upload session may carry up to 50 files, and on the shared 300-per-15-minutes IP budget one drag-and-drop would spend a sixth of a user's whole quota, leaving ordinary browsing to 429 behind their own upload. On every other deployment those PUTs go straight to object storage and never touch this process.

The PUT is now skipped from the general limiter and given a dedicated lane (1000 per 15 minutes, both tunable via `RATE_LIMIT_BLOB_UPLOAD_*`) — the same treatment `main` already gives the upload-session endpoints. Generous is not unbounded: a replayed token still cannot hammer the disk forever, which is the property the rule exists to protect.

---

## Two fixes from hands-on first-run testing

Testing the packaged app as a real downloader surfaced one blocker and one UX gap. Full writeup in [the PR comment](https://github.com/open-legal-products/mike/pull/351#issuecomment-5338378169).

### `9599b5bb` — nothing in the product was clickable after local boot

**Base case (on `6426dd6e`):** package with `local:stage` + `dist:local`, wipe `~/Library/Application Support/Mike`, open the app, choose "Run everything on this Mac", wait for the login page → **no physical click lands anywhere.** The page is healthy (it renders and hydrates; a CDP-injected click focuses `#email`) — native clicks simply vanish.

**Cause:** `local-boot.html` marked its whole body `-webkit-app-region: drag` with no `no-drag` island, and Chromium only replaces a window's native drag regions when the *next* document reports regions of its own. The web app declares none, so the boot page's full-window drag region stayed in force over the product — and a drag region consumes every mouse event ([electron/electron#1354](https://github.com/electron/electron/issues/1354)).

**Why every test stayed green:** the e2e suites click over CDP, which injects *below* the native drag layer. **This bug is invisible to the entire automated suite, on both sides of the fix.**

**Fix:** after every committed http(s) navigation the shell inserts `html, body { -webkit-app-region: no-drag; }` on `did-navigate`, forcing the renderer to report a fresh (empty) region set; the boot page also gets the same `no-drag` content island as welcome/connect.

**Replicate the fix:** same steps on `9599b5bb` or later — the login page accepts clicks. The decisive check is a **physical** click.

### `066e7df9` — "Continue as guest" in local mode

**Base case:** local mode demanded an email+password signup for a database living in your own `~/Library`.

**Now:** the login page shows **Continue as guest** — one click in. The guest is a *real* GoTrue user whose random per-install password lives with the other first-run secrets in `userData/local/secrets.json`; auth, sessions, and the data model stay byte-identical to a server deploy. The shell hands the credentials over one gated IPC call (`mike:guest-credentials`): local mode **and** a sender frame on the local frontend's origin (`http://localhost:42815`). A hosted page gets `null`; in a plain browser `window.mikeDesktop` doesn't exist, so the button never renders.

Note this is the **one** privileged bridge call the product web app can make. #302's invariant was "the web app gets zero privileged APIs"; this PR qualifies it to "one origin-gated, local-mode-only read of a local credential", and the preload comment says so. That qualification is the price of the one-click local experience — flagged here rather than buried.

---

1. **Bundle real GoTrue + PostgREST + Postgres binaries** instead of reimplementing their APIs: 436 PostgREST call sites and the full GoTrue MFA surface make shims the expensive, drift-prone path. Cost: ~80 MB of binaries and a from-source GoTrue build step (upstream's macOS release assets are mislabeled Linux binaries), so `local:fetch` needs a Go toolchain and python3 on the builder's machine.
2. **`pg` node client instead of psql** for bootstrap/migrations — the zonky Postgres build ships no psql. Multi-statement files run over the simple protocol in one implicit transaction; verified nothing in schema.sql/migrations is non-transactional.
3. **Fixed ports (42810–42815)** because Next bakes its API origins at build time. A port conflict fails loudly instead of auto-shifting (which would strand the baked URLs). *Caveat:* only the in-process gateway (`:42813`) raises a dedicated bind error; a collision on the other five surfaces as a health-check timeout. Legibility of that path is on the follow-up list.
4. **Placeholder-anon-key swap at the gateway** rather than baking a real key: per-install secrets never exist in the bundle. The swap is exact-match and anon-for-anon only; the service-role key is never routed through the gateway.
5. **No SMTP in v1** — signup autoconfirms offline; email password recovery is unavailable in local mode and the connect screen says so. Guest mode reduces how often this matters.
6. **LibreOffice detect-don't-bundle** — supervisor sets `SOFFICE_BINARY_PATH` when `/Applications/LibreOffice.app` exists; otherwise the product's existing docx-preview fallback applies. Bundling would add ~700 MB.
7. **Blob-token routes are unauthenticated by design** — they replace S3 presigned URLs, which carry no session either; the token is the expiring, object-scoped capability. The permanent `/download/:token` route (auth + DB access check) is untouched. Three token families now share the route and are separated by three HMAC domains (`""`, `"blob:"`, `"blob-put:"`), so a read capability can never be spent as a write capability, in either direction. Tested both ways.
8. **The fs driver is `if (FS_DRIVER)` branches, not a StorageAdapter — and should be re-expressed once #403 lands.** Open PR #403 (`olp-pr/storage-adapter-seam`) introduces exactly the abstraction an fs driver belongs behind. This branch is deliberately **not** rebased onto it: stacking two unmerged PRs would make both harder to land, and #351 is already stacked on #302. So the driver lives as dispatch branches inside `storage.ts`, which is more in-line code in a hot module than the seam would need. **Recommended follow-up: once #403 merges, re-express `STORAGE_DRIVER=fs` as a `StorageAdapter` implementation and delete the branches.** Behaviour is complete and tested either way, so this is cleanup, not a blocker.
9. **The signed PUT streams to disk rather than using a body parser.** `express.raw` would buffer a whole file in the process (upload sessions allow up to 2 GB total), so the route is registered *ahead* of the global JSON parser and pipes `req` straight to a file handle. That ordering is load-bearing: registered after `express.json()`, a `.json` upload would arrive with its stream already drained and the file would be written empty. There is a test that fails on exactly that mistake.
10. **LLM calls still need a key or Ollama** — unchanged product behaviour; keyless demo mode remains a follow-up (#260/#259 lineage).
11. **Guest credentials live in `userData/local/secrets.json` (0600).** Readable by anything on the machine that can read that directory — but so is `pgdata` sitting beside it, so this is the same trust boundary, not a new exposure. Wiping `secrets.json` while keeping `pgdata` orphans the guest account; the button surfaces the auth error rather than silently recovering, because the shell should never delete account data on its own. Supported reset stays "delete `userData/local`".
12. **Guest mode qualifies #302's zero-privileged-APIs invariant** to exactly one origin-gated, local-mode-only IPC read. Rejected alternative: a fixed credential baked into the build — that would put a real secret in the bundle, which the whole per-install-secrets design exists to avoid.
13. **Drag-region reset on `did-navigate`, not a boot-page-only fix.** Fixing only `local-boot.html` would leave the same trap for the next full-window drag region anyone adds. Cost: the CSS insert runs on every navigation, and `did-navigate-in-page` (SPA route changes) is deliberately **not** hooked — the initial commit into the product already resets the regions, and the product declares no drag regions of its own.
14. **CodeQL fixes (`46e4679c`) changed the *shape* of two checks, not their strength.** The path guard was already safe, but `if (resolved !== root && !resolved.startsWith(root + sep))` only establishes a *disjunction* on fall-through, which CodeQL's barrier-guard analysis cannot discharge — hence three `js/path-injection` alerts. Reduced to a single `if (!resolved.startsWith(rootPrefix)) throw`, falling through proves exactly one thing: containment. Behavioural delta is one case — a key resolving exactly to the root is now rejected instead of returned; no call site can produce one. **Where a safe check is illegible to a machine, make it legible rather than suppress the alert.**

---

## Testing performed

**Local, 2026-09-14, on tip `eb30a98d`:**

| Command | Result |
|---|---|
| `npm run build --prefix backend` | clean (`tsc`) |
| `npm test --prefix backend` | **1620 passed / 39 skipped**, 132 files |
| `npm test --prefix frontend` | **1072 passed**, 156 files |
| `npm run lint --prefix frontend` | **0 errors** (34 pre-existing warnings) |
| `npm run build --prefix frontend` | clean |
| `node --check` on every file under `desktop/src`, `desktop/scripts`, `desktop/e2e` | 12 files, 0 errors |
| `sh -n desktop/scripts/*.sh` | 4 files, clean |
| `git diff --check` | clean |

**Tests added on this rebase — +26 backend, +3 frontend:**

- `backend/src/lib/__tests__/storageFs.test.ts` → **+7 cases** covering the five new fs branches: `headFile` size/absent; `copyFile` into a nested key; `uploadFileFromPath`; `createFileReadStream` (the function raw downloads and zip exports go through); `getSignedUploadUrl` mints a redeemable PUT capability *and* that capability fails the read verifier; `writeBlobFromStream` + `discardBlob`.
- `backend/src/routes/__tests__/blobUpload.route.test.ts` → **new file, 6 cases**: a matching body is stored; a forged or expired token is refused without writing; a *read* token spent as a write token is refused; a size mismatch is rejected and the partial file discarded; a content-type mismatch likewise; and a JSON-typed body is streamed to disk rather than eaten by a parser (the regression that middleware ordering exists to prevent).
- `frontend/src/app/login/page.test.tsx` → **+3 cases**: the guest button stays absent in a plain browser (no `window.mikeDesktop`) — the guarantee that this desktop affordance changes nothing for web users; a returning guest signs in without touching signup; a first-ever click falls through to signup.

**Pre-existing test suite** (unchanged by this rebase, from earlier commits): `storageFs.test.ts` already covered driver enablement without R2 config, upload→download→delete, delete-missing-ok, S3 string-prefix `listFiles`, blob-token `getSignedUrl`, storage-root containment across every fs entry point (6 cases), and blob/permanent token non-interchangeability.

**GitHub CI on `eb30a98d`:** green at the time of writing, including CodeQL (the `js/missing-rate-limiting` alert on the first push of this rebase is fixed by `eb30a98d`).

### ⚠️ Desktop e2e — not re-run this round

**Every desktop e2e result previously recorded on this PR predates two rebases**, and none of the suites were executed in this pass: they need a packaged `Mike.app` plus a booted local stack on a physical machine, which was explicitly out of scope. The suites' selectors were source-verified against `main`'s current signup and onboarding pages in `44274d52`, and all four shared pieces (arch path, signup, onboarding, first-run-overlay dismissal) now live in one `desktop/e2e/helpers.mjs` instead of three divergent copies. Every touched `.mjs` passes `node --check`.

**A green packaged run is owed**, and the previously recorded numbers (`local.e2e.mjs` seven assertions, `flows.e2e.mjs` 11/11, `app.e2e.mjs`, the 548 MB / 115-Mach-O packaging figures, the no-web network proof) should all be re-recorded on it. They are left in the history above as what *was* observed, not as claims about this tip.

---

## Demo

**No GIF was recorded this round, and this PR does not claim one.** The packaged-app demos require a physical machine run — `local:fetch`, `local:build`, `local:stage`, `dist:local`, then launching the app and driving it — which was out of scope for this pass.

More importantly, **the single most valuable demo on this branch cannot be recorded by automation at all.** The drag-region bug above is invisible to CDP: a synthetic click dispatched to the page succeeds while a real one is swallowed by the window's native drag layer, so a screen recording of an automated run would show the bug passing. Only a human clicking with a real trackpad is evidence. It is the first item on the checklist below.

---

## Human-only checks

1. **A physical click after local boot on a freshly packaged build.** The single most important manual check on this branch: CDP clicks pass on *both* sides of the drag-region bug, so no automated suite can prove the fix. Use a real trackpad.
2. **Boot the app and exercise upload + download end to end.** The five new fs branches and the signed PUT are covered by unit and route tests but have **never run inside a live `Mike.app`**. This is the biggest gap this rebase opened.
3. **Click "Continue as guest" in a booted app.** Guest sign-in now goes through the backend auth routes instead of supabase-js. The local stack serves those routes, so it should work — but "should" is doing real work in that sentence.
4. **`npm run local:fetch` end to end.** Network- and toolchain-dependent (GitHub release for PostgREST, a `git clone` of supabase/auth at v2.189.0, npm for the zonky package; go and python3 locally). The claim that upstream's `darwin-arm64` GoTrue asset is actually Linux ELF is also only verifiable by running it.
5. **The packaged-bundle numbers** — 548 MB `local-stack`, 115 nested Mach-O binaries, pg dylib symlinks surviving packaging — all require running `local:stage` + `dist:local` and inspecting the bundle.
6. **The upgrade path over an existing `pgdata`.** The ledger's non-fresh branch (apply only unseen migrations) is never exercised by the e2e, which wipes userData every run. Boot once, add a migration, boot again.
7. **`service_role` on a pre-existing install.** `bootstrapRoles` sets `BYPASSRLS` only on the CREATE path and never `ALTER`s an existing role. Unreachable for anyone who first installs from this build, but not proven safe for an install whose roles predate it.
8. **Port-conflict legibility** — see tradeoff 3. Confirm the error page is readable when something else already owns `:42815`.
9. **Cold-boot timing on a clean Mac** with no Postgres installed. `welcome.html` promises "about a minute"; the e2e allows 180 s.
10. **Executing the signed build** (org Apple Developer certificate — human-only).

---

## Signing (wired + documented; needs the org certificate to run)

Notarization requires every nested Mach-O to carry its own hardened-runtime signature — scanning `local-stack/` found **115** (postgres + dylibs, gotrue, postgrest, plus natives nobody would hand-list: skia in the backend's `@napi-rs/canvas`, sharp/libvips in the frontend standalone). So the signed config is **generated at build time** (`scripts/make-signed-local-config.mjs` scans for Mach-O magic bytes and emits `electron-builder.release.local.json` with `mac.binaries` filled), and `npm run dist:local:signed` builds a notarized DMG under the same `APPLE_*` env contract as the shell's `dist:signed`. The operator runbook is in `desktop/README.md` → "Signing the self-contained build". Like #302's signed path, it can only be *run* with the org's certificate.

## Not in this PR (follow-ups)

- **Re-express the fs driver as a `StorageAdapter` once #403 merges** (see tradeoff 8).
- A CI job that actually runs `desktop/e2e` — nothing does today, which is how the signup steps rotted unnoticed twice.
- Executing the signed build (org Apple Developer certificate — human-only).
- Backup/export menu item; Postgres major-version upgrade path (pinned to 17).
- SMTP sink; demo mode.

https://claude.ai/code/session_01NSG87RB94ikHfQyjT6TP1E


amal66:
## Two additions from hands-on first-run testing (cca0596 → 5500e84)

Testing the packaged app as a real downloader surfaced one blocker and one UX gap.

### 1. `62b8314` fix: nothing in the product was clickable after local boot

**Base case (replicate on cca0596):** package with `local:stage` + `dist:local`, wipe `~/Library/Application Support/Mike`, open the app, choose "Run everything on this Mac", wait for the login page → **no physical click lands anywhere**. The page is healthy (renders, hydrates; a CDP-injected click focuses `#email`) — but native clicks vanish.

**Cause:** `local-boot.html` marks its whole body `-webkit-app-region: drag` with no `no-drag` island, and Chromium only replaces a window's native drag regions when the *next* document reports regions of its own. The web app declares none, so the boot page's full-window drag region stays in force over the product — and a drag region consumes every mouse event (electron/electron#1354). The e2e suites click via CDP, which injects *below* the native drag layer, so they stayed green.

**Fix:** after every committed http(s) navigation the shell inserts `html, body { -webkit-app-region: no-drag; }`, forcing the renderer to report a fresh (empty) region set; the boot page also gets the same `no-drag` content island as welcome/connect.

**Replicate the fix:** same steps on 5500e84 — login page accepts clicks. Note the decisive check is a *physical* click; CDP clicks pass on both sides of the fix.

### 2. `5500e84` feat: "Continue as guest" in local mode

**Base case:** local mode demanded an email+password signup for a database living in your own `~/Library`.

**Now:** the login page shows **Continue as guest** (only inside the local desktop product) — one click in, no typing. The guest is a *real* GoTrue user whose random per-install password lives with the other first-run secrets in `userData/local/secrets.json`; auth, sessions, and the data model stay byte-identical to a server deploy. The shell hands the credentials over one gated IPC call: local mode only, and only to a sender frame on the local frontend's origin — a hosted page gets `null`, and in a plain browser the bridge doesn't exist, so the page behaves identically everywhere else.

**Replicate:** `npm run local:build && npm run local:stage && npm run dist:local`, launch, reach the local login page → "Continue as guest" → product. `e2e/local.e2e.mjs` now proves both the first click (signUp fallback, autoconfirmed offline) and the repeat click (plain sign-in).

**Tradeoffs (explicit):**
- Guest creds are readable by anything on the machine that can read `userData/local` — the same trust boundary as `pgdata` beside it, so no new exposure.
- Wiping `secrets.json` while keeping `pgdata` orphans the guest account (button surfaces the auth error rather than silently recovering — the shell should never delete account data on its own). Supported reset stays "delete `userData/local`".
- Existing installs get guest fields topped up in the secrets file on next read; no version bump.

**Verification:** dev-mode and packaged-app `e2e:local` both fully green including the two guest rounds; remote-mode behavior untouched (bridge answers `null` outside local mode).

PR #355, opened 2026-08-19
[Document panel updates] Version-panel resilience

amal66:
## Summary

Part **3 of 3** of the fix stack for #350 (base: `olp-pr/350-split-3-ask-inputs-text`). Fixes the version-scoped document-panel findings from the #350 review + adversarial validation.

- **V1 (medium)**: a failed version lookup (network error, or a document whose `current_version_id` is null via the `ON DELETE SET NULL` FK) no longer silently swallows the click. On `main`, clicking a citation pill / file chip / doc card while the versions request fails does **nothing** — no tab, no spinner, no error. Now the document opens unpinned on its current version (the pre-#350 behavior). The fallback also **nulls the stale `version_number`**, so the tab is keyed `doc::current` (no duplicate-tab collision with a later successful pinned open) and the badge can't claim "V2" over current-version bytes.
- **V2 (low/medium)**: the resolver skips soft-deleted versions. On `main`, a `version_number` reference to a deleted version resolves to it and the content fetch then errors.
- **V3 (low)**: edit-card "View" is routed through the same version resolver as the other five tab-creation sites, so a legacy annotation without `version_id` no longer opens a duplicate tab alongside the resolved one.

## Base-case replication (on current `main`, which contains #350)

1. **V1**: DevTools → Network → block `*/single-documents/*/versions`, then click any citation pill or unversioned doc card → **nothing happens**. Indistinguishable from a broken button.
2. **V1 collision**: with the request blocked, click a citation carrying `version_number: 2` (tab would key `doc::number:2` — but nothing opens on main, so observe post-fix); unblock, click again → on this PR's *unfixed* precursor this made two tabs for one version.
3. **V2**: delete a document version that an earlier `doc_read` event referenced by number, then click that event → error panel.
4. **V3**: click a download card for a doc, then "View" on an edit card whose annotation predates version columns → two tabs for the same content.

## PR replication (this branch)

1. **V1**: same fault injection → the document opens on its current version; unblock and the next click pins normally. The fallback tab is keyed `doc::current` and carries no version badge.
2. **V2**: the event resolves past the deleted version (falls back to current per V1); no error panel.
3. **V3**: one tab; the legacy annotation resolves to the same key as the download card's.

- **V1's fallback can't distinguish "network hiccup" from "document deleted"** — both open unpinned and the viewer surfaces its own error. Strictly better than a silent no-op; branching on status code is a follow-up.
- **V2 changes an error into a silent substitution**: a citation naming a soft-deleted version now opens a *different* (live) version's bytes — the cited quote won't highlight there. Wrong-content-quietly vs. error-loudly is a judgement call; flagging it.
- **V2 filters client-side on purpose**: the versions route keeps returning soft-deleted rows because `DocumentSidePanel` and `ProjectPageParts` render them with a deleted affordance; server-side filtering would break both.
- **V3 makes `openEditor` async** — "View" on a pre-version-era annotation awaits one fetch; newer annotations carry `version_id` and short-circuit inside the resolver.
- **Rapid interleaved clicks can activate tabs out of order** (slow legacy resolve landing after a fast citation click steals focus). Cosmetic; no request-sequence guard added.
- **The project assistant chat page keeps its own document-id-keyed tab system** — version-keyed tabs apply to the main `ChatView` panel; unifying the project page is a follow-up.
- Reminder of #350's own behavior (unchanged here): "current" is pinned at click time, so a repeat click after an edit opens a second tab rather than refreshing, and every unversioned click costs one `GET /single-documents/:id/versions` round-trip.

## Verification

- Backend + frontend: `tsc` clean, full vitest suites green; word add-in typecheck clean (re-verified after the post-merge rebase).
- V1/V2/V3 tests were added against the unfixed code and failed exactly on the findings (null instead of a document; pinned to the deleted version; duplicate `doc::current` tab), then passed after — including the new fallback-key assertion (`doc::current`, no stale `number:N`).

## Before / After evidence

Reproduced live — **before** on `main` @83b5ce0, **after** on this branch. Seed: a real `sample-nda.docx` uploaded and read by Claude (two clickable citations + a `doc_read`), then edited to create **v2** (a Delaware→New York tracked change with an edit card). All three findings are driven through the version-number resolution path (a citation whose inline `version_id` is absent), which is the path `resolvePanelDocumentVersion` actually governs.

### V1 — a citation click no longer silently swallows on a failed version lookup

`GET …/single-documents/:id/versions` is blocked (window.fetch wrapper; console confirms the block), then a citation is clicked. Before: **nothing** happens — no tab, no error (`resolvePanelDocumentVersion` returns null → `if(!document) return`). After: the document opens **unpinned** (tab has no version badge).

| Before (main) | After (this PR) |
|---|---|
|  |  |

Before-state proof (console block + no-panel DOM check): 

### V2 — a reference to a soft-deleted version falls back to current

The cited version (v1) is soft-deleted (`deleted_at`); v2 stays current. The same citation resolves by `version_number`. Before: the resolver pins to the soft-deleted v1 and the content fetch 404s → error panel. After: the resolver skips the deleted version and opens the current one — no error.

| Before (main) | After (this PR) |
|---|---|
|  |  |

(As the tradeoffs note above says, the after opens *live* bytes for a cited-but-deleted version — the quote won't highlight there. That's the deliberate wrong-content-quietly vs. error-loudly call.)

### V3 — one tab instead of two for a legacy edit annotation

The edit annotation's `version_id`/`version_number` are nulled (legacy row). Clicking the download card **and** the edit-card "View". Before: **two** tabs for the same version (the card resolves to `id:{v2}`; `openEditor` used `ann.version_id ?? null` → a `current` key). After: **one** tab — `openEditor` now routes through the resolver, matching the card's key.

| Before (main) | After (this PR) |
|---|---|
|  |  |

> Honesty note: the fix lives in `resolvePanelDocumentVersion`, which short-circuits when a citation already carries an inline `version_id`. A citation pinned inline to a soft-deleted version therefore still 404s on both sides — that path is out of scope here. These before/after pairs use the version-number path (inline `version_id` absent), which is what the fix governs.

### DIFF
--- a/frontend/src/app/components/assistant/AssistantSidePanel.tsx
+++ b/frontend/src/app/components/assistant/AssistantSidePanel.tsx
@@ -51,8 +51,9 @@ export type AssistantSidePanelTab = DocumentTab | CitationTab | EditTab;
 
 /**
  * A document version owns one panel tab. Explicit versions use their stable
- * version id; older links without one fall back to the version number and then
- * to the document's current version.
+ * version id and legacy links can fall back to a version number. "current" is
+ * reserved for non-versioned sources such as cases; file links are resolved to
+ * a concrete version before their tab is created.
  */
 export function assistantSidePanelTabId(document: PanelDocument): string {
     const version = document.version_id
--- a/frontend/src/app/components/assistant/ChatView.tabIdentity.test.tsx
+++ b/frontend/src/app/components/assistant/ChatView.tabIdentity.test.tsx
@@ -82,6 +82,18 @@ vi.mock("./AssistantMessage", () => ({
             >
                 open document
             </button>
+            <button
+                onClick={() =>
+                    onOpenDocument({
+                        documentId: "document-1",
+                        filename: "agreement.docx",
+                        versionId: null,
+                        versionNumber: 2,
+                    })
+                }
+            >
+                open historical document
+            </button>
             <button onClick={() => onEditViewClick(EDIT, "agreement.docx")}>
                 open edit
             </button>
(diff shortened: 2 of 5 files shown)


willchen96 (approved):
Push a commit to show a warning when a document version is not available instead of showing an unversioned tab.

Issue #323, opened 2026-08-12
Add a component catalog (Storybook or Ladle) for components/ui/*


bmersereau:
## Problem

Follow-up to #322. The first phase of formalizing the design system (docs/design-system.md, primitive consolidation, accessibility pass, contribution guideline) intentionally deferred the component catalog piece to keep that PR scoped and reviewable.

## Proposal

Add a browsable, testable catalog for the shared primitives in `frontend/src/app/components/ui/*` (button, input, dropdown-menu, toggle-switch, search-bar, pill-button, tab-pill-button, cite-button, liquid-dropdown, liquid-surface, plus whatever #322 adds).

Pick between Storybook and a lighter alternative (e.g. Ladle) based on actual setup/maintenance cost for this repo — Storybook is the industry default but has real ongoing overhead for a small OSS frontend; Ladle is a lighter Storybook-compatible alternative worth evaluating first.

Scope to `components/ui/*` only, same as #322 — don't try to catalog all 144 components in `frontend/src/app/components/` up front.

## Acceptance criteria

- [ ] Catalog tool chosen and documented (why, in the PR description)
- [ ] Stories/examples for each primitive in `components/ui/*`
- [ ] Catalog builds/runs cleanly (smoke-tested in CI or locally)
- [ ] Linked from `docs/design-system.md` (added in #322)

## Depends on

#322 (docs/design-system.md should exist first so this can link to it)

PR #325, opened 2026-08-12
feat: add PRD template for scoping larger features

bmersereau:
## Summary

- Adds a lightweight PRD template (`docs/templates/PRD.md`) for scoping larger features/architecture changes, covering Problem, Non-goals/out of scope, Proposed solution, Alternatives considered, Technical approach/affected areas, Success metrics, and Open questions.
- Adds a matching GitHub issue form (`.github/ISSUE_TEMPLATE/prd.md`) so "New PRD" appears as an issue-creation option, using the same sections.
- Adds one sentence to `CONTRIBUTING.md` on when to use the PRD template vs. a plain issue.

Closes #324

## Changes

- `docs/templates/PRD.md` — new PRD template, copy-paste source for scoping issues.
- `.github/ISSUE_TEMPLATE/prd.md` — new issue form with YAML frontmatter (`name`, `about`, `title`) and the same section headers as the template.
- `CONTRIBUTING.md` — one-line guideline on when to reach for the PRD template vs. a plain issue.
- `.gitignore` — ignore dev-cycle scratch files (`.current-issue.md`, `.implementation-plan.md`, `.implement-config`).

## Test plan

- [x] No application code touched — docs/template-only change.
- [x] Verified `.github/ISSUE_TEMPLATE/prd.md` YAML frontmatter parses correctly (Python `yaml.safe_load`), since the repo has no existing pattern for validating issue-template YAML.
- [x] `npm run build --prefix backend` passes.
- [x] `npm run lint --prefix frontend` passes (0 errors).
- [x] `npm run build --prefix frontend` passes (includes typecheck), using CI's placeholder `NEXT_PUBLIC_*` env vars.

PR #357, opened 2026-08-19
feat: formalize design system on existing shadcn/Tailwind foundation

### LINKED ISSUE #322: Formalize a design system on top of the existing shadcn/Tailwind foundation

bmersereau:
## Summary

Formalizes the design system that already exists (shadcn/ui `new-york` + Tailwind v4 CSS-variable tokens) rather than introducing a new one: documents the current tokens and primitives, consolidates three duplicated feature-component patterns into shared primitives, and fixes the accessibility gaps in `components/ui/*`.

No visual redesign, no new dependency, no Storybook (that stays in #323).

Closes #322

### 1. `docs/design-system.md` (new)

Documents the current foundations, derived from `frontend/src/app/globals.css` and an audit of `frontend/src/app/components`:

- **Color** — the `--app-*` surface family (Mike's own liquid-glass chrome) and the shadcn semantic set, with intent per token. Flags two real gotchas: `@theme inline` overrides part of Tailwind's blue scale to Mike azure (and leaves `blue-300/400/500/800/900` untouched, so the scale is discontinuous), and the `--app-*` tokens have no `.dark` values, so dark mode is currently incomplete.
- **Typography** — Inter (`font-sans`) / EB Garamond (`font-serif`), how they're loaded in both the web app and the add-in, and the de facto type scale.
- **Spacing and radius** — the actual subset in use (`px-3`/`py-2`/`gap-2` dominate; `h-7` for compact chrome), plus the `--radius` token and which radii the app really uses.
- **Where components live** — the `components/ui/` vs `shared/ui/` vs `components/shared/` vs feature split, including the rule that `shared/ui/` cannot import from `app/` and that new files there must be registered in the add-in's `@source` list.
- **Choosing** — a stop-at-the-first-match list for existing primitive vs. shadcn registry vs. one-off, with "promote on the second occurrence" as the consolidation trigger.
- **Accessibility baseline** — the rules the primitives now follow.

### 2. Consolidation

Targets were re-derived from the current tree rather than reused from earlier analysis, so they differ from what the issue's "audit for consolidation candidates" bullet might have anticipated. Ranked by (number of duplicate sites) x (how identical the markup is):

- **`EmptyState`** (`components/ui/empty-state.tsx`) — icon + serif display heading + muted copy + optional action. `TableEmptyState` already existed as a *wrapper*, but 10 call sites across 7 files each re-hand-rolled the interior; the heading string had already drifted into two class orderings (`font-serif text-2xl font-medium` vs `text-2xl font-medium font-serif`), and `TRTable` had additionally re-implemented the wrapper with a different max-width. Now one component, one class string.
- **`GlassIconButton`** (`shared/ui/GlassIconButtonUI.tsx` + `components/ui/glass-icon-button.tsx`) — the circular glass icon button. `ModalUI` owned a copy inline but never exported it, so `TREditColumnMenu`, `TRSidePanel` and `DocumentSidePanel` each carried a byte-identical 400-character class string. It lives in `shared/ui/` (following the `PillButtonUI`/`GlassCardUI` convention) so `ModalUI` can use it too — otherwise the most-used instance of the button would have kept its own copy and missed the focus-ring fix. `aria-label` is required by the type.
- **`CheckSquare`** (`components/ui/check-square.tsx`) — the selection square in directory and picker rows, 4 sites across `FileDirectory` and `ProjectPickerModal`, including the indeterminate/disabled variants. Decorative (`aria-hidden`) by default since the row usually owns the semantics; a caller that passes a `role` keeps it exposed and supplies its own `aria-checked`. This also removes a visual drift where `ProjectPickerModal` drew a dot instead of the check mark used everywhere else.

`AddProjectDocsModal` no longer exists — the picker markup it used to hold now lives in `FileDirectory`, which is where the checkbox consolidation landed.

### 3. Accessibility pass on `components/ui/*`

Each was re-verified against current source; fixes applied where still broken.

- `cite-button` — added `type="button"` (it would submit an enclosing form), an `aria-label` for the icon-only mode (previously only `title`, the weakest name source) that is deliberately *not* set when the visible "Cite" label is shown so the accessible name still contains the visible label (WCAG 2.5.3), a focus-visible ring, and `role="status"` so the Cite→Copied swap is announced.
- `pill-button` — the base class in `PillButtonUI` gained a focus-visible ring. Fixing it there covers the plain button, the `asChild`/`Slot` path, and the Word add-in in one place.
- `tab-pill-button` — focus-visible ring.
- `toggle-switch` — the off-state track was `bg-gray-100`, roughly 1.07:1 against white and well under the 3:1 that WCAG 1.4.11 requires for a control boundary. Now `bg-gray-300` with an inset `ring-gray-400`. The on state already passes.
- `search-bar` — the input has `outline-none` and the only focus signal was a near-imperceptible wrapper tint; the wrapper now also gets a ring, and the clear button gets its own. Added a `label` prop (default "Search") because a placeholder is not an accessible name.
- `form-field` — `FORM_CONTROL_GLASS_CLASS` and the `minimal` variant both suppressed the outline with nothing in its place, so these inputs had no keyboard focus state at all. Both now carry a ring. `FieldLabel` also silently dropped all forwarded props on its `as="p"` / `as="span"` branches, which meant an `id` passed for `aria-labelledby` never reached the DOM.
- `liquid-dropdown` — the glass skin replaced the shadcn focus background with `app-surface-hover`, a ~1% luminance step that made the highlighted item indistinguishable from a hovered one. Now `app-surface-active` plus an inset ring.
- `button` (shadcn) and `dropdown-menu` (Radix) were checked and left alone — their focus handling is already adequate and diverging from the registry costs more than it gains. `button`'s missing `type` default is upstream shadcn behavior and is called out in the docs instead.

### 4. `CONTRIBUTING.md`

One additive "Frontend UI Work" section on checking `components/ui/` and the shadcn registry before writing a new component, and on the promote-on-second-occurrence rule. Existing sections are untouched. Also linked the new doc from `docs/README.md`.

## Test plan

- [x] Unit tests added — `empty-state`, `check-square` and `glass-icon-button` test files, plus new accessibility assertions in `cite-button`, `pill-button`, `toggle-switch`, and new `search-bar`, `form-field` and `tab-pill-button` test files. Written failing first (17 failures), then implemented.
- [x] `npx vitest run` (frontend) — 447 passing, up from 404. The 4 failures in `src/app/lib/mikeApi.test.ts` (`blob.text is not a function`) are pre-existing, present on unmodified `main` locally, unrelated to this change, and green in CI — a local jsdom/Node artifact, so no issue was filed for them.
- [x] `npm run lint --prefix frontend` with the ESLint cache cleared — 0 errors.
- [x] `npx tsc --noEmit` (frontend) — clean.
- [x] `npm run build --prefix frontend` — compiles, 25 static pages generated.
- [x] Word add-in `npm run typecheck` and `webpack --mode production` — both pass, since `ModalUI` now depends on the new `shared/ui` file and the add-in's Tailwind `@source` list had to learn about it.
- [ ] No E2E added — this is a refactor plus a11y attributes with no new user-facing flow.

No changes to `package.json` or `package-lock.json`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,6 +13,22 @@ Thanks for helping improve Mike. Please keep contributions small, focused, and e
   infrastructure or migration requirements in the same PR.
 - Do not commit secrets, API keys, private documents, or local `.env` files.
 
+## Frontend UI Work
+
+Before writing a new component, check whether one already exists. Look first in
+`frontend/src/app/components/ui/` (and `frontend/src/shared/ui/` for anything the
+Word add-in also renders), then in the [shadcn/ui](https://ui.shadcn.com)
+registry — the project is configured for it in `frontend/components.json`, so
+`npx shadcn@latest add <component>` lands a component in the right place with
+the right style and tokens. Write a one-off in the feature directory only when
+the markup is genuinely specific to that feature; once the same markup shows up
+in a second feature file, promote it into `components/ui/` with a test and
+replace the copies. Use the documented color, typography, spacing and radius
+tokens rather than raw hex values, and keep the accessibility baseline (visible
+focus ring, accessible name on icon-only controls, `type="button"`, ARIA state
+alongside color). See [docs/design-system.md](docs/design-system.md) for the
+tokens, the primitive inventory, and the full baseline.
+
 ## Before Opening a PR
 
 - Run the relevant build or test command for the area you changed.
--- a/docs/README.md
+++ b/docs/README.md
@@ -21,6 +21,11 @@
 - [Tamper-evident exports](tamper-evident-exports.md) — document hashes and
   optional signed manifests
 
+## Frontend
+
+- [Design system](design-system.md) — color/typography/spacing tokens, the shared
+  `components/ui` primitives, and the accessibility baseline
+
 ## Testing and CI
 
 - [End-to-end tests in CI](e2e-ci.md)
--- a/docs/design-system.md
+++ b/docs/design-system.md
@@ -1,0 +1,222 @@
+# Design system
+
+Mike's UI is shadcn/ui (`new-york` style, see `frontend/components.json`) on
+Tailwind v4 with CSS-variable tokens, plus Lucide icons. This page documents what
+already exists so contributors can reuse it instead of re-deriving it. It is a
+description of the current system, not a proposal for a new one.
+
+Everything below lives in `frontend/src/app/globals.css` unless stated otherwise.
+
+## Where components live
+
+| Location | What belongs there |
+| --- | --- |
+| `frontend/src/app/components/ui/` | Shared web primitives. Small, unopinionated, no data fetching, no feature knowledge. |
+| `frontend/src/shared/ui/` | Primitives shared by the web app **and** the Word add-in. Framework-light: React, `lucide-react`, `clsx`, `tailwind-merge` only. |
+| `frontend/src/app/components/shared/` | App-level building blocks that assume the app shell — the table layer (`TablePrimitive.tsx`, `TableToolbar.tsx`), page headers, sidebars. |
+| `frontend/src/app/components/<feature>/` | Feature components. Compose the above; do not restate their markup. |
+
+`src/shared/ui/` may not import from `src/app/`. It is compiled into the add-in
+build, which has no Next.js app directory. When you add a file there you must
+also register it for Tailwind class scanning in
+`word-addin/src/taskpane/styles.css`:
+
+```css
+@source "../../../frontend/src/shared/ui/YourThingUI.tsx";
+```
+
+The convention for a cross-target primitive is a `XxxUI.tsx` in `src/shared/ui/`
+plus a thin re-export in `components/ui/` so web callers use one import path —
+see `GlassCardUI`/`glass-card.tsx`, `PillButtonUI`/`pill-button.tsx`, and
+`GlassIconButtonUI`/`glass-icon-button.tsx`.
+
+## Color tokens
+
+Two families coexist. Prefer a token over a raw Tailwind palette class whenever
+one exists; prefer either over a hex literal.
+
+### App surfaces (Mike's own)
+
+These back the "liquid glass" chrome and are the ones most feature code needs.
+
+| Token | Utility | Light value | Intent |
+| --- | --- | --- | --- |
+| `--app-background` | `bg-app-background` | `#f9fafb` | Page canvas behind panels. |
+| `--app-surface` | `bg-app-surface` | `#fdfdfe` | Resting surface of a panel, table header, menu. |
+| `--app-surface-hover` | `hover:bg-app-surface-hover` | `#f9fafb` | Row/item hover. |
+| `--app-surface-active` | `bg-app-surface-active` | `#eff0f3` | Selected or pressed row/item. |
+| `--app-floating` | `bg-app-floating` | `#fefefe` | Detached floating elements above a surface. |
+
+Use the class-name constants in `components/ui/liquid-surface.ts`
+(`APP_SURFACE_HOVER_CLASS`, `APP_SURFACE_ACTIVE_CLASS`,
+`LIQUID_PANEL_SURFACE_CLASS`, …) rather than retyping the utilities, so the
+hover/active pair stays consistent.
+
+`--color-azure: 0, 136, 255` is a raw RGB triple consumed as
+`rgb(var(--color-azure))` by the statute/regulation viewer styles.
+
+### shadcn semantic tokens
+
+The standard shadcn set is present and wired through `@theme inline`:
+`background`/`foreground`, `card`, `popover`, `primary`, `secondary`, `muted`,
+`accent`, `destructive`, `border`, `input`, `ring`, the `sidebar-*` family, and
+`chart-1`…`chart-5`. Values are `oklch()`; a `.dark` block overrides them.
+
+Use these when you pull a component from the shadcn registry, or when you want
+the meaning ("this is the destructive action") rather than a specific color.
+
+### Blue is overridden
+
+`@theme inline` redefines part of Tailwind's blue scale to Mike's azure:
+
+```css
+--color-blue:     rgb(0, 136, 255);
+--color-blue-50:  rgba(0, 136, 255, 0.05);
+--color-blue-100: rgba(0, 136, 255, 0.1);
+--color-blue-200: rgba(0, 136, 255, 0.3);
+--color-blue-600: rgb(0, 136, 255);
+--color-blue-700: rgb(0, 120, 230);
+```
+
+`bg-blue-600` is therefore Mike azure, not Tailwind blue. `blue-300`, `-400`,
+`-500`, `-800` and `-900` are *not* overridden, so the scale is discontinuous —
+stay on the overridden steps for brand blue.
+
+### Dark mode
+
+`@custom-variant dark (&:is(.dark *))` — class-based, not
+`prefers-color-scheme`. The shadcn tokens have `.dark` values; the `--app-*`
+surface tokens do **not**. Since the app chrome is built on `--app-*`, dark mode
+is currently incomplete. Do not treat a `dark:` variant as supported until those
+tokens gain dark values.
+
+## Typography
+
+Both faces are loaded with `next/font/google` in `frontend/src/app/layout.tsx`
+and exposed as CSS variables on `<body>`:
+
+| Face | Variable | Utility | Used for |
+| --- | --- | --- | --- |
+| Inter | `--font-inter` → `--font-sans` | `font-sans` (body default) | All UI text. |
+| EB Garamond | `--font-eb-garamond` → `--font-serif` | `font-serif` | Display headings, legal document body copy, tracked-change cards. |
+
+The Word add-in supplies the same two variables from
+`word-addin/src/taskpane/styles.css` (fonts loaded via `<link>` in
+`index.html`), so `shared/ui` components render identically in both targets.
+
+De facto type scale in the app chrome, most to least common: `text-xs` (dense
+table and control text — the default for most chrome), `text-sm` (body copy,
+normal-size buttons), `text-[10px]` (badges, superscript citations),
+`text-2xl font-serif` (display headings and empty states). There is no separate
+heading component; use `EmptyState` for empty-state headings and `PageHeader`
+for page titles so the display style stays in one place.
+
+## Spacing and radius
+
+There is no bespoke spacing scale — Tailwind's default applies. The de facto
+subset actually in use across `frontend/src/app/components`, by frequency:
+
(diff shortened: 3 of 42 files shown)


bmersereau:
## PR Review: feat: formalize design system on existing shadcn/Tailwind foundation

### Summary
This is a well-executed, correctly-scoped consolidation + accessibility pass: three shared primitives (`EmptyState`, `GlassIconButton`, `CheckSquare`) extracted from genuinely duplicated markup, real accessibility bugs fixed with correct WCAG reasoning, and no unrelated scope creep. Code quality is high. The one real problem is process, not content: the branch is stale against current `main` with an actual conflict.

### Risk Assessment
- **Risk Level**: Low (content) / Blocker (mergeability, until rebased)
- **Blast Radius**: `GlassIconButtonUI` now backs `ModalUI` plus 3 direct consumers, and `pillButtonUIClassName`'s base string backs every pill button app-wide (including the Word add-in). Both are correctly consolidated, but a mistake in either would be widely visible.
- **Rollback Plan**: Standard revert; no data/migration involved, no dependency changes.

### severity:blocker — stale branch, real conflict
`gh pr view 357` reports `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. Confirmed the cause: this branch's merge base is `64e2865`, but `origin/main` has moved to `31ee5da` (23 commits, a word-addin review-flow feature). Two files genuinely collide:
- **`frontend/src/shared/ui/PillButtonUI.tsx`** — upstream removed `border` from the base class string and simplified the `toneClasses` (same line region this PR edits to add the focus-visible ring). Verified via `git merge-tree`: real content conflict, not adjacent-line noise — git will not auto-resolve this.
- **`CONTRIBUTING.md`** — both this PR and #325 (merged since) add new sections in the same area.

Neither is a silent-revert risk — GitHub already refuses to merge rather than pick a side. But it needs a rebase before merge, and the `PillButtonUI.tsx` resolution needs to keep *both* changes (upstream's border removal + this PR's focus ring), not just favor one side mechanically.

### severity:major — pre-existing test failures, now tracked
4 tests in `frontend/src/app/lib/mikeApi.test.ts` fail locally (`blob.text is not a function`). Verified independently: same 4 failures reproduce on unmodified `origin/main`, so this is not a regression from this PR. This exact failure has now been observed and separately waved off as "local-only, no issue needed" across three PRs in a row (#325, and both attempts at #322) without ever being tracked, so I filed **[#361](https://github.com/willchen96/mike/issues/361)** — likely a Node 22 (CI) vs. Node 24 (local) skew in how the mock `Blob`/`Response` behaves. Not blocking this PR, but it needed to stop being re-discovered and re-dismissed every review.

### Correctness — praise
- `GlassIconButtonUI` extraction: verified all 4 sites (`ModalUI`, `TREditColumnMenu`, `TRSidePanel`, `DocumentSidePanel`) migrated to byte-identical rendered output plus the new focus ring, no behavior drift. `aria-label` required at the type level, `type="button"` defaulted — good compile-time enforcement.
- `CheckSquare`: correctly `aria-hidden` by default, only exposed to assistive tech when a caller passes `role` (and then the caller owns `aria-checked`) — the right ownership split for a presentational indicator inside a semantic row control.
- `cite-button`: the conditional `aria-label={showText ? undefined : "..."}` is exactly right for WCAG 2.5.3 — it would have been an easy, common mistake to always set the label and silently break "accessible name doesn't contain the visible label" for the text variant. `role="status"` on the Cite→Copied swap is a real fix, not decoration.
- `form-field.tsx`: `FieldLabel`'s `as="p"`/`as="span"` branches previously dropped all forwarded props, meaning an `id` passed for `aria-labelledby` silently never reached the DOM. Real bug, correctly found and fixed.
- Verified no `frontend/src/shared/ui/*` file imports from `app/` (the import-direction rule this PR itself documents) — held.
- `package.json`/`package-lock.json` untouched, confirmed via diff.
- `AddProjectDocsModal` correctly recognized as already-deleted upstream (merged into `AddDocumentsModal`/`FileDirectory`); the check-square consolidation correctly targets `FileDirectory.tsx` instead of chasing the deleted file.

### Test Coverage Assessment
- Ran the full frontend suite in an isolated worktree: 447 passing, 4 failing (pre-existing, see above) — matches the PR's own claim exactly.
- Lint (cache cleared): 0 errors, 35 warnings, all pre-existing (spot-checked `WorkflowList.tsx`'s `setLoadError` warning via `git blame` — predates this PR).
- `tsc --noEmit`: clean.
- New/extended test files for every touched primitive, written failing-first per the PR description — consistent with the repo's existing test style.

### Follow-up issue discipline — praise
[#358](https://github.com/willchen96/mike/issues/358), filed by this PR's author, correctly identifies that `FileDirectory.tsx`'s `role="checkbox"` `CheckSquare` instances are nested inside `<button>` row wrappers (invalid nested-interactive HTML, mouse-only for keyboard users) — and correctly scopes the structural fix *out* of this PR rather than trying to restructure rows while also doing the consolidation. Confirmed this is real (the two sites with `role="checkbox"` are inside `<button onClick=...>` rows) and correctly not "fixed" here.

### Questions
None outstanding — no ambiguity in scope or implementation that needed clarification.

### Suggestions (Non-Blocking)
- Once rebased, double check the `PillButtonUI.tsx` resolution keeps upstream's `toneClasses` border removal *and* this PR's focus-visible ring in the same base string — a mechanical "take ours" resolution would silently reintroduce the `border` class upstream just removed.

### Verdict
- [x] **Request changes** — one blocking item: rebase onto current `main` and resolve the `PillButtonUI.tsx` / `CONTRIBUTING.md` conflicts (both are combine-both-sides, not pick-one). No code-quality concerns; this will be **Approve** once rebased.

### What I Verified
- [x] Read the full diff (37 files, +1105/-238)
- [x] Ran tests locally (isolated worktree, both this branch and unmodified `main` for the pre-existing-failure comparison)
- [x] Ran lint (cache cleared) and `tsc --noEmit` locally
- [x] Checked related code not in diff (`ModalUI.tsx`'s consumption of the new primitive, import-direction rule, `AddDocumentsModal`/`FileDirectory` post-refactor state)
- [x] Reviewed linked issue (#322) and its correctly-filed follow-up (#358)
- [x] Confirmed stale-branch/conflict status via `git merge-tree` simulation, not just GitHub's summary field

willchen96:
Pushed follow-up commit `534fe2b7` with the review changes:

- Preserved the borderless `PillButtonUI` styling from current `main` while retaining the PR's keyboard-only `focus-visible` rings.
- Updated the Files, Templates, and project Documents root empty tables to use the shared `EmptyState` styling with their respective titles, the text “Upload documents or drop them here”, and an `Upload` pill button.
- Kept the existing whole-placeholder click target, upload/import action, and drag-and-drop handlers intact.
- Fixed `EmptyState` icon sizing so both SVG icons and Next/Image-backed app icons render at 32px.
- Corrected `frontend/components.json` aliases to the actual `src/app` component, UI, utility, library, and hooks directories.
- Clarified `docs/design-system.md` that Mike uses selected shadcn-derived `new-york` controls alongside its custom liquid-glass component system.

Validation: frontend TypeScript, Word add-in TypeScript, targeted ESLint, and 19 pill-button/empty-state tests all pass.

willchen96 (approved):
Looks great - thank you for this useful contribution

PR #364, opened 2026-08-20
feat: add Synthetic as a fourth model router

duncanmcqueen:
## Summary

Adds [Synthetic](https://synthetic.new) — a subscription gateway for open-weights models (GLM-5.2, Kimi K3, Qwen3.6, gpt-oss-120b, Nemotron) on an OpenAI-compatible endpoint — as a **fourth model router**.

> ⚠️ **Stacked on #340.** The first commit here is #340's; review only the second (`feat: add Synthetic as a fourth model router`). Please merge #340 first — this rebases onto `main` cleanly afterwards. #340 generalized the router layer to `ROUTER_SLUGS`, which is what makes this one small.

## What's added

- `lib/llm/openrouter.ts` — `synthetic` joins the `RouterProvider` union: `https://api.synthetic.new/openai/v1`, overridable via `SYNTHETIC_BASE_URL`.
- `lib/userApiKeys.ts` + migration `20260821_01` — `synthetic` API-key provider (`SYNTHETIC_API_KEY` env or encrypted per-user storage), with `schema.sql` kept in step for the drift check.
- `routes/models.ts` — `GET /models/synthetic`, filtered to tool-capable text models exactly as the OpenRouter and Vercel routes are.
- Settings → API Keys gains a Synthetic field; Settings → Routers gains a Synthetic picker; both model pickers and the Word add-in gain a "Synthetic" group.

## Two things that needed handling rather than reuse

**1. Prices carry a currency symbol.** Synthetic quotes `"prompt": "$0.000001"` where every other router sends a bare number. The client's cost formatter does `Number(value) * 1_000_000`, so the `$` renders every Synthetic price as `NaN`. `catalogPrice` now strips a leading symbol, which is a no-op for the existing routers.

**2. Ids come in two families, and the shared label helper mangled one.** Synthetic publishes pinned `hf:<vendor>/<model>` ids *and* floating aliases `syn:<size>:<capability>`. `modelDisplayName` reads `:` as a variant separator (for Ollama tags like `llama3:8b`), so `syn:large:text` rendered as **"Syn (Large:text)"**. Synthetic options get a dedicated `syntheticModelDisplayName` — mirrored in the add-in and pinned by the cross-package drift guard — giving:

| catalog id | label |
|---|---|
| `syn:large:text` | Large (text) |
| `syn:small:vision` | Small (vision) |
| `hf:openai/gpt-oss-120b` | GPT OSS 120B |
| `hf:moonshotai/Kimi-K3` | Kimi K3 |

**One judgement call worth a reviewer's eye:** in Settings → Routers an alias is listed under the Hugging Face id it *currently* resolves to (`syn:large:text` → "zai-org/GLM-5.2", with the raw id on the line beneath), because that is the only field saying what actually answers behind a floating alias. The composer picker has only the id, so it shows "Large (text)". Happy to make both surfaces use the terse form if you'd rather they match exactly.

## Validation

Verified the route's transform against the live catalog: all 10 published models pass the tool-capable text filter, and prices parse to the expected per-million figures ($0.10–$3.00/M input).

- **backend**: `tsc --noEmit` clean; `vitest run` — **710 passed**, 24 skipped (+11 over #340's 699).
- **frontend**: `tsc --noEmit` clean; `eslint` **0 errors**; `vitest run` — **469 passed** (+3).
- **word-addin**: `typecheck:app` and `typecheck:e2e` clean; catalog-parity and composer-availability drift guards extended to the new router.

Three frontend files (`mikeApi`, `useSelectedModel`, `ChatInput.modelSelection`) fail identically on unmodified `main` in my sandbox — `Blob.text is not a function`, `window.localStorage` undefined, a local jsdom/env issue rather than a code change. CI should run them.


willchen96:
@duncanmcqueen thanks for this - but will close this for now without merging in as I'm not sure how many people actually use Synthetic, open to adding it later on if it gets more traction. Would prefer to keep the list of routers limited to those with the highest usage.

duncanmcqueen:
Makes sense.  Synthetic is a pretty interesting alternative that I just discovered - and it heavily just uses open weight models.  May make sense to keep on your radar :)

PR #371, opened 2026-08-22
Support external_ref get-or-create on POST /chat/create

christopher-soapbox:
## Summary
- Adds an optional `external_ref` field to `POST /chat/create`. When provided, a repeat call with the same `(user_id, external_ref)` pair returns the existing chat's id instead of creating a new one.
- Lets an external integration (e.g. a Slack thread bridge) key a chat by a stable external identifier without tracking `chat_id` itself.
- No behavior change when `external_ref` is omitted.

## Test plan
- [x] New tests: get-or-create returns the same id on repeat calls with the same external_ref; a fresh external_ref creates a new chat; omitting external_ref entirely is unaffected (regression check)
- [x] `npx vitest run src/routes/__tests__/chat.test.ts` — 3/3 passing
- [x] Full backend suite run with no regressions

### DIFF
--- a/backend/src/routes/__tests__/chat.test.ts
+++ b/backend/src/routes/__tests__/chat.test.ts
@@ -1,0 +1,134 @@
+import express from "express";
+import request from "supertest";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("../../middleware/auth", () => ({
+    requireAuth: (
+        _req: unknown,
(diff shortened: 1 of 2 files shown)


christopher-soapbox:
Closing — this is an internal fork-only change for our deployment, not something we intend to propose upstream. Keeping it on our fork (christopher-soapbox/mike-1) instead. Sorry for the noise.

PR #378, opened 2026-08-23
fix(models): stop silently defaulting to Gemini outside Google-OAuth deployments

duncanmcqueen:
## Summary

Removes the hardcoded `gemini-3-flash-preview` default. Deployments with Sign-in-with-Google enabled keep it; everywhere else, models resolve from what the signed-in user can actually use, and requests fail with an actionable message when nothing is configured.

## What changed

**Backend**
- New `lib/googleOauth.ts`: detects the Supabase external Google provider (`SUPABASE_AUTH_EXTERNAL_GOOGLE_CLIENT_ID`/`_SECRET`) and exposes `legacyDefaultModel(tier)` — the historical Gemini constants, retained **only** for Google-OAuth deployments.
- `getUserModelSettings`: title/tabular fallbacks become availability-based chains (first-party keys by tier → saved router selections). When nothing is usable they return `null` for non-Google-OAuth deployments; OAuth deployments keep today's Gemini defaults.
- Chat streaming: requests without a model use saved router selections first, then first-party keys, and fail loudly ("No AI provider is configured…") instead of running an unchosen model. Explicit-but-unknown model ids now also fail loudly instead of silently degrading to the default.
- Title generation and tabular generation entry points skip or return a clear 409 when no provider is usable.
- Profile payloads expose nullable `titleModel`/`tabularModel` plus `composerDefaultModel` so clients preselect the deployment's intended default without hardcoding it.

**Frontend / add-in**
- Composer renders an explicit "Select model" state instead of assuming Gemini, resets stale selections atomically, and preselects the first usable candidate — server default first, then saved router selections, then first-party models filtered by real key availability.
- Word add-in catalog mirrors the change; `REACT_APP_DEFAULT_MODEL` remains available as an operator override.
- Tabular views skip key checks while no model is selected.
- `.env.example` documents that setting both Google vars retains the legacy default.

## Why

A stale composer selection (e.g. after a catalog rename) or an omitted model field silently ran Gemini — surfacing only as an opaque stream failure (in one observed deployment, a monthly spend-cap 400) that users could not connect to their own selection. Model choice should come from configuration and user keys, not a compile-time constant.

## Testing performed

- `backend`: full vitest suite green on Node 22 (727 passed), including new coverage: availability-based fallbacks, null-when-nothing-usable outside Google-OAuth deployments, retained Gemini defaults under the Google flag, and loud rejection of unknown model ids.
- `frontend`: full vitest suite green (548 passed; three pre-existing `wordAddin/*` load failures are identical on pristine `main` in containerized runs and unrelated to this change), `tsc` clean, eslint clean.
- `word-addin`: `tsc` clean.
- Manually exercised both modes against a local stack: with no Google env, a request omitting `model` streams from the user's saved OpenRouter selection; with the Google vars set, the legacy Gemini default is restored.

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -15,6 +15,10 @@ SUPABASE_SECRET_KEY=your-supabase-service-role-key
 
 # Local Supabase CLI Google OAuth credentials. Google is enabled by default;
 # set [auth.external.google].enabled to false to opt out locally.
+# When BOTH values are set, the deployment keeps its historical Gemini
+# default model. Without them there is no hardcoded default: models resolve
+# from each user's saved selections and API keys, and requests fail loudly
+# when nothing is configured.
 SUPABASE_AUTH_EXTERNAL_GOOGLE_CLIENT_ID=
 SUPABASE_AUTH_EXTERNAL_GOOGLE_CLIENT_SECRET=
 
--- a/backend/src/lib/__tests__/userSettings.test.ts
+++ b/backend/src/lib/__tests__/userSettings.test.ts
@@ -20,6 +20,7 @@ vi.mock("../routerModels", async () => ({
 vi.mock("../supabase", () => ({ createServerSupabase: vi.fn() }));
 
 import { getUserModelSettings } from "../userSettings";
+import { providerForModel } from "../llm/models";
 
 function profileDb(row: Record<string, unknown> | null) {
     const chain: Record<string, unknown> = {};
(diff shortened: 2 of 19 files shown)


willchen96:
Thanks @duncanmcqueen this PR points out a valid issue of the Gemini fall backs applying even when the API key is not available. However, it think this issue needs a fuller refactor that (1) lets users select models per tab review and the ability to set a default in settings (2) does not include an fallback to Gemini keys where Google OAuth API keys are available, since Google OAuth users might not want that (3) title generation defaults to the cheapest model of the selected provider in the main chat or if a router is selected, the same router model, can be overriden by a setting in model preferences. Will be making a PR on this.

willchen96:
See #383

PR #379, opened 2026-08-23
fix(chat): surface empty upstream model completions instead of silence

duncanmcqueen:
## Summary

When an upstream provider ends a chat stream cleanly but produces **zero content** (observed intermittently via OpenRouter), the client currently renders nothing — the composer just sits there looking hung. This PR emits an explicit, safe-to-display error event instead:

> *The model returned an empty response. Try again, or pick a different model.*

## What changed

- `routes/chat.ts`: after `runLLMStream` resolves, if `fullText` is empty/whitespace and no events were produced, emit the retry-guidance error event followed by `[DONE]`.
- Turns that produced tool artifacts (documents, ask-inputs flows, etc.) keep their existing completion path — the guard only fires when nothing visible was generated.

## Why

An empty 200-complete stream is indistinguishable from a hang to the user, and the underlying trigger is provider-side (retrying or switching models works). Surfacing it turns a dead-end into actionable guidance.

## Testing performed

- New integration test in `chat.routes.test.ts`: empty `runLLMStream` result → SSE contains the `error` event with `safe_to_display: true` and terminates with `[DONE]`.
- Full backend suite on Node 22: 726 passed / 25 skipped.

### DIFF
--- a/backend/src/__tests__/integration/chat.routes.test.ts
+++ b/backend/src/__tests__/integration/chat.routes.test.ts
@@ -341,6 +341,28 @@ describe("POST /chat — streaming endpoint", () => {
         );
     });
 
+    it("surfaces an empty upstream completion as a visible retry error", async () => {
+        // Some providers end the stream cleanly but produce no content.
+        // Silence reads as a hung composer, so the route emits an explicit,
+        // safe-to-display error event before closing the stream.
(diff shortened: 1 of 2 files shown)


willchen96 (approved):

willchen96:
Thanks for this @duncanmcqueen

PR #392, opened 2026-08-27
fix(citations): reject out-of-order, distant, and meaning-inverting ellipsis quotes

### LINKED ISSUE #391: Citation verifier marks meaning-inverting ellipsis quotes as verified
## Problem

`verifyQuoteAgainstSource` (`backend/src/lib/chat/verifyCitations.ts`) splits an ellipsis-abbreviated quote into fragments and locates each fragment **independently** anywhere in the source. Nothing checks fragment order, how much text the ellipsis omitted, or what it omitted. Any quote assembled from words that occur somewhere in the document earns `verified: true` — the green badge in the citation UI.

## Reproduction (current `main`)

```ts
import { verifyQuoteAgainstSource } from "./backend/src/lib/chat/verifyCitations";

const source = "Under clause 9, the landlord is not permitted to terminate this lease during the fixed term.";
verifyQuoteAgainstSource(source, "the landlord is ... permitted to terminate");
// → { verified: true, needs_correction: false, source_excerpt: "the landlord is ... permitted to terminate" }
```

The quote asserts the **opposite** of the source — the ellipsis swallowed "not" — and it is badged verified. Other abuse classes that also verify today:

- **Reversed fragments**: `"beta clause ... alpha clause"` verifies even when alpha precedes beta in the document.
- **Distant quilting**: two fragments pages apart, joined by one `...`, verify as if they were one abbreviated passage.
- **Edge-truncation**: `"... permitted to terminate this lease"` verifies against `"...is not permitted to terminate this lease"` — the leading ellipsis hides the same-clause negator.

## Why this matters

The verification badge exists so a user does *not* have to re-read the source to trust a quote. In a legal product, a badge that can endorse a meaning-inverted quote is worse than no badge: it converts a hallucinated or drifted citation into an apparently checked one. The existing tests only exercise the ellipsis path positively (legitimate abbreviations verify) — none attack it.

## Expected behavior

An ellipsis attests: fragments in document order, a short omission, and an omission that does not change meaning. The verifier should refuse the badge (`verified: false`) when a quote violates any of these — order violations, omissions beyond a bounded span, or omissions/edge-truncations that swallow a negation or exception word in the same clause.

Fix incoming in a linked PR.

amal66:
## Summary

Fixes #391.

On `main`, this verifies — badge and all:

```ts
const source = "Under clause 9, the landlord is not permitted to terminate this lease during the fixed term.";
verifyQuoteAgainstSource(source, "the landlord is ... permitted to terminate");
// → { verified: true }   // the quote asserts the OPPOSITE of the source
```

The ellipsis branch of `verifyQuoteAgainstSource` located each quoted fragment **independently, anywhere** in the source. Fragment order, omission size, and omission content were never checked, so reversed fragments, fragments quilted from distant pages, and omissions that swallow "not" all earned `verified: true`. In a product whose purpose is that users *don't* re-read the source, a badge that can endorse a meaning-inverted quote is worse than no badge.

## What changed

`backend/src/lib/chat/verifyCitations.ts` — the ellipsis branch now enforces what an ellipsis actually claims to a reader:

1. **In-order matching with backtracking** (`locateSegmentsInOrder`): fragments must appear in document order; each search starts after the previous match. If an early match strands a later fragment or overruns the gap bound, the matcher retries from the next occurrence, under a fixed attempt budget (64) that fails **closed**.
2. **Bounded omission**: at most `MAX_OMITTED_SPAN_CHARS` (500) may be omitted between consecutive fragments — ~two long legal sentences. Beyond that the quote is quilting distant text, not abbreviating one passage.
3. **Meaning-inversion guard**: if the omitted span contains a negation/exception word (`not`, `no`, `never`, `cannot`, `n't`, `without`, `except`, `unless`, `exclud*`, `prohibit*`, `forbid*`, `denie[sd]/deny`, `void`), the badge is refused. A **leading/trailing** ellipsis gets the same check against the hidden *same-clause* context (up to the nearest `.;:!?\n`), closing the trivial bypass of moving the ellipsis to the quote's edge.

`backend/src/lib/chat/verifyCitations.test.ts` — 11 new tests. Eight **attack** the matcher (inversion, contraction, exception carve-out, reversed order, distant quilting, leading/trailing edge-truncation, repeated fragment); all eight verified as `true` under the old implementation. Three guard against over-blocking (benign in-order omission, negator safely in a *previous* sentence, backtracking to a later anchor).

## Reproduce the bug on `main`

```bash
git checkout main
git checkout olp-pr/citation-quote-ordering -- backend/src/lib/chat/verifyCitations.test.ts
npm test --prefix backend -- src/lib/chat/verifyCitations.test.ts
```

Expected: **8 failed | 27 passed** — every `rejects …` test fails because `main` verifies the abuse quotes (run performed on `1b58c7aa`; output in the demo below).

## Verify the fix on this branch

```bash
git checkout olp-pr/citation-quote-ordering
npm test --prefix backend -- src/lib/chat/verifyCitations.test.ts
```

Expected: **35 passed** — all 24 pre-existing tests unchanged and green (legitimate abbreviated quotes still verify), all 8 attacks rejected.

## Demo

Live recording of the replication above — the identical test file run against `main`'s implementation (8 attack quotes verify → 8 failures) and then against this branch (all 35 pass):

PR #392 demo: adversarial tests vs main, then vs this branch

## Testing performed

- `npm test --prefix backend -- src/lib/chat/verifyCitations.test.ts`: 35/35 pass on this branch; the 8 new rejection tests fail on `main`'s implementation (red-before-green demonstrated by swapping only the source module).
- Full `npm test --prefix backend`: 68 files / 851 tests pass, 4 stack-gated files skipped as usual.
- `npm run build --prefix backend` (tsc): clean.

- **The inversion guard is a word-list heuristic, biased conservative.** A false positive only *withholds* the badge from an abbreviation a human should eyeball (e.g. omitting a harmless "without prejudice to clause 4" now yields unverified); a false negative would endorse a misquote. Given the badge's meaning, the asymmetry favors withholding.
- **`MAX_OMITTED_SPAN_CHARS = 500` is a calibration, not a derivation** — roughly two long legal sentences. Legitimate omissions of an entire long paragraph now come back unverified; reviewers may prefer a different constant, and the constant is the single knob.
- **Clause boundaries are approximated** as `[.;:!?\n]` for the edge-ellipsis checks. Qualifiers that *limit* rather than negate ("only", "provided that", "to the extent") are not in the word list, and a negator beyond an abbreviation like "cl. 9" is cut off by the period — residual risk accepted to keep false positives down.
- **Cross-page `[[PAGE_BREAK]]` segments are still verified per-segment without cross-segment ordering.** Deferred deliberately: the sentinel is constructed by our own viewer expansion rather than free-form model text, so the abuse surface is far smaller — but it is the same defect class, and I'll file it as a follow-up rather than widen this PR.
- **No char offsets for ellipsis quotes** (unchanged from before): offsets remain attached only to contiguous single-segment quotes.
- **Perf**: backtracking re-normalizes the source suffix per attempt (O(n) each). Bounded by the 64-attempt budget; typical quotes use 2–4 attempts.

https://claude.ai/code/session_01S4VRDD1DZat4SLYtt2q2eU

### DIFF
--- a/backend/src/lib/chat/verifyCitations.test.ts
+++ b/backend/src/lib/chat/verifyCitations.test.ts
@@ -137,6 +137,132 @@ describe("verifyQuoteAgainstSource", () => {
     expect(v.verified).toBe(false);
   });
 
+  // ── Adversarial ellipsis abuse ─────────────────────────────────────────
+  // These tests ATTACK the matcher instead of confirming it: each quote is
+  // built from real source words, so a matcher that merely finds every
+  // fragment somewhere would happily verify all of them. The original
+  // implementation did exactly that (each segment located independently),
+  // so every rejection case below verified as true before this fix.
+
+  it("rejects an ellipsis that swallows a negator and inverts meaning", () => {
+    const src =
+      "Under clause 9, the landlord is not permitted to terminate this lease during the fixed term.";
+    // Reads as "the landlord IS permitted to terminate" — the omitted word
+    // is "not". A verified badge here endorses the opposite of the source.
+    const v = verifyQuoteAgainstSource(
+      src,
+      "the landlord is ... permitted to terminate",
+    );
+    expect(v.verified).toBe(false);
+  });
+
+  it("rejects an ellipsis that swallows a negating contraction", () => {
+    const src = "The guarantor isn't liable for the tenant's costs.";
+    const v = verifyQuoteAgainstSource(
+      src,
+      "The guarantor ... liable for the tenant's costs",
+    );
+    expect(v.verified).toBe(false);
+  });
+
+  it("rejects an ellipsis that swallows an exception carve-out", () => {
+    const src =
+      "The deposit shall be returned in full except where the tenant is in breach.";
+    const v = verifyQuoteAgainstSource(
+      src,
+      "The deposit shall be returned in full ... the tenant is in breach",
+    );
+    expect(v.verified).toBe(false);
+  });
+
+  it("rejects fragments stitched together in reverse document order", () => {
+    const src =
+      "The alpha clause applies to assignments. The beta clause applies to subletting.";
+    const v = verifyQuoteAgainstSource(
+      src,
+      "beta clause applies to subletting ... alpha clause applies to assignments",
+    );
+    expect(v.verified).toBe(false);
+  });
+
+  it("rejects fragments quilted from distant parts of the document", () => {
+    const filler = "Unrelated boilerplate filler sentence follows here. ".repeat(
+      15,
+    ); // ~795 chars — far beyond any plausible single omission
+    const src = `The tenant shall maintain the garden. ${filler}The landlord shall insure the building.`;
+    const v = verifyQuoteAgainstSource(
+      src,
+      "The tenant shall maintain the garden ... The landlord shall insure the building",
+    );
+    expect(v.verified).toBe(false);
+  });
+
+  it("rejects a leading ellipsis that truncates away a same-clause negator", () => {
+    const src =
+      "Under clause 9, the landlord is not permitted to terminate this lease.";
+    const v = verifyQuoteAgainstSource(
+      src,
+      "... permitted to terminate this lease",
+    );
+    expect(v.verified).toBe(false);
+  });
+
+  it("rejects a trailing ellipsis that truncates before a same-clause carve-out", () => {
+    const src =
+      "The tenant may assign this lease unless the landlord objects in writing.";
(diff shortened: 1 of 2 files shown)


willchen96:
Closing this as this deterministic method can lead to false positives (e.g. the negative does not relate to the statement that is quoted or there is a double negative). The AI generated quotes are also immediately verifiable by the user in the assistantSidePanel when clicked on.

PR #406, opened 2026-08-31
fix: correct blob mock so mikeApi tests pass on Node 22 and 24

### LINKED ISSUE #361: fix: mikeApi.test.ts blob tests fail locally on Node 24 (blob.text is not a function)
## Problem

4 tests in `frontend/src/app/lib/mikeApi.test.ts` fail when run locally with Node 24.18.0, but pass in CI (which runs Node 22). Confirmed pre-existing on unmodified `main` at 31ee5da — not a regression from any specific PR. Observed independently across at least 3 recent PRs (#325, and the two #322 attempts), each time reasoned away as "local-only, don't file" — tracking it now since it keeps recurring and the skill's review checklist requires pre-existing failures to be tracked, not just noted per-PR.

## Failing tests

- `blob requests (exportAccountData) > returns the blob and the filename from content-disposition`
- `audit history > exports with the same active filters and server-side sort`
- `downloadDocumentsZip > POSTs the document ids and returns the blob`
- `unwrapping and blob wrappers > exportChatData and exportTabularReviewsData hit their export routes`

All fail with the same error:
```
TypeError: blob.text is not a function
```

## Likely cause

These tests construct a mock `Response`/`Blob` and call `.text()` on it. Node 24's built-in `Blob`/`Response` (via undici) vs. Node 22's differ in how they interact with jsdom's polyfills in this test environment — likely a version-skew issue between the repo's pinned Node version (CI: 22) and a locally-installed Node 24, not a bug in the mocked test helper itself per se, but worth confirming.

## Fix direction

Either: (a) pin/document the expected Node version for running this test file locally (repo root `package.json` currently has no `engines` field), or (b) make the blob mock in this test file construct a real `Blob`/`Response` that behaves consistently across Node 22/24, so contributors on newer Node don't see spurious local failures.

## Validation

`cd frontend && npx vitest run src/app/lib/mikeApi.test.ts` should pass locally regardless of local Node version, matching CI.

bmersereau:
## Summary

Fixes 4 tests in `frontend/src/app/lib/mikeApi.test.ts` that fail locally on Node 24 with `TypeError: blob.text is not a function`, while passing in CI (Node 22).

## Root cause

The tests already construct real `Response` objects (`new Response(...)`), so the request/response mocking itself was not the problem. The actual cause is in the shared Vitest jsdom test environment setup:

- Vitest's built-in jsdom environment always overrides `globalThis.Blob` with jsdom's own `Blob` implementation, which has never implemented `.text()`, `.arrayBuffer()`, or `.stream()` — only `slice`, `size`, and `type`.
- Node's built-in `Response.prototype.blob()` resolves its `Blob` constructor dynamically from `globalThis` at call time. On Node 24 this means `response.blob()` returns a jsdom-shaped blob missing those methods; on Node 22 it apparently does not exhibit this behavior, so the same tests pass there.
- Since the repo's own `package.json` pins `engines.node: ">=22"` with no upper bound, Node 24 is a legitimately supported local dev version, so this was a real (if narrow) bug, not just an unsupported-toolchain situation.

## Fix

Restore the real Node `Blob` (imported directly from `node:buffer`, independent of the `globalThis` override jsdom performs) in `frontend/vitest.setup.ts`, which runs before every test file. This makes `response.blob()` behave identically on Node 22 and Node 24, and does not touch any production code path.

## Changes

- `frontend/vitest.setup.ts`: restore `globalThis.Blob` to the native `node:buffer` `Blob` after jsdom's environment setup overrides it, with a comment explaining why.

## Test plan

- [x] `cd frontend && npx vitest run src/app/lib/mikeApi.test.ts` — 178/178 pass (previously 174/178, with the 4 tests named in #361 failing).
- [x] `cd frontend && npx vitest run` (full frontend suite) — 98 files / 641 tests, all pass — confirming no regression from the global `Blob` change on other tests that rely on `File`/`Blob`/`FormData` (upload flows, clipboard export, etc).
- [x] `cd frontend && npx tsc --noEmit` — clean.
- [x] `cd frontend && npx eslint vitest.setup.ts` — clean.
- [ ] `npm run build --prefix frontend` was not run in this environment: this worktree's `node_modules` is a symlink to a sibling worktree (to avoid an `npm install` under a non-pinned local Node version per repo policy), and Turbopack refuses to build through a symlink that points outside its detected project root. Confirmed this Turbopack error is pre-existing and unrelated to this change by reproducing it identically with the fix stashed out. `tsc --noEmit` covers the type-checking this build step would otherwise gate.

Closes #361

### DIFF
--- a/frontend/vitest.setup.ts
+++ b/frontend/vitest.setup.ts
@@ -1,1 +1,13 @@
 import "@testing-library/jest-dom/vitest";
+import { Blob as NodeBlob } from "node:buffer";
+
+// Vitest's jsdom environment always overrides `globalThis.Blob` with jsdom's
+// own Blob implementation, which never implemented `.text()`/`.arrayBuffer()`/
+// `.stream()` (it only has `slice`, `size`, `type`). On Node 24 (unlike
+// Node 22), the built-in `Response.prototype.blob()` resolves its `Blob`
+// constructor dynamically from `globalThis` at call time, so blobs returned
+// from a real `Response` end up jsdom-shaped and missing those methods.
+// Restoring the real Node `Blob` here (independent of the global, via
+// `node:buffer`) keeps `response.blob()` usable the same way on every
+// supported Node version. See #361.
+globalThis.Blob = NodeBlob as unknown as typeof Blob;


bmersereau:
## PR Review: fix: correct blob mock so mikeApi tests pass on Node 22 and 24

### Summary
Single-file, 12-line fix to `frontend/vitest.setup.ts` that restores the native Node `Blob` (from `node:buffer`) as `globalThis.Blob` after Vitest's jsdom environment overwrites it with jsdom's incomplete `Blob` (no `.text()`/`.arrayBuffer()`/`.stream()`). This makes `Response.prototype.blob()` — which resolves its `Blob` constructor from `globalThis` at call time on Node 24 — return a fully-featured blob instead of a jsdom-shaped one. No production code touched. Closes #361.

### Risk Assessment
- **Risk Level**: Low
- **Blast Radius**: `vitest.setup.ts` runs before every frontend test file, so a bad global override could break any test touching `Blob`/`File`/`FormData`. Verified this is not the case (see below).
- **Rollback Plan**: One-line revert of the added block; no migrations, no runtime/production surface.

### Root Cause Verification
Read `vitest.config.mts`: `setupFiles: ["./vitest.setup.ts"]`, single setup file, `environment: "jsdom"`. Vitest runs environment setup (which installs jsdom's globals, including its `Blob`) before executing `setupFiles`, so overriding `globalThis.Blob` at the top level of `vitest.setup.ts` is the correct place to win the race — confirmed this is architecturally sound, not a coincidental fix.

Reproduced the bug directly rather than trusting the PR description:
- Local Node version in the review environment: `v24.18.0`.
- Temporarily reverted `vitest.setup.ts` to its pre-fix content (`git show <main-tip>:frontend/vitest.setup.ts`) in an isolated worktree and ran `npx vitest run src/app/lib/mikeApi.test.ts`: **4 failures**, all `TypeError: blob.text is not a function` / `TypeError: result.blob.text is not a function`, exactly the tests and error named in #361.
- Restored the fix (`git checkout -- vitest.setup.ts`) and reran the same command: **178/178 pass**.
- CI (`.github/workflows/ci.yml`) pins `node-version: 22` for the frontend job — consistent with "passes in CI, fails locally on Node 24."

This confirms the mechanism as described, not just the symptom.

### Blast-Radius Verification (Blob/File/FormData elsewhere in the suite)
- `grep -rn "instanceof Blob\|instanceof File" src` → **no hits**. This matters: jsdom's `File` class extends jsdom's own `Blob` at class-definition time, so if any code did `x instanceof Blob` after the global reassignment, jsdom-constructed `File` instances could fail that check (prototype chain baked in before the override). Absence of any `instanceof` usage removes this risk entirely.
- Files constructing `Blob`/`File`/`FormData` in tests or src: `history/page.test.tsx`, `AssistantMessage.tsx`, `SpreadsheetView.tsx`, `exportToExcel.ts`, `NewTRModal.test.tsx`, `workflowZipExport.ts`, `documentDirectoryUpload.test.ts`, `documentUploadValidation.test.ts`, `mikeApi.test.ts`/`mikeApi.ts`. Spot-checked `documentUploadValidation.test.ts`: uses `new File(["content"], name)` (unaffected — `File` itself isn't reassigned) and only checks `.type`/`.size`/`.name`, not `instanceof`.
- Ran the **full frontend suite myself** (did not just trust the PR's "641 tests pass" claim): `npx vitest run` in an isolated worktree → **98 files / 641 tests, all pass**. Matches the PR's reported numbers exactly.
- `npx eslint vitest.setup.ts` → clean.
- `npx tsc --noEmit` → clean (0 errors).

### Review by Category

#### Correctness
- [severity:praise] Root cause is correctly diagnosed and the fix targets the actual mechanism (globalThis race between jsdom env setup and setupFiles), not a symptom-level mock rewrite in the test file itself.

#### Security
- N/A — test-infrastructure-only change, no production code path touched.

#### Performance
- N/A — negligible; one extra import + one global assignment per worker setup.

#### Code Quality
- [severity:nit] `globalThis.Blob = NodeBlob as unknown as typeof Blob;` uses a double cast because `node:buffer`'s `Blob` type isn't structurally identical to `lib.dom`'s `Blob` type. This is a reasonable/necessary escape hatch given the two types will never structurally unify, and it's well-commented — no action needed, just flagging why the double-cast exists in case a future reader wonders.

#### Test Coverage
- No new test needed — this is a fix to test infrastructure, and the existing 4 previously-failing tests in `mikeApi.test.ts` are the regression check. Confirmed they fail without the fix and pass with it (see Root Cause Verification above), so the existing suite already proves the fix.

#### Branch Hygiene / Issue Lifecycle
- 1 commit, conventional format, branch fresh off `main` (merge-base == origin/main tip).
- PR body contains `Closes #361` — correct closing keyword, issue will auto-close on merge.

### CI State (point-in-time snapshot, not blocking verdict)
At review time: `Backend build and tests` pass, `gitleaks` pass, `Eval harness` pass, `CodeQL`/`Analyze (javascript-typescript)` pass, `Fresh install vs upgraded deployment` pass, `license/cla` pass. `Frontend build and tests`, `Supabase stack integration tests`, and `playwright` were still pending/queued. Given the change is test-infra-only and the local full-suite run (641/641) plus targeted mutation-testing (revert → 4 fail, restore → pass) already prove the fix, I'm not blocking on these finishing.

### Suggestions (Non-Blocking)
- None beyond the nit above.

### Verdict
- [x] **Approve** - Ship it

### What I Verified
- [x] Read the full diff (`git diff origin/main..<branch> -- frontend/vitest.setup.ts`)
- [x] Ran tests locally: targeted (`mikeApi.test.ts`, before/after the fix) and full suite (98 files / 641 tests) in an isolated worktree
- [x] Checked related code not in diff (grep for `instanceof Blob/File`, other Blob/File/FormData call sites, `vitest.config.mts` setupFiles ordering, CI workflow Node version)
- [x] Reviewed linked issue (#361) and confirmed closing keyword present in PR body

willchen96 (approved):

willchen96:
Thanks for this @bmersereau

PR #423, opened 2026-09-04
fix(llm): replace the one shared 16k output cap with per-model ceilings

terpstra-209:
## Summary

`MAX_OUTPUT_TOKENS` is a single hardcoded 16,384 applied to every model, every provider and every request. This replaces it with a small, version-scoped table of the ceilings models actually accept, keeping 16,384 as the default for anything unlisted.

**Scope note, stated up front:** this is a headroom and correctness change, not a bug fix. An earlier revision of this PR claimed the shared cap was truncating long turns. Instrumented runs since then do not support that — see Testing below — and the claim has been removed from the description, the code comments and `.env.example`.

## Why the number is arbitrary

16,384 was never chosen as a cross-provider policy. Tracing it:

- It first appears in the OpenAI Responses adapter (`openai.ts`, commit `bef75b0`).
- The Anthropic adapter carried its own copy of the same figure, because `max_tokens` is a **required** field on Anthropic's API.
- `gemini.ts`, and the streaming paths in `ollama.ts` and `openrouter.ts`, sent **no output limit at all**. The only `max_tokens` in those two files was in their non-streaming title helpers (`?? 512`).
- The AI SDK migration (`4ac3a34`) replaced all five adapters with one `streamAiSdk` and applied the single 16,384 constant to all of them.

So a value that existed because one vendor mandates the field was extended to three providers that had previously sent nothing. There is no comment, doc or test in the repo explaining it. This PR restores per-model headroom to those paths.

## What changed

- `maxOutputTokensFor(provider, modelId)` in `backend/src/lib/llm/aiSdk.ts` replaces the shared constant at the `streamText` call site.
- Ceilings, each taken from the provider's published figure:

| model pattern | ceiling |
| --- | --- |
| `gemini-` (2.5 and 3, native or routed) | 65,536 |
| `qwen3.8` (flash, max, 27b) | 131,072 |
| `glm-5` through `glm-5.3` | 128,000 |
| `deepseek-v4`, `deepseek-v4.1` | 384,000 |
| everything else | 16,384 (unchanged) |

- `LLM_MAX_OUTPUT_TOKENS` overrides every built-in value, so a model that ships after this code can be given its ceiling without a patch.
- Router ids carry the upstream model, so one reached through OpenRouter, the Vercel gateway or a local OpenAI-compatible proxy matches the same entry as a native one.

Patterns are version-scoped deliberately. Ceilings vary *within* a vendor's lineup — `qwen3.5-35b-a3b` caps at 16,384 and `deepseek-r1` at 16,000, while `qwen3.8-flash` accepts 131,072 — and provider behaviour on an over-large value is not uniform: `@ai-sdk/anthropic` clamps to the model maximum and warns, while `@ai-sdk/google` and `@ai-sdk/openai-compatible` pass the value through to an upstream whose response I have not verified. Unlisted models keep the old value, so nothing regresses.

## Worth a maintainer's view

The AI SDK providers already know per-model ceilings themselves — `@ai-sdk/anthropic` substitutes the model maximum when `maxOutputTokens` is omitted, and clamps when it is too high. A plausible alternative to this table is to **stop sending `maxOutputTokens` entirely** unless the operator sets `LLM_MAX_OUTPUT_TOKENS`, letting each provider apply its own model-aware default. That is less code and cannot rot, at the cost of giving up an explicit backstop. Happy to rework it that way if preferred.

## Testing performed

- New unit tests in `backend/src/lib/llm/aiSdk.test.ts` cover each family natively and through a router, the `LLM_MAX_OUTPUT_TOKENS` override and its rejection of unusable values, and negative cases pinning `qwen3.5-35b-a3b`, `qwen3.7-flash`, `deepseek-r1` and `deepseek-v3.1-terminus` to the conservative default.
- `npm test --prefix backend` passes; `npm run build --prefix backend` is clean.
- Raw provider streams were captured (`RAW_LLM_STREAM_LOG_DIR`) across nine real runs on Gemini 3.8 Flash, DeepSeek v4.1 Flash and Qwen 3.8 Flash. **No run produced `finish_reason: "length"`;** the largest single-step completion was 12,036 tokens against the 16,384 cap. That is the evidence behind the scope note above, and the reason this PR no longer claims to fix a truncation.

### DIFF
--- a/backend/src/lib/llm/aiSdk.ts
+++ b/backend/src/lib/llm/aiSdk.ts
@@ -12,7 +12,42 @@ import type {
 } from "./types";
 import { createRawLlmStreamRecorder, logRawLlmStream } from "./rawStreamLog";
 
-const MAX_OUTPUT_TOKENS = 16_384;
+/**
+ * Conservative output ceiling, used for any model whose real ceiling we do not
+ * know. Kept low because exceeding a model's own limit is a hard provider
+ * error, not a truncation.
+ */
+const DEFAULT_MAX_OUTPUT_TOKENS = 16_384;
+
+/**
+ * Gemini's ceiling. Every Gemini 2.5 and 3 model accepts 65,536 output tokens
+ * (the figure @ai-sdk/google uses for its own thinking-budget math).
+ *
+ * This is not a nicety: Gemini counts *thinking* tokens against
+ * maxOutputTokens, so a ceiling sized for the prose alone lets a long
+ * deliberation consume the whole budget. The turn then ends with no text and
+ * no tool call — which surfaces as a silently empty answer, not as an error.
+ */
+const GEMINI_MAX_OUTPUT_TOKENS = 65_536;
+
+/**
+ * Router ids carry the upstream model, so a Gemini reached through OpenRouter
+ * or the Vercel gateway gets the same ceiling as a native one.
+ */
+const GEMINI_MODEL_ID = /(?:^|\/)gemini-/;
+
+/**
+ * A model that ships after this code does still needs a way to raise its own
+ * ceiling. LLM_MAX_OUTPUT_TOKENS overrides every value here.
+ */
+export function maxOutputTokensFor(provider: Provider, modelId: string): number {
+  const override = Number(process.env.LLM_MAX_OUTPUT_TOKENS);
+  if (Number.isSafeInteger(override) && override > 0) return override;
+  if (provider === "gemini" || GEMINI_MODEL_ID.test(modelId)) {
+    return GEMINI_MAX_OUTPUT_TOKENS;
+  }
+  return DEFAULT_MAX_OUTPUT_TOKENS;
+}
 
 /** Ensure a proxy-closed final SSE event is still visible to SDK parsers. */
 export async function aiSdkFetch(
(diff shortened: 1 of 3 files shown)


backend/src/lib/llm/aiSdk.ts
```diff
@@ -275,7 +338,7 @@ export async function streamAiSdk(
       system: params.systemPrompt,
       messages: params.messages,
       tools,
-      maxOutputTokens: MAX_OUTPUT_TOKENS,
+      maxOutputTokens: maxOutputTokensFor(config.provider, config.modelId),
```
DIYA73, line 341: The LLM_MAX_OUTPUT_TOKENS env override looks useful, but should there be a warning log when it's set to a value that exceeds a known model's ceiling? Right now it overrides silently — someone could set it too high for Anthropic models without realizing the provider is clamping it.

willchen96:
Thanks @terpstra-209. I would prefer the approach below. Grateful if you can submit a PR. Appreciate your contributions, let me know if you would like to join the discord and would like to get connected on socials. 

> The AI SDK providers already know per-model ceilings themselves — @ai-sdk/anthropic substitutes the model maximum when maxOutputTokens is omitted, and clamps when it is too high. A plausible alternative to this table is to stop sending maxOutputTokens entirely unless the operator sets LLM_MAX_OUTPUT_TOKENS, letting each provider apply its own model-aware default. That is less code and cannot rot, at the cost of giving up an explicit backstop. Happy to rework it that way if preferred.

backend/src/lib/llm/aiSdk.ts
willchen96: Thanks for flagging @DIYA73 - closing this PR (see my comment below).

PR #428, opened 2026-09-04
fix: review fixes for organization access (#426)

amal66:
Review-fix child of #426, based on `feat/organization-access` so the diff reads as fixes on top of the PR. Draft on purpose.

## Summary

#426 was reviewed the way the earlier org PRs were: five read-only review agents over the diff, then the branch booted on an isolated local stack (upgrade-path database: `main` schema plus the two PR migrations, applied twice), then every serious finding replicated in Chrome or via the API, then fixes on this branch, then every replicated finding re-run against the fixed code. The evidence report is linked in the review comment.

Verdict on #426 as pushed: the model is coherent and the org lifecycle, last-admin guard, deny hiding and inherited access all work as described. It should not merge as is because of four blockers that were replicated live and three that were verified in code, all fixed here. The remaining items are product decisions, listed at the end.

## Replicated live, then fixed

| # | What a user experiences on #426 | On this branch |
|---|---|---|
| 1 | An org member **denied** on a matter opens "Add documents", types part of the matter or file name, and sees the matter's name and the document's filename, size and date in the picker, and can attach it. Bytes were still refused. | The picker's search applies the same deny verdict as every other route. Denied member sees nothing from the walled matter; owner and org admins still see everything. |
| 2 | The **only admin** of an org presses "Delete account". The app signs them out and shows the login page. The background job fails forever (trigger `org_member_protect_resource_ownership`), and they log straight back in. | The route answers 409 before destroying anything: "You are the only admin of Frank Solo LLP. Make another member an admin, or delete the organization, before deleting your account." The settings page shows it and keeps the session. The job path refuses terminally instead of retrying 20 times. |
| 3 | New-project wizard with a file attached: if the grant step fails, "Project created, but access was not granted" appears; clicking Create again uploads the file **again** (measured 1 document before the retry, 2 after). | Grants run before uploads and completed work is remembered: 0 documents before the retry, 1 after, grant applied. Same class fixed for reviews (a retry created a second review) and workflows. |
| 4 | `scripts/audit-gate.mjs` prints "audit gate passed" and exits 0 when the registry answers 200 with `{"error":...}` or `{}`, and when the lockfile has no `packages` map (no network call at all). `main`'s version exits 1 in the same situation. | Fails closed again: error bodies and non-array values are rejected, an empty package list exits, and at least one batch must have been answered. `node --test scripts/audit-gate.test.mjs` pins it. |
| 5 | A **viewer** who can open a project creates a folder and is told "Project not found" (404) instead of being refused. Thirteen write routes. | 404 only when the project is invisible; otherwise 403 with an intentional message that the client shows verbatim. |

deny leak before/after

*Before: denied member Bob sees "Matter P" and `secret-merger-agreement.docx` in his document picker. After: the same search shows nothing.*

account delete before/after

*Before: Frank, sole admin, is back on the assistant page after "deleting" his account. After: the deletion is refused with the reason, session intact.*

wizard retry before/after

*Before: one Create click plus one retry left two copies of the file. After: one copy.*

## Verified in code, then fixed

- **Migration 02 destroyed every direct share on a review inside a project** and then dropped the column. Those recipients are now archived to `tabular_review_legacy_shares` before the drop (see design decisions).
- **Heir auto-promotion**: deleting the sole admin promoted the earliest-joined member, and the DB trigger then erased that member's deny overrides, making a walled-off person an owner of the walled matter with no audit row. Removed; see the 409 above.
- **Zero-member org**: sole admin, no members, org owns content: the org was left with nobody in it, invisible and undeletable. Now refused.
- **Eight new tables never revoked from `anon`/`authenticated`** in the migration (schema.sql had it). The drift CI cannot see ACL drift because `reset_public()` drops the default privileges; that CI gap is not fixed here.
- **workflow_shares email case**: the only grant column without a lowercase CHECK; list lowered both sides, detail did not, so a legacy mixed-case row listed then 404'd. Deduped, lowercased, CHECK added, comparison fixed.
- **Creator's own email backfilled as an editor grant** on their own project (reads as "Shared with 1 user"). Excluded.
- **Document context, audit trail and async export** still decided access by `documents.user_id`. All three resolve per document now.
- **DELETE /single-documents/:id** by `user_id` only: an org owner could not delete a detached document. Uses the role model now.
- **Workflow share revoke** ignored its result (DB error and unknown id both 204); org share batches wrote partially. Fixed.
- **Frontend**: self-row Remove was an instant lockout and self role change always failed; the deny list was always collapsed with no count; the delete-org confirm outlived its modal and still deleted; self-demotion kept the admin UI; a sent invitation was reported as failed when the refresh failed; a failed invitations fetch rendered as "you have no invitations"; project-chat gates were creator-based and contradicted the server in both directions; refused renames and deletes were unhandled rejections; shared-chat citation pills did nothing; a new project chat was stamped Owner; `ModalUI` stole focus from the auto-focused field one frame after every modal opened.
- **e2e**: `critical-path` and `chat-management` were never migrated to the three-step wizard and would fail every CI run; two locators became strict-mode ambiguous after the new aria-label. One shared `createProject()` helper now.

## Design decisions taken here (stated so they can be vetoed)

1. **Refuse, do not promote.** Deleting an account that is the sole admin of an org with members or content is refused with a named org. The alternative on #426 silently escalated a possibly-denied member. If auto-promotion is wanted, it needs an audit row and must skip anyone holding a deny override.
2. **Archive, do not widen.** Contained-review shares that have no home in the new model go to `tabular_review_legacy_shares` (service_role only). Promoting them to project-level viewer grants would widen a one-review share into the whole matter.
3. **A chat share does not share its documents.** Recipients of a standalone chat see the transcript; citation pills now explain the refusal instead of doing nothing. Widening document access along a chat grant is a product call.
4. **403 wording** is intentional and shown verbatim by the client.

## Not addressed here (product decisions for the maintainer)

- **Outside counsel cannot be given any access to an organisation project.** `POST /projects/:id/access` on an org project accepts org members only (400 "Only current organization members can be assigned a role") and `validate_direct_access_scope` refuses the row at the DB. That is the case the #267 review asked for ("outside counsel as a viewer without joining the org"); the new model has no representation for it.
- **Every org member's global chat list now contains every colleague's chat in every org project**, with no scope filter. Not a leak, but a large undocumented product change.
- The org access editor only exposes **owner** and **deny**; editor/viewer overrides the API accepts are invisible and unrevocable in the UI.
- `PillButton` `sm`/`normal` were resized, changing about 98 existing buttons including the Word add-in's accept/reject row.
- Select-all under an access filter selects only the loaded page; the "Shared" filter returns org rows and there is no "Organisation" filter.
- `validate_direct_access_scope` takes `FOR UPDATE` on the project row to read an immutable column, so sharing and renaming block each other.
- The document-mutation tool gate in the chat routes is unreachable (chat standing can never exceed project standing) and its tests mock an impossible state.
- The org member typeahead in the access editor is mouse-only.

## Full Chrome run on this branch (5 September 2026)

Thirteen flows driven in a real browser against this branch with five accounts (Alice admin/creator, Dave second admin, Bob member, Carol outsider, Frank sole admin of his own org), one recording per flow. Every flow behaved as the model says. Four small defects surfaced and are fixed in the round-two commits below.

| # | Flow | Result |
|---|---|---|
| 1 | Create org, invite Member and Admin |  |
| 2 | Recipients accept from the Invites tab |  |
| 3 | Sole admin tries to demote themselves: confirm, then Postgres refuses |  |
| 4 | Personal project sharing: own row read-only, Carol Viewer → Editor |  |
| 5 | Plain member creates a folder in the org matter (Editor by default) |  |
| 6 | Viewer: no Folder button, uploads disabled, delete refused with a contact |  |
| 7 | Outsider pastes the org matter URL: Project not found |  |
| 8 | Deny list (1); Bob's tab becomes Project not found; matter leaves his list |  |
| 9 | Shared chat: Viewer read-only with Shared marker, Editor can type, Share refused as owner-only |  |
| 10 | Org admin deletes a colleague's chat from the Chats tab |  |
| 11 | Standalone review shared as Viewer: Add Columns refused, "ask Alice" |  |
| 12 | Sole admin's account deletion refused with the reason, session kept |  |
| 13 | Wizard retry after a refused grant leaves one document |  |

Found during the run, fixed in round two:

- Add member modal listed an already-accepted invitation as Pending and the roster stayed stale until reload.
- A viewer's Upload menu disabled Saved files and Upload files but left Upload folder enabled.
- Deleting a colleague's chat from the project Chats tab had no confirmation.
- A review viewer could open and fill the New column modal; the "Editors only" refusal fired only on submit.

## Verification round (8 September 2026)

Four independent verifiers were pointed at the first twelve commits: a live backend attacker, a live migrations verifier, an audit-only reader of every commit, and a frontend verifier running every suite including the production build and the coverage ratchet. What they found is fixed in the last two commits, each pinned by tests shown to fail on the pre-fix source.

**Held under attack:** every deny-bypass vector tried (lists, detail, sub-resources, library search, chats, reviews, workflows, org inventory, audit, exports, upload sessions, forged download tokens, copy and import routes, cm_number and practice search, a direct grant inserted by SQL); the 403/404 split on the 13 fixed sites; account deletion in all five shapes plus a leave-then-delete race and a job inserted by hand; grant validation and the promote-clears-deny path; the workflow share batch, revoke semantics and the lowercase CHECK; fresh-versus-upgraded schema parity (identical fingerprints); the eight revokes proven effective against Supabase default privileges (TRUNCATE by `authenticated` succeeded before, is refused now); RLS on every org table; frontend typecheck, 1060 tests, lint, production build, and coverage floors.

**Found and fixed in round three:**

- A review viewer could start generation (cells written, audit event under the viewer); review chat, version delete and the non-project upload branch still answered 404 to a viewer who could see the resource.
- The sole-admin refusal inside the deletion job was retryable and ran after documents and storage were destroyed; the 409 sentence ignored the per-org reason.
- The document picker read overrides for every org project and still showed a creator who had left the org that org's project.
- Accessible project ids were unordered, unpaged and an N+1; the audit trail interpolated every reachable project id into one URL filter; document context asked one verdict per document; attach-from-documents omitted `org_id`; a failed creator-profile read was ignored.
- The org workflow share batch was a loop of upserts, not one write.
- The audit gate still passed on a registry answering `200 {}`; it now carries a canary package that must be echoed, OSV must answer every query, and CI runs the gate's own tests.
- Migration 02: creator exclusion now falls back to `auth.users`, the archive excludes self-shares, and the archive keeps its review id as a plain indexed column instead of a cascading foreign key.
- Frontend: sidebar still stamped Owner on project chats; the chat share modal could self-lock and lost the roster on a refused grants read; deleted documents in the owner's own chat were blamed on the sharer; two org roster fetches raced (a 3-member roster collapsed to 1 live); refresh failures were console-only; the delete confirm stayed live over a deleted org; a 4xx without detail leaked "API error: 409"; uploads keyed by filename; asset copies ran before grants; Back stayed enabled after a partial creation; viewers still saw enabled empty-state Upload and New subfolder; self-deny offered; deny list re-expanded; the Shared marker was under contrast; ConfirmPopup lacked a dialog role; one e2e locator was still ambiguous.
- Tests that passed on broken code were replaced with filter-aware stubs and `instanceof` assertions.

**Still open, with a reason:**

- The commit `7655a736` does not compile on its own (its chat.ts hunk lands one commit early); the branch builds at every later commit and the history was not rewritten to keep the review-fix diff stable.
- The last-admin trigger can reach zero admins under two concurrent demotions at `REPEATABLE READ`; the app runs `READ COMMITTED`, where it holds. Pre-existing in #426's migration 01; a one-line write to the org row would harden it.
- Two older migrations (`20260421_01_docx_editing`, `20260502_secure_user_api_keys`) never revoked their tables from `anon`/`authenticated` either, so the sentence "every earlier table-creating migration carries the revoke" in commit `818b7a4c` is overstated; the schema-drift job cannot see privilege drift because it drops the default privileges with the schema.
- `tabular_review_legacy_shares` is documented in `docs/deployment.md` but nothing in the product surfaces it to an upgrading operator.
- The design calls listed under "Not addressed here" are unchanged.

Gates at the tip: backend typecheck clean, 1448 tests passing (39 stack-gated skips); frontend typecheck clean, 1097 tests passing, lint 0 errors; add-in typecheck clean; audit-gate node tests 7/7; both migrations replayed twice on an upgraded scratch database and fingerprint-identical to a fresh install; `git diff --check` clean. Not re-run after round three: the Playwright suite and the production build (both green at ffbe6fff); the live stack was not rebooted after round three because the machine was under load, so round-three backend behaviour is pinned by route tests rather than re-driven in Chrome.

## Testing performed

- Backend: `npx tsc --noEmit` clean; `npx vitest run` 113 files / 1420 tests passed (6 files skipped: stack-gated).
- Frontend: `npx tsc --noEmit` clean; `npx vitest run` 148 files / 1060 tests passed (after round two); `npx eslint .` 0 errors.
- Word add-in: `npm run typecheck` clean.
- `node --test scripts/audit-gate.test.mjs` 2/2; stub registry returning `{"error":..}` now exits 1.
- Both migrations replayed twice on the upgraded database with `ON_ERROR_STOP=1`; a scratch fresh install of `schema.sql` compared object-by-object (functions, constraints, ACLs) against the upgraded database: identical.
- `git diff --check` clean.
- Playwright was not run here (needs the documented stack on :3100); the three touched specs typecheck.
- Full Chrome run of thirteen flows on this branch with five accounts (table above); the four defects it found are fixed in the last commit and one of them was re-verified live.

## Replication

**Base case on #426 (`28b8fb8a`)**, local stack, org with admin A and member B, org project P with one document, B denied on P via the Access modal:
1. As B: Assistant, "Add documents", Projects tab, type the file name. P and the file appear. (`GET /projects?view=directory-search&search=<name>` returns the row.)
2. As a user who is the only admin of an org that owns a project: Settings, Delete account, confirm. You land on the login page; `select status, last_error from db_jobs where kind='account.delete'` shows pending with "Database error deleting user"; log in again and it works.
3. New project, add a recipient on the Access step, attach a file, then `revoke insert on public.project_access_grants from service_role;` and click Create. "Project created, but access was not granted". `grant insert ...` back, click Create again. Two documents.
4. From `frontend/`: run a stub registry that answers `200 {"error":"nope"}` on :4599 and `npm_config_registry=http://127.0.0.1:4599 node ../scripts/audit-gate.mjs; echo $?` prints 0.

**On this branch**: step 1 shows nothing for B; step 2 shows "Account not deleted: You are the only admin of ..." and the session stays; step 3 leaves one document; step 4 exits 1 with "Both advisory services failed".

## Tradeoffs

- The 409 makes "delete my account" a two-step task for sole admins. That is the explicit alternative to silently handing the firm to somebody else.
- The archive table is one more object in the schema; it holds nothing on a fresh install and only what the old backfill would have thrown away on an upgrade.
- Grants-before-uploads means a refused grant now stops the wizard before any file leaves the browser; users who wanted the files regardless will retry once more.

https://claude.ai/code/session_01J5Ja9B2yyXEAByq8YZMexw

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -13,6 +13,23 @@ permissions:
   contents: read
 
 jobs:
+  # The gate's own regression suite. It asserts the FAIL-CLOSED contract —
+  # every bug this script has had was a path where "we could not check"
+  # printed the same "passed" as "we checked and it is clean" — and it runs
+  # against stub servers, so it needs no network and no install. Kept as its
+  # own job rather than a step in the matrix below so it runs once, and so a
+  # broken gate is legible as a broken gate rather than as four failing
+  # workspace audits.
+  audit-gate-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version: 22
+      - name: Audit gate self-tests
+        run: node --test scripts/audit-gate.test.mjs
+
   # Blocking for EVERY tree in the matrix below: high/critical advisories
   # fail the build unless their GHSA id is documented (with a reason) in
   # scripts/audit-allowlist.json — see scripts/audit-gate.mjs. Auditing
--- a/backend/migrations/20260904_01_organization_access.sql
+++ b/backend/migrations/20260904_01_organization_access.sql
@@ -708,10 +708,12 @@ AS $function$
       and (p_type is null or w.type = p_type)
   ),
   org_shared as (
-    -- Same org-membership arm as get_workflows_overview, including the
-    -- workflow_shares NOT EXISTS dedup, so a row visible via both routes
-    -- contributes its options exactly once. Tagged 'shared' to match the
-    -- overview's scope bucketing.
+    -- Same org-membership arm as get_workflows_overview. The `shared` arm
+    -- above takes only org_id IS NULL workflows and this one only org_id IS
+    -- NOT NULL, so the two are disjoint by construction and a row visible via
+    -- both routes still contributes its options exactly once -- no dedup
+    -- predicate is needed. Tagged 'shared' to match the overview's scope
+    -- bucketing.
     select w.practice, w.language, w.jurisdictions, 'shared'::text as source
     from public.workflows w
     where w.org_id is not null
@@ -1373,10 +1375,16 @@ AS $function$
     )
     when p_workflow_user_id::text = p_user_id then 'owner'
     else (
+      -- BOTH sides lowered. get_workflows_overview lists a share with
+      -- lower() on both, so a legacy mixed-case row listed for its recipient
+      -- and then 404'd the moment they opened it (and re-sharing produced a
+      -- second row rather than updating the first). The 02 migration
+      -- normalizes the stored values and adds a lowercase CHECK; this stays
+      -- symmetrical with the overview regardless.
       select s.role from public.workflow_shares s
       where s.workflow_id = p_workflow_id
         and coalesce(p_user_email, '') <> ''
-        and s.shared_with_email = lower(p_user_email)
+        and lower(s.shared_with_email) = lower(p_user_email)
       limit 1
     )
   end;
@@ -1770,6 +1778,23 @@ END $do$;
 CREATE INDEX IF NOT EXISTS idx_workflows_org ON public.workflows (org_id);
 DROP TRIGGER IF EXISTS workflows_cleanup_direct_grants ON public.workflows;
 CREATE TRIGGER workflows_cleanup_direct_grants AFTER INSERT OR UPDATE OF org_id ON public.workflows FOR EACH ROW EXECUTE FUNCTION public.cleanup_inherited_direct_grants();
+
+-- Every new table is service-role only, like every other table in this
+-- schema: the API is the only thing that reads them, and RLS ENABLE with no
+-- policy is not by itself a grant boundary — `anon` and `authenticated`
+-- inherit table privileges from PUBLIC unless they are revoked. schema.sql
+-- revokes all eight for fresh installs; without these lines an UPGRADED
+-- deployment ends up with a different, weaker posture than a fresh one for
+-- precisely the tables that decide who can see a firm's matters.
+-- Idempotent: revoking a privilege that is not held is a no-op.
+REVOKE ALL ON public.organizations FROM anon, authenticated;
+REVOKE ALL ON public.org_members FROM anon, authenticated;
+REVOKE ALL ON public.org_invitations FROM anon, authenticated;
+REVOKE ALL ON public.project_access_grants FROM anon, authenticated;
+REVOKE ALL ON public.project_org_access_overrides FROM anon, authenticated;
+REVOKE ALL ON public.chat_access_grants FROM anon, authenticated;
+REVOKE ALL ON public.tabular_review_access_grants FROM anon, authenticated;
+REVOKE ALL ON public.workflow_org_access_overrides FROM anon, authenticated;
 
 notify pgrst, 'reload schema';
 
--- a/backend/migrations/20260904_02_migrate_legacy_sharing.sql
+++ b/backend/migrations/20260904_02_migrate_legacy_sharing.sql
@@ -10,6 +10,29 @@
 
 begin;
 
+-- Archive for direct review shares the new model has nowhere to put. See the
+-- header comment on this table in schema.sql, and the review backfill below,
+-- for why they are neither converted nor discarded.
+create table if not exists public.tabular_review_legacy_shares (
+  id uuid primary key default gen_random_uuid(),
+  tabular_review_id uuid not null
+    references public.tabular_reviews(id) on delete cascade,
+  project_id uuid,
+  email text not null,
+  archived_at timestamptz not null default now(),
+  unique(tabular_review_id, email),
+  constraint tabular_review_legacy_shares_email_lowercase
+    check (email = lower(email))
+);
+
+create index if not exists idx_tabular_review_legacy_shares_review
+  on public.tabular_review_legacy_shares(tabular_review_id);
+
+alter table public.tabular_review_legacy_shares enable row level security;
+revoke all on public.tabular_review_legacy_shares from anon, authenticated;
+grant select, insert, update, delete
+  on public.tabular_review_legacy_shares to service_role;
+
 do $migration$
 begin
   if exists (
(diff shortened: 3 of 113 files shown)


amal66:
Superseded by #463, the same 14 commits rebased onto `main` (#426 merged on 5 September, so this PR's base is gone). The forward migration `20260912_02` replaces the in-place edits to the merged 0904 migrations. Leaving this open until the maintainer closes it.

PR #462, opened 2026-09-12
feat(i18n): add next-intl with English messages and pt-BR locale file

brauliogusmao:
## Summary

Adds internationalization support to the web frontend using `next-intl`, with **no behavior change for existing users**: the default locale is `en`, and every user-facing string in the frontend is now resolved from `messages/en.json` (which contains exactly the strings that were previously hardcoded).

## What changed

- **Scaffold**: `next-intl` plugin wired in `next.config.ts`; `src/i18n/request.ts` with locale `en`; `NextIntlClientProvider` in the root layout. No routing changes — URLs, links, and SEO are untouched (i18n without locale routing).
- **String extraction**: all user-facing strings across pages, components, and shared UI are resolved via `t()` from message files. Namespaces mirror the app's domains (`auth.*`, `projects.*`, `documents.*`, `workflows.*`, `tabular.*`, `organizations.*`, `settings` sections, etc.).
- **Messages**: `messages/en.json` holds the original English (source of truth); `messages/pt-BR.json` is included as a working example of a second locale.
- **Tests**: new `src/test/withIntl.tsx` helper wraps renders with the provider (pt-BR messages) — updated affected tests to assert through the message catalog.
- **Compat fixes needed by the extraction**: `ModalUI` gained optional `closeLabel` (default `"Close"`), `EditCardUI` an optional `labels` prop (English defaults), `AuthDividerUI` an optional `label` (default `"or"`) — all backward-compatible; `AccessEditor` now takes an explicit `resourceKind` prop instead of deriving the resource noun from a translated label (fixes a latent bug where the derivation depended on English label text).

## Why

This fork is maintained with a pt-BR interface. Keeping the translation as message files instead of hardcoded strings makes upstream syncs conflict-free and lets this i18n layer benefit the original project without any runtime cost for English users.

## Testing

- `npm run build --prefix frontend` — clean (36/36 pages)
- `npm test --prefix frontend` — 153 files / 1.060 tests passing
- Full `en.json` ↔ `pt-BR.json` key parity (script-verified), ICU args preserved

## Notes for reviewers

- `git diff main..HEAD -- ':!frontend' \* _is empty_`: backend and word-addin are untouched.
- The default locale stays `en`; a follow-up could add locale negotiation, but that is intentionally out of scope here.


willchen96:
@brauliogusmao thanks for this would you mind signing the CLA and rebasing?

PR #466, opened 2026-09-12
fix(citations): enforce document order for [[PAGE_BREAK]] segments

### LINKED ISSUE #393: Cross-page [[PAGE_BREAK]] citation segments are verified without cross-segment ordering
Follow-up to #391 / #392, deliberately left out of that PR to keep it single-purpose.

`verifyQuoteAgainstSource` splits a `[[PAGE_BREAK]]` quote into page segments and recursively verifies each segment independently (`backend/src/lib/chat/verifyCitations.ts`). As with the ellipsis defect fixed in #392, nothing requires the segments to appear in document order — `"B[[PAGE_BREAK]]A"` verifies even when A precedes B — and nothing bounds their separation, so two fragments from unrelated pages can be badged as one contiguous cross-page quote.

Risk is materially lower than the ellipsis case: the sentinel is constructed by our own viewer expansion (`expandDocumentQuoteEntry`) rather than free-form model text, so a hostile arrangement requires the model to emit the literal sentinel itself. It can, though — the sentinel is just text in the annotation.

Suggested fix: reuse `locateSegmentsInOrder` from #392 across page segments with a generous gap bound (page furniture between segments — headers, footers, footnotes — argues against the tight ellipsis bound; no negation check needed for system-generated breaks). Requires threading located spans out of the recursive per-segment verification.

DIYA73:
Fixes #393

## What
Enforces document order across `[[PAGE_BREAK]]` segments in `verifyQuoteAgainstSource`.

## Why
The old path verified each segment independently with no position check, so `"B[[PAGE_BREAK]]A"` earned `verified: true` even when A preceded B in the document.

## How
Locate the first plain-text fragment of each segment as a positional anchor (splitting on ellipsis if needed) and reject any arrangement where a segment starts before the previous one ends. Segments containing ellipsis are handled recursively by the existing ellipsis path. Mirrors the fix from #392.

## Tests
All 1520 tests pass.


willchen96 (approved):

willchen96:
Thanks @DIYA73

PR #463, opened 2026-09-12
[Orgs fix] Access follow-up on main (supersedes #428): 51 defects traced to #426, replicated and verified

amal66:
**Explainer** (each bug, why it happened, the fix, guardrails, proof): https://claude.ai/artifact/7J74jMgyDeaRbeeGfw5uX6 · **15 September case file:** https://claude.ai/artifact/16G2drgVYEfr4mURFr8WeT · **the access model itself:** https://claude.ai/code/artifact/8cdcf417-d926-44c2-90a7-5499eb7ca079

Supersedes #428 and folds in child PR #481 (its nine review-fix commits are here). Rebased onto `main` `8a36c344` on 17 September after #295's modularization; tip `ef304e0d`, 27 commits.

## Summary

#426 gave Mike organizations, two org roles, three resource roles and a per-person deny override. The model is right; what shipped did not apply it in one place. This branch fixes **38 defects relative to `main`** (29 from #426, one split between #426 and #399, 8 older ones the new model exposed), plus 7 labelled improvements, 2 e2e repairs and a CI rider (B07). The evidence comment's atlas traces every entry to its introducing and fixing line. The two that matter most: an ethical-wall leak (a denied colleague could read a walled matter's name and file names out of the document picker) and an account deletion that destroyed the session, queued a job that failed forever, and left the firm's data behind.

**Second round (15 Sept):** one fail-open in the audit scope, three frontend majors, four leftover 404-for-a-viewer sites. Fixed with tests in child PR #481, whose commits are on this branch.

**Third round (16 Sept, `f2cccfa8` → `9cec171a`): two regressions inside this PR's own audit rework.** (a) `listAccessibleProjectIds` paged the *projects* but read the *deny overrides* unpaged, so past the server's row cap a walled matter entered the scope — and `pageProjectRows` ended its loop with a `break`, treating a truncated read as complete. Both now fail closed: the denies are paged with an explicit order, and exhausting the page budget throws. (b) The chunked audit read merged only the first page of each partition, so a page past ~1,000 rows came back **empty while the total still said 1,206**; `readEventWindow` reads through the window, and the sort gained an `id` tie-break so no row lands on two pages. Both are pinned by new cases in `access.test.ts` and `auditExport.test.ts`; (b) is shown live below.

### Scorecard against #426's own follow-up list

| #426 follow-up | On `main` | Here |
|---|---|---|
| Sole-admin deletion fails after signing the user out | Open (B02) | 409 naming the org; session kept; job fails terminally |
[...]

PR #481, opened 2026-09-15
fix: second-round review fixes for #463 (fail-closed audit scope, one-layer Escape, wizard retry locks)

amal66:
Review fixes for #463, branched off its tip `1ca6957c`. Draft on purpose. Base is the PR branch so the diff reads as review fixes on top of #463; merge (or cherry-pick) into `olp-pr/org-access-followup-main` and the parent picks them up.

## Summary

A second, independent read of #463 (one backend reader, one frontend reader, both verifying every claim against the checked-out tree; all suites re-run) found one security-relevant fail-open, three frontend majors and a handful of leftovers of the classes #463 itself fixes. This branch fixes the mechanical ones with a regression test each. Design-level items are listed under "Not addressed" for the parent PR's discussion.

### Backend

1. **`listAccessibleProjectIds` fails closed** (`backend/src/lib/access.ts`). Every access read in the audit/memory scope helper ignored `error`, so an unreadable deny-override table widened a denied member's audit history to the walled matter; a failed project page silently truncated the firm. Each read now throws with the PostgREST error as `cause`; the picker already behaved this way. *Pinned:* 4 tests in `lib/__tests__/access.test.ts`.
2. **`queryEvents` catches that throw** and returns the `{ error }` shape GET `/audit` already maps to a generic 500 (the handler has no try/catch of its own; an unhandled rejection would hang the request). **Deep audit pages are clamped to the merge window** instead of being sliced from rows of unknown global order. *Pinned:* `routes/__tests__/audit.test.ts`; the existing `cause` test in `auditExport.test.ts` still passes.
3. **Two more 404→403 sites**: `POST /single-documents/:id/versions/from-document` for a viewer, and both workflow branches of the upload-session destination check. *Pinned:* `documents.routes.test.ts`, `uploadSessions.routes.test.ts`.
4. **Comments corrected**: DELETE `/single-documents/:id` no longer claims an org admin can remove a colleague's document (it cannot; the departed-uploader case is what the rule opens); `checkWorkflowAccess` attributes the lowercase CHECK to `20260912_02`, not `20260904_02`.

### Docs

5. **`docs/deployment.md`** said `20260904_02` populates `tabular_review_legacy_shares`. It does not (it is merged and unchanged, and dropped the column unrecorded). The note now says `20260912_02` creates the archive, that it lands empty on an already-upgraded deployment, and gives the backup-restore recipe.

### CI

6. **Audit gate**: a real package named `error`/`message`/`code` with an advisory (an array value) is judged, not mistaken for an error envelope. The test file's `respond` gets the `(url, body)` arity CodeQL asked for. *Pinned:* new case in `scripts/audit-gate.test.mjs`.

### Frontend

7. **One Escape, one layer** (`ConfirmPopup.tsx`). The popup's Escape listener and `ModalUI`'s both fired on one press, so declining "Delete organization?" also closed the settings modal and lost an unsaved rename. The popup now claims the key in the capture phase and stops it; the next press reaches the modal. The B26 test asserted the double-close as intended; it now asserts the modal survives, and a new case covers B26 via the Close button. *Pinned:* `ConfirmPopup.test.tsx`, `OrganizationWorkspace.test.tsx`.
8. **Wizards lock what a retry would drop.** `NewProjectModal` locks the name, client/matter number and practice area alongside the workspace once the project exists (a retry only re-persists the memory toggle); the stale "Back has to retire" comment is rewritten. `NewTRModal` disables Upload and freezes the document selection once the review exists (a retry reuses the review and never re-reads the document set) and says so in plain text; `FileDirectory` gains a `selectionDisabled` prop (default false). *Pinned:* `NewProjectModal.test.tsx`, `tabular-reviews/page.create.test.tsx`, `FileDirectory.test.tsx`.
9. **Citation refusal wording** (`ChatView.tsx`): the "the person who shared this chat has not shared its documents" copy is used only for a standalone chat; a project chat's role is the project role, so editors and viewers were being blamed on a sharer for a simply-deleted document. *Pinned:* `ChatView.citations.test.tsx`.
10. **Self-demotion confirmation** stays up with "Continuing..." until the role change settles instead of unmounting before the request starts. *Pinned:* `OrganizationWorkspace.test.tsx`.
11. **`RowActions.test.tsx`**: the three component tests the parent's rewrite dropped are restored next to the two new subfolder cases.

## Corrections to #463's own body/artifact, made today in the parent (not in this diff)

- B16 (async zip export drops `org_id`/`workflow_id`) had been withdrawn as "no hunk in the diff". It is in the diff (`handlers.ts:382-390`), real on `main`, and pinned by `dbq/__tests__/handlers.test.ts`. Reinstated as a verified defect.
- B31/B45/B49 were #463's own draft regressions, never on `main`; folded into their parent entries. Seven entries are labelled improvements rather than defects. Headline is now 38 defects relative to `main`, not 51.
- The wizard's Back button is not disabled after a partial creation; only the workspace selector locked. Fix 8 below extends that lock to the other identifying fields.

## Not addressed here (for the parent PR)

- `ChatAccessModal` re-issues `getChatPeople` and `getChatAccess` and flashes the roster skeleton because the roster effect keys on grant content (`AccessModal.tsx:88-131`). Needs a key on `resourceId` plus an explicit refresh counter.
- The disabled "New subfolder" row action and the empty-state Upload button carry their reason in `title`/nothing, unreachable by keyboard (`RowActions.tsx:132-147`, `DocTable.tsx:4061-4090`). AGENTS.md wants the reason as visible text.
- The document picker's org-project read is still unpaged (`projects.ts:571-577`); same truncation class the parent fixed in `listAccessibleProjectIds`.
- `auditExport.test.ts` never exercises the multi-chunk merge (count summation, dedup, window); every case runs the zero-projects branch.
- `compareRows` sorts in JS string order while each partition was top-N'd under Postgres collation; non-timestamp sorts can interleave differently.
- Sole-admin deletion refusal also covers a one-person org that owns any content; stated in the parent's decision 1 now, still a product call.
- Commit `d5c3444f` (CI audit gate) is unrelated to organization access and can be lifted into its own PR.

## Testing performed

| Command | Result |
|---|---|
| `npm test --prefix backend` | 132 files passed, 6 skipped; 1,665 tests passed, 39 skipped (parent: 1,655) |
| `npx tsc --noEmit` (backend) | clean |
| `npm test --prefix frontend` | 166 files, 1,196 tests passed (parent: 1,186) |
| `npx tsc --noEmit` (frontend) | clean |
| `npm run lint --prefix frontend` | 0 errors, 34 pre-existing warnings |
| `node --test scripts/audit-gate.test.mjs` | 8 passed (parent: 7) |
| `git diff --check` | clean |

Every new test was checked to fail on the pre-fix source (the component hunk reverted, the targeted file re-run). No stack was booted for this round: the fail-open (fix 1) is proven by the unit tests above, which inject a failed override read and assert the walled project is refused rather than listed; the parent PR's live recordings on `assets/pr463-demo-2026-09-15` stand for the rest.


amal66:
Folded into #463 on 16 September: all nine commits of this branch are now on `olp-pr/org-access-followup-main` (tip `452b954e`, rebased onto `main` `f024fe01`), so this child PR is closed as merged-by-fold. Nothing here is lost; the commit subjects are unchanged on the parent.

PR #486, opened 2026-09-16
[Chat fix] Fix leaving a thread detaches the streaming answer instead of cancelling it

amal66:
## Summary

**Regression introduced by #483** ("Add project chat IDE workspace and retain document viewer state", merged 2026-09-15 as squash commit `f024fe01`).

On `main`, switching to another chat inside a project workspace, or pressing **New chat**, while an answer is streaming destroys that answer: the stored assistant message ends mid-sentence with **"Cancelled by user."**, and because that truncated row is what every later load shows, the loss is permanent. This PR makes leaving a thread *detach* from the in-flight answer instead of cancelling it, and makes a reader who comes back to the thread pick the finished answer up. Only the Stop button aborts a request now.

**Review round 2026-09-20 (willchen96): returning to a thread now shows the answer streaming, not a loading state — see the section below. Rebased onto main `ec53dc94`; tip `e1ab4457`.**

**Rebased onto main `8a36c344` on 17 September, after #295** (the backend modularization, which also rewrote this hook's SSE loop). The conflict in `useAssistantChat.ts` was resolved main-wins: the shared `lib/sse.ts` reader stays, and the detach logic was re-applied on top of it. Tip: `40ca10ad`. Live A/B and the fail-without-fix pin were both re-run against this base.

## Where the bug came from

Before #483 the only client-side abort in `useAssistantChat` was the Stop button, and New chat in the project workspace was a `router.push` (a remount, which never touched the socket). #483 added the in-place chat switcher and, with it, every abort path this PR removes. `git blame` on `f024fe01` attributes each line below to that squash commit; none of them exist in `aba3cfca`, the commit before it.

| Abort path on `main` | Added by #483 as | Before #483 |
|---|---|---|
| Hook's thread-switch `useLayoutEffect` calls `abort()` | new effect: the hook did not react to `chatId` changes | Standalone pages remount on switch; nothing aborted |
| Workspace `navigateToChat()` calls `cancel()` (the Stop handler) before switching threads | new function, added with the switcher | Did not exist |
| `handleNewChat` → `resetChat()` → `cancel()` | `resetChat` is new in #483 | `handleNewChat` was `router.push`, no cancel |
| Superseded turn leaves the read loop, cancelling the reader | new request-generation guard (#295 moved it into `lib/sse.ts`, which cancels on any early exit) | No generation guard, no stale loop |

The backend side (`res.on("close")` → persist as "Cancelled by user.") predates #483 and is unchanged here; it is the correct contract for Stop.

## What changed

- **`useAssistantChat`**
  - New `detach()`: retires the request generation, drops the controller reference and clears the loading flags, **without aborting**. `resetChat` (New chat) uses it. The thread-switch layout effect no longer aborts either.
  - A superseded turn no longer breaks out of the `readSseFrames` loop — breaking out cancels the reader, which the backend reads as Stop. It sets a `superseded` flag and keeps consuming frames, discarding all but `chat_id`, so the socket stays open until the server has stored the answer. Stop is the one exit that still aborts (`readSseFrames` throws on its signal).
  - New `lib/assistantTurns.ts`: a module-level registry of in-flight turns keyed by chat id. Each turn record **owns the assistant message while it streams** (`beginAssistantTurn` / `identify` / `update` / `finish`, plus `cancel` for Stop and `subscribe` for readers). `getAssistantTurn(chatId)` hands the live record to any hook viewing that chat; `withLiveTurn(messages, turn)` lays it over a transcript (matching the stored row by id, the sender's own placeholder, or appending). `loadAssistantChat(chatId)` fetches history at once — it only re-reads when a turn starts or finishes during the GET, and waits (bounded to 5 s) for a row that Stop closed the socket ahead of.
  - The hook's frame handling writes into a per-turn event sink (the handler bodies are unchanged) instead of the hook's own state. Every hook viewing the chat — the sender, a reader who switched away and back, or a freshly mounted page — attaches to the record and mirrors it into `messages`; hosts' `setMessages` is wrapped so a history snapshot can never displace the answer streaming into the thread, and the finished record stays attached until the thread changes so a stale read cannot erase it either.
  - `isResponseLoading` also reports a detached turn for the chat being viewed (`useSyncExternalStore`), so a second turn cannot start on top of the first.
- **Both chat pages** load history through `loadAssistantChat`; the workspace's `navigateToChat` calls `detach()` instead of `cancel()`; the standalone page reloads per chat id.
- **`ChatInput`**: Stop stays enabled while the composer is closed, so a returning reader can stop a detached turn.
- **`ChatView`**: new required `detach` prop for New chat.
- **`ChatInput` names the reason it is closed.** `canSend` means permission again; a new `chatLoading` prop (forwarded by `ChatView`/`ChatInputPrompt`) carries "the history has not arrived". It says "A response is still arriving…" while a detached turn runs, "Loading this chat…" otherwise, and keeps the read-only copy for a missing grant.
- **Tests**: `useAssistantChat.lifecycle.test.tsx` (7 cases), `assistantTurns.test.ts` (8), 3 workspace page cases, `ChatInput.canSend`, `ChatView.actions`.

## Why

The backend treats a closed response socket as the user pressing Stop: it registers `res.on("close")`, stops the model and persists the partial text labelled "Cancelled by user." That is correct for Stop. It cannot distinguish a navigation from a Stop, so the client has to — and the Stop control is the only place that should close the socket.

## Reproduce the bug on `main` (8a36c344)

Setup: a model that streams slowly enough to switch chats mid-answer. The stub used below (OpenAI-compatible, pointed at `OLLAMA_BASE_URL`, 600 ms per chunk) gives a ~19 s answer; a real model with a long prompt works too.

1. Build and run `main` (`npm run build` in `frontend/`, then `API_BASE_URL=http://localhost:<be> npx next start -p 3020`), log in, pick the slow model.
2. Open chat **A** in a project workspace (`/projects/<id>/assistant/chat/<A>`).
3. Send "summarize the agreement". While it streams, open the switcher at the top of the chat pane and pick chat **B**.
4. Wait for the stream to finish, then reload chat A.

**Observed on main:** the assistant row ends with `Cancelled by user.` — **480 chars** vs **605** for the same answer left alone, cut at "…the governing law is the State o". Backend log: `[project-chat/stream] client aborted stream { chatId: <A> }`. **New chat** mid-stream does the same (**220 chars**). Coming *back* to A while it should still be streaming shows the same dead row (**280 chars**) — the answer is already gone.

main: switching chats mid-stream truncates the answer

main: stored answer ends with Cancelled by user.

main: New chat mid-stream

main: returning mid-stream finds a dead row

## Verify the fix on this branch (40ca10ad)

Same steps, same database, same stub, on a production build of this branch.

| Flow | main `8a36c344` | this branch `40ca10ad` |
|---|---|---|
| (a) switch chats mid-stream | 480 chars, "Cancelled by user." | **616 chars**, complete, ends "…fixed for the initial term." |
| (b) New chat mid-stream | 220 chars, "Cancelled by user." | **615 chars**, complete |
| (c) leave, then return while still streaming | dead 280-char row | the stored transcript **plus the answer streaming token by token**, Stop available, 0.4 s after returning (review round below; the 17 Sep version showed a loading state until the server had stored the answer) |
| (d) Stop response | (unchanged) | 220 chars + "Cancelled by user." — contract preserved |

Backend `client aborted stream` lines over the session: **3 on main** (one per leave), **1 here** (the Stop in flow d).

fix: switching chats mid-stream keeps the answer

fix: returning mid-stream restores the finished answer

fix: the complete answer was persisted

The waiting composer, before and after `40ca10ad`:

before: the wait blamed edit access

after: the wait says a response is still arriving

fix: Stop still truncates and labels the answer

## Review round 2026-09-20: returning to a thread shows the answer streaming

The reviewer's finding: after navigating away from a streaming chat and back, the chat sat on a loading state until the stream finished. Cause: `loadAssistantChat` held the history GET open until the detached turn had been persisted, and the workspace renders its message shimmer (and no composer) while the history is outstanding.

**Before (tip `bf3770c8`, this branch on main `ec53dc94`) — project workspace.** Send in Chat A, switch to Chat B, come back 1.5 s later: shimmer for the rest of the answer (~17 s with the 700 ms/chunk stub), no Stop, then the whole answer at once.

before: shimmer until the stream ends

before: 10 s after returning

**After (tip `e1ab4457`).** Same steps, same stub, same database: 0.4 s after returning the transcript is on screen with the answer as far as it has got, Stop is available, and the rest streams in.

after: the answer keeps streaming on return

after: 0.4 s after returning

after: complete

**Standalone chat page** (sidebar switch remounts the page; a fresh mount attaches to the answer already streaming):

before: standalone page

after: standalone page

after: standalone, 0.4 s after returning

Measured from the recordings' page-text log: before, 0.4 s after returning, 4 shimmer bars, no Stop, 0 answer characters; after, Stop present and the answer growing (59 → 134 → 549 characters on the workspace, 100 → 176 → 597 on the standalone page). Stored row for the returned chat: complete, ends "…fixed for the initial term."; backend `client aborted stream` count over both runs: **0**.

Tests: 4 new `useAssistantChat.lifecycle` cases (switch away and back shows the live answer token by token; a fresh mount attaches; a history read that predates the finished answer cannot erase it; Stop from a returning reader labels the record), the workspace case now pins the streaming return, and `assistantTurns.test.ts` covers the record, `getAssistantTurn` and `withLiveTurn`. Fail-without-fix pin against `bf3770c8`'s production files: the 5 behaviour cases fail (no live answer on return), 45/45 pass with the fix. `tsc` clean, `npm run test:coverage` **202 files / 1480 tests**, thresholds met, `git diff --check` clean.

Rebase note: main's #463 made `canSend` tri-state (`null` = standing unknown, "Loading…"); kept, with `chatLoading` layered on top. Three `ChatView` test renders main added since gained the required `detach` prop.

- **Drain, not cancel, for superseded bodies.** Keeping the socket open costs the client the rest of the answer's bytes and a loop that discards them. The alternative — teaching the backend to tell navigation from Stop — would need a new protocol for something the client already knows.
- **Return-to-thread attaches to the live stream** (changed in the 2026-09-20 review round; the first version waited for the server to store the answer and showed a loading state until then). The client still holds the socket, so no server-side replay channel is needed: the turn record is the single owner of the streaming message and any hook that views the thread mirrors it.
- **A detached turn blocks a second turn in the same thread.** While a turn is registered for a chat, `handleChat` refuses and the composer is disabled. Two concurrent writers on one thread would be worse. The closed composer names that reason rather than borrowing the read-only copy, which is what the first version of this PR did — it told a reader who owns the thread they lacked edit access.
- **The 5 s persistence bound.** If Stop closes the socket just before the backend writes the cancellation row, a returning reader would adopt an empty row. `loadAssistantChat` retries for up to 5 s, then raises a load error rather than waiting forever.
- **`ChatView.detach` is required**, so every host decides what New chat means instead of falling back to cancel.
- **A new message in the same thread while one streams still aborts it** (unchanged from `main`).
- **Stop is unchanged** and still writes "Cancelled by user."
- **Not a revert of #483.** Only its cancel-on-leave behaviour was wrong.

## Testing performed

All numbers from a clean `npm ci` worktree on `8a36c344`.

- **Fail-without-fix pin, re-run against the new base.** Branch test files kept, all six production files checked out from `origin/main`: **12 of 73 fail on main's code, 73 of 73 pass with the fix.** The twelve: 4 lifecycle detach cases (thread switch and `resetChat` die with "Invalid state: Controller is already closed" — main's reader is already cancelled; `detach` is not a function; the unmount case still calls `router.replace`), 2 duplicate-turn cases (`isResponseLoading` false), 1 "stop a detached request from a newly mounted owner" (`signal.aborted` false), the `ChatInput` Stop-while-loading case (button disabled), the `ChatView` New chat case (`detach` not called), and all 3 workspace cases (`requestSignal().aborted` true, and `getChat` called for a chat whose turn is still running).
- **Second fail-without-fix pin, for the composer copy.** New tests kept, the five files `40ca10ad` touches checked out from `69db9d11`: **4 of 37 fail** (three `ChatInput` cases — the placeholder is "How can I help?" instead of "A response is still arriving…" / "Loading this chat…", and the textarea is not disabled — plus the workspace case, where the composer reports `data-can-send="false"` during the wait instead of `"true"`), **37 of 37 pass** with it.
- `npx vitest run` (full frontend suite): **186 files, 1271 tests passed.**
- `npm run test:coverage`: passes the 100 % / 99 % thresholds; `assistantTurns.ts` is at 100 % statements, branches, functions and lines.
- `npx tsc --noEmit` clean. `npm run lint`: **0 errors**, 36 warnings — `8a36c344` itself has 38; this branch drops two `eslint-disable` directives that are no longer needed and adds none.
- Live A/B on production builds of `8a36c344` and this branch against one database and one stub model (table and recordings above), with the stored rows verified by SQL and the backend logs checked for `client aborted stream`.
- `git diff --check` clean.
- **CI on `40ca10ad`: all 13 checks green**, including frontend, backend, playwright, the Supabase stack suite and the schema-drift check.

## Related

- **#483** introduced the regression; this PR keeps its workspace and switcher intact.
- **#295** (`4df3edd5`) replaced this hook's hand-rolled SSE loop with the shared `lib/sse.ts` reader, which cancels on any early exit — so the detach fix is expressed as "keep iterating" rather than "release the body by hand"; `releaseSupersededBody` is gone.

## Summary by CodeRabbit

- **New Features**
  - Assistant responses can continue generating when switching chats, starting a new chat, or navigating away.
  - Returning to a chat reconnects to its in-progress response and preserves streamed content.
  - Chat history now loads reliably before enabling message composition.
- **Bug Fixes**
  - Prevented stale responses from appearing in the wrong conversation.
  - Improved loading, permission, and active-response messaging.
  - Explicitly stopping a response now cancels it while navigation leaves it running.


amal66:
Fixed the detached-response regression in af991be9.

### Chrome reproduction

On original head `395ce297`, with two existing chats and a controlled delayed local Ollama-compatible response:
1. Send a message in one chat and wait for streaming to start.
2. Select the other chat from the sidebar, then return before the response finishes.
3. Release the delayed response.

The returned chat did not display the completed answer. The real local database contained the full persisted response, `Chrome reproduction: the detached answer completed.`, while Chrome still showed only the latest user prompt. This used the real app, API and persistence code; the local provider controlled only response timing/content.

### Chrome fix verification

Repeated the same sidebar navigation on the fixed head. Returning while the response was pending kept **Stop response** available and prevented a follow-up send. After releasing the response, the saved answer appeared automatically in the returned chat **without refreshing or navigating again**. The other chat continued to show its own history.

The fix tracks active turns across same-tab navigation, waits for persistence before adopting history, rejects stale loads and callbacks, and keeps the Stop action available while loading. Tests also cover remounts, delayed history reads, failure handling and cancellation.

Validation: **73 affected tests pass across 8 files**, TypeScript (`tsc --noEmit --incremental false`) and targeted ESLint pass, and `git diff --check` passes. Three new revisit/remount regression tests fail against original head `395ce297` and pass with the fix.

Chrome before/after screenshots were captured locally. GitHub's browser attachment flow stalled, so they are not attached to this comment.

willchen96:
Thanks for this @amal66. However, it would be better for the detached chat to show the messages including the latest response streaming when the user navigates back. Currently when the usage navigates away from a chat and comes back, the chat is stuck in a loading state until the stream finishes.

Screenshot 2026-09-19 at 7 39 43 PM

amal66:
Thanks @willchen96 — agreed, and fixed in `e1ab4457` (rebased onto main `ec53dc94`).

The loading state came from `loadAssistantChat` holding the history GET open until the detached turn had been persisted; the workspace shows its shimmer while the history is outstanding. The turn registry now owns the assistant message while it streams, and any hook viewing that chat (the sender, a reader who switched away and back, or a freshly mounted page) attaches to the record and mirrors it. History is fetched immediately and the live turn is laid over it, so coming back shows the stored transcript plus the answer streaming token by token, with Stop available.

**Before** (your screenshot, reproduced on the previous tip): shimmer until the stream ends.

before

**After**: 0.4 s after returning the transcript is on screen and the answer keeps arriving.

after

after, 0.4 s after returning

Same on the standalone chat page (sidebar switch remounts the page):

after, standalone

Stored row for the returned chat is complete and the backend logged no aborted stream in either run. Tests: 4 new hook cases (switch away and back, fresh mount attaches, stale history read keeps the finished answer, Stop from a returning reader) plus the updated workspace case; these fail against the previous tip. Details and the standalone before/after are in the PR body under "Review round 2026-09-20".

willchen96 (approved):

willchen96:
Resolved the CodeRabbit findings and verified the fixes against the current branch:

- **Block file drops while chat history is loading** — resolved in `6dabab7b`. File drops now use the composer-open guard, and global drop listeners are disabled while the composer is closed.
- **Discard upload updates after the chat generation changes** — resolved in `d5e09e5f`. `ChatInput` is keyed by `chatKey`, and late upload progress, success, failure, warning, and cleanup callbacks from the previous chat are ignored. Regression tests cover late success and failure after switching chats.
- **Streaming status for the selected chat** — updated in `fa7cf241`. Both selected and detached streaming chats show the loader; only detached chats retain the green completion state.

Verification is green: focused frontend tests, ESLint, TypeScript, `git diff --check`, the full frontend/backend CI suites, Playwright, Supabase integration, schema drift, security scans, and CodeRabbit. CodeRabbit's latest review reports no actionable comments.

willchen96:
### Additional implementation notes

This change treats an in-progress assistant response as a chat-scoped background turn instead of tying it to the currently mounted page.

#### Turn lifecycle

- Navigating away from a chat no longer cancels its active response.
- The live turn is stored by chat ID, so multiple chats can continue streaming independently while the user moves between threads.
- Returning to a chat while its response is still running reattaches the live assistant message and cancellation handler to that page.
- Stored chat history and the live overlay are reconciled so the in-progress response is not duplicated or lost when history finishes loading.

#### Navigation and positioning

- Reopening a detached streaming chat waits for its complete stored history before positioning the transcript.
- The viewport is positioned at the latest user question using scroll-container-relative geometry, instead of briefly showing the first message at the top.
- Stale scheduled scroll work is cancelled when the active chat changes.

#### Composer and uploads

- The composer remains in its streaming state when a user returns to an active response.
- File drops are blocked while chat history is loading or the composer is otherwise closed.
- Composer/upload state is isolated by `chatKey`. Async callbacks from an upload started in a previous chat cannot update the newly selected chat or attach the old document to its next message.

#### Sidebar feedback

- Every actively streaming chat shows a spinner, including the chat currently being viewed.
- A detached chat changes to the green completed state when its response finishes and keeps that state until opened.
- A response that finishes in the currently selected chat returns directly to the normal chat icon; it does not show a redundant green completion state.

The added regression coverage exercises turn detachment and reattachment, returning before and after history loads, correct latest-question positioning, upload completion after switching chats, drop guards, and selected/detached sidebar status transitions.

willchen96:
### Additional changes made in this PR

- Fixed reopening a detached conversation so it loads at the **latest user message with the correct offset**, instead of showing the first user message at the top. The chat now waits for the full stored history before calculating the scroll position.
- Updated sidebar response indicators so **all actively streaming conversations show a loading spinner**, whether the conversation is detached or currently selected.
- Kept the green completion indicator only for **detached conversations**. If the response finishes while the user is already viewing that chat, the icon returns to its normal state without briefly showing green.
- Prevented file drops while the composer is unavailable or chat history is still loading.
- Isolated pending uploads between conversations so a late upload callback from one chat cannot attach a document or display progress/errors in another chat after navigation.
- Added regression coverage for latest-message positioning, selected and detached loading states, selected-chat completion behavior, and uploads completing after a chat switch.

PR #493, opened 2026-09-16
Explain provider access and quota failures in chat

### LINKED ISSUE #492: Chat shows 'Sorry, something went wrong' when a provider key can't access the selected model
## Problem

Adding a Gemini API key on the free tier, then chatting with a model that tier doesn't include (for example `gemini-2.5-pro`), fails with a generic "Sorry, something went wrong." The user gets no hint that the key or plan is the cause.

## Cause

1. `backend/src/lib/llm/aiSdk.ts` rewraps AI SDK stream errors as plain `Error`s, dropping the provider's status code and response body. `runLLMStream` then treats them as unknown failures and replaces the message with the generic one.
2. The chat routes (`chat`, `projectChat`, `wordChat`, `tabular`) always send `ASSISTANT_ERROR_MESSAGE` over SSE, even when the stream recorded a `safe_to_display` error. The specific message only reaches saved history.

## Expected

Known provider failures get a clear, actionable message:

- Rejected key (401, or "API key not valid") → check the key.
- 402 → out of credits.
- 403 / 404 → the model isn't available with this key.
- 429 with `limit: 0` (Gemini free tier) → the plan doesn't include the model.
- Other 429 → rate limit or quota reached.

Unclassified errors still show the generic message, so raw provider output never reaches the UI.

## Follow-up (out of scope)

The model picker still lists models a key can't use. Checking model access when a key is saved would stop users picking those models at all.

mcclowes:
## Summary

When a provider key can't access the selected model, for example a Gemini free-tier key with `gemini-2.5-pro`, chat now says what went wrong instead of "Sorry, something went wrong."

Fixes #492

## Why / Motivation

Free-tier keys return a 429 (`limit: 0`), 403, or 404 for models they don't include. `main` now classifies rejected API keys (`apiKeyErrors.ts`) and forwards safe errors to the client (`assistantStreamErrorPayload`), but every other access failure still reaches users as the generic message. Retried failures are also missed: a 429 arrives wrapped in the AI SDK's `RetryError`, so its status code sits on `lastError`.

## Changes

- New `backend/src/lib/llm/providerErrors.ts` with `toProviderStreamError`, which unwraps retry errors and then:
  - Uses the existing `asInvalidApiKeyError` first, so key rejections keep their current message.
  - 402: out of credits.
  - 403 / 404 (not naming the key): model isn't available with this key.
  - 429 with `limit: 0`: plan doesn't include the model.
  - Other 429: rate limit or quota reached.
- `aiSdk.ts` uses it for stream error parts and thrown errors, replacing `streamFailure`. Unclassified errors are unchanged.
- No route changes needed: they already forward `UserFacingError`s through `assistantStreamErrorPayload`.

## Tradeoffs & risks

- Provider errors are matched by shape (`statusCode`, `lastError`) rather than with `APICallError.isInstance`, because the AI SDK is ESM-only and only imported dynamically here.
- Messages say "the" key/plan, not "your", because `requiredKey()` may fall back to the deployment's key (same reasoning as `InvalidApiKeyError`).
- Detecting the free tier relies on Gemini's `limit: 0` wording. If that changes, the error falls back to the general rate-limit message.
- Only our own messages reach the UI. Raw provider bodies are never shown.
- Follow-up, out of scope: the model picker still lists models a key can't use.

## How verified

- Rebased onto `main` after the `routes/` → `modules/` move.
- `npx vitest run` in `backend`: 2,167 passed, 39 skipped.
- `npx tsc --noEmit -p backend`: clean.
- Not yet checked manually against a real free-tier Gemini key.

## Checklist

- [ ] Updated docs / env examples if setup, config, or behavior changed. (Not needed.)


willchen96:
Thanks @mcclowes, there was a large refactor to modularise the backend routes that was merged in a day ago. Grateful if you could rebase. Thanks.

willchen96 (approved):

willchen96:
Thanks @mcclowes

Issue #494, opened 2026-09-16
CourtListener settings show Remove for a server-provided API key


mcclowes:
## Problem

On **Settings → Features**, the CourtListener `ApiKeyField` decides whether a key is saved with `apiKeys.courtlistener.configured`. That flag is also true when the key comes from the server environment (`source === "env"`), not only from the user.

With a deployment-wide CourtListener key set, users see a saved-key state and a **Remove** button. Remove deletes the user's (non-existent) `user_api_keys` row, so nothing changes and the saved state persists.

The BYOK page already gets this right by checking `apiKeys[provider].source === "user"`.

## Repro

1. Set a CourtListener API key in the backend environment.
2. Sign in as a user with no personal CourtListener key.
3. Go to Settings → Features and enable CourtListener.
4. The key field shows as saved with a Remove button; clicking Remove has no effect.

## Fix

In `frontend/src/app/(pages)/settings/features/page.tsx`, use `profile?.apiKeys.courtlistener.source === "user"` for `hasSavedKey`, and add a regression test.

Optionally, tell users when the server's key is in use, e.g. "Using the server's key. Add your own to override."

DIYA73:
Taking this. Fix is one line in `frontend/src/app/(pages)/settings/features/page.tsx` — changing `hasSavedKey` to check `source === "user"` instead of `.configured`, matching the BYOK page pattern. Will include a regression test.

willchen96:
Thanks both, the exact fix is already on main via [PR #498](https://github.com/open-legal-products/mike/pull/498): hasSavedKey now checks source === "user". [PR #500](https://github.com/open-legal-products/mike/pull/500) by @DIYA73 was closed.

Issue #489, opened 2026-09-16
CLA form is blank/can't be completed


mcclowes:
<img width="1854" height="1191" alt="Image" src="https://github.com/user-attachments/assets/20c07c64-ab3d-4aec-9e48-fe043ec44515" />

mcclowes:
Now seems to be working ok

PR #495, opened 2026-09-16
Show saved API keys as a masked input

mcclowes:
## Summary

Saved API keys now appear as a masked password field instead of an empty input with "Saved key hidden" placeholder text.

## Why / Motivation

An empty box with grey placeholder text reads as "no key set". A masked value is the expected pattern for a stored secret.

## Changes

- With a saved key, `ApiKeyField` shows a read-only masked value.
- Focusing the field clears the mask so a replacement key can be typed (placeholder: "Enter a new key to replace").
- Blurring without typing restores the mask. The saved key is never modified by clearing the box; deletion is still only via Remove.
- Fields with no saved key are unchanged.

## Tradeoffs & risks

- The backend only returns whether a key exists, so the mask is a fixed-length stand-in rather than a last-4 hint. Adding one would need the API to expose a key suffix.
- Related, pre-existing: CourtListener uses `configured` rather than `source === "user"`, so a server-provided key now shows as masked with a no-op Remove. Tracked in #494.
- Out of scope, possible follow-ups: apply the same pattern to connector bearer tokens, replace `alert()` errors with inline messages, and confirm before Remove.

## How verified

- Added `ApiKeyField.test.tsx` covering the masked state, the empty state, replacing a key, and restoring the mask on blur.
- `npx vitest run src/app/components/settings/ApiKeyField.test.tsx` (4 passed), `eslint` clean, no type errors in the changed files.
- Not checked manually in the running app.

## Checklist

- [ ] Updated docs / env examples if setup, config, or behavior changed. (N/A)


willchen96:
Thanks @mcclowes please sign the CLA

mcclowes:
Hi Will, it was showing as blank when I raised this (https://github.com/open-legal-products/mike/issues/489), but it's working for me now. Have signed.

willchen96:
Thanks @mcclowes , appreciate the contribution

willchen96 (approved):

PR #497, opened 2026-09-18
Bind chat Retry to the turn it was raised for (no duplicate chats, no cross-thread resends)

amal66:
## Summary

Binds every chat error toast's **Retry** to the exact turn it was raised for, so a retry re-sends the same message into the same chat and refuses, with a plain notice, when that is no longer true. Stacked on #496 (the branch that introduced the Retry); found by the due-diligence review of that branch.

## What changed

- `useAssistantChat.ts`: a `chatIdRef` mirrors the chat id and is written wherever it changes (route prop, the stream's `chat_id` frame, `handleNewChat`, `resetChat`); both send paths post `chat_id: chatIdRef.current`.
- Each send captures an immutable turn snapshot `{ turnId, threadKey, chatId, message, opts, transcriptLength, anchor }`; each toast's Retry is bound to its own snapshot (`retryTurn(turn)`), not to a mutable last-turn ref.
- Before resending, `retryTurn` checks the thread is still the same, no newer turn is in flight (it never aborts a healthy stream), and the transcript still ends in the turn's own question. Otherwise it shows "This chat has changed…" or "A newer message has replaced this one…" with no Retry and does nothing destructive.
- Adjacent, same file: a `safe_to_display` error frame (configuration refusal such as a missing API key) now gets Contact support but no Retry (`retryable: !safeToDisplay`); the non-ok response path carries `status`, `code`, `requestId` (body or `x-request-id`) and the 4xx `detail`, so the message is specific and the support email carries the id.

## Why

On #496, the toast's Retry closed over the render that sent the message. In a project's New Chat the stream assigns the chat id after the send, so Retry re-sent `chat_id: undefined`: the backend created a **second chat** and the page navigated to it, orphaning the failed turn. `lastTurnRef` was last-write-wins, so a sticky toast could resend a different turn and abort a healthy stream; after a thread switch, Retry dropped the new thread's last bubble and re-sent the old question into it.

## Reproduce the base case (branch `olp-pr/error-handling-ux`)

1. Configure a provider key the provider rejects (any invalid `ANTHROPIC_API_KEY`), open a project, click "+ Chat" (New Chat state, URL ends in `/assistant/chat`), pick an Anthropic model.
2. Send a message. The stream assigns a chat id (URL becomes `/assistant/chat/<id-1>`), then the provider refuses and the toast "Couldn't get a response" appears with Retry.
3. Click Retry. **URL changes to `/assistant/chat/<id-2>`** and the project's chat list has two chats.

before: retry creates a duplicate chat

Recorded on the base branch: chat list after the run = `7d1c09de, 8d756741` (two chats for one question).

## Verify on this branch

Same steps. After Retry the **URL stays on `<id-1>`** and the chat list gains exactly the one chat this run created.

after: retry resends into the same chat

Recorded on this branch: path stayed `e2328e1a`, chat list = `e2328e1a` plus the two pre-existing ones.

Also: a `safe_to_display` refusal shows Contact support with no Retry; a 429 shows the backend's own detail.

- **Retry is an exact resend.** It re-sends with the model the message was sent with; changing the composer model first does not change the resend. Unchanged from #496.
- **Refusals are conservative.** If the host switches away and back to the same chat, an old toast's Retry may say "A newer message has replaced this one…" rather than "This chat has changed"; either way it refuses safely and re-sends nothing.
- The end-of-stream navigation to a newly created chat still compares against the render-time `chatId`; switching it to the ref would suppress the legitimate first-time navigation (pinned by an existing test).

## Testing performed

- New `useAssistantChat.retry.test.ts` (6 tests, real hook, real `ReadableStream` SSE bodies). Without the fix, 5 fail: the resend created a second chat and navigated to it; chat A's question was re-sent into chat B; a stale toast resent during a healthy stream; a refusal offered Retry; a 429's detail and request id were dropped. With it, `npx vitest run src/app/hooks` = 15 files / 81 tests green; `tsc --noEmit` clean; eslint 0 errors.
- Live in Chrome against a local stack (dev server on 127.0.0.1:3030, backend :3031): the two GIFs above, before on the base branch and after on this branch, with the chat-list counts read from `/api/projects/<id>/chats`.

## Update 2026-09-17 (late): re-stacked on #496 at be16e531

Rebased onto the new #496 head (which is itself on #450's final 62ad7bc0). One test-only change folded into the commit: the retry test mocks `@/app/lib/errorReporting`, and #496 now also imports `isReported` from it, so the mock exports that as well. Behaviour unchanged; retry + chat-history suites 44/44 green, tsc clean.

https://claude.ai/code/session_01CBzWK35dDM95RfzFbkudKA


amal66:
Superseded: folded into #496 as commit 11 (`5ad9698b`) so the branch that introduces the chat Retry also ships the turn-scoping fix, instead of a follow-up PR fixing a bug the first one introduced. The replication steps, the before/after GIFs and the tests moved into #496's body under "Chat Retry turn scoping". The branch `olp-pr/chat-retry-turn-scope` stays on the remote for reference.

PR #500, opened 2026-09-19
fix(settings): hide Remove for server-provided CourtListener API key

### LINKED ISSUE #494: CourtListener settings show Remove for a server-provided API key

DIYA73:
Fixes #494

## What
`ApiKeyField` on Settings → Features was using `apiKeys.courtlistener.configured` for `hasSavedKey`. That flag is `true` even when the key comes from the server environment (`source === "env"`), so users saw a Remove button that had no effect.

## Fix
Check `profile?.apiKeys.courtlistener.source === "user"` instead — matching the pattern already used on the BYOK page.

## Test
Added `frontend/src/app/(pages)/settings/features/page.test.tsx` with two cases:
- env key → Remove button not rendered
- user key → Remove button rendered


DIYA73:
Closing — it looks like this was already addressed in upstream main during my rebase. Happy to pick up another issue.

